PR #9, opened 2016-01-21
Forcing

colinjcotter:
Creating a new class to implement the forcing.

### DIFF
--- a/dcore/forcing.py
+++ b/dcore/forcing.py
@@ -12,7 +12,7 @@ class Forcing(object):
         self.state = state
     
     @abstractmethod
-    def apply(self,scale,):
+    def apply(self, scale, x, x_out):
         """
         Function takes x as input, computes F(x) and returns 
         x_out = x + scale*F(x) 
--- a/dcore/forcing.py
+++ b/dcore/forcing.py
@@ -61,9 +61,19 @@ class CompressibleForcing(Forcing):
         L = (
             -inner(w,cross(Omega,u0))*dx #Coriolis term
             -div(theta0*w)*pi*dx #pressure gradient (volume integral)
-            +jump(*pi*dx #pressure gradient (volume integral)
+            +jump(w*theta0,n)*avg(pi)*dS_v #pressure gradient (surface integral)
+            +div(w)*state.Phi*dx #gravity term (Phi is geopotential)
         )
 
+        bcs = [DirichletBC(V2, 0.0, "bottom"),
+               DirichletBC(V2, 0.0, "top")]
+        
+        u_forcing_problem = LinearVariationalProblem(
+            a,L,self.uF, bcs=bcs
+        ) 
+
+        self.u_forcing_solver = LinearVariationalSolver(u_forcing_problem)
+        
     def apply(self, scaling, x_in, x_out):
 
         self.x0.assign(x_in)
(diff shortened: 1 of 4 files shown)


dcore/forcing.py
```diff
@@ -0,0 +1,73 @@
...
+            )
+
+    def apply(self, scaling, x_in, x_out):
+
+        self.x0.assign(x_in)
+
+        self.u_forcing_solver.solve() #places forcing in self.uF
+        self.uF.scale(scaling)
+        
+        uF, _, _ = split(x_out)
```
wence-: Wrong split?

dcore/forcing.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        Function takes x as input, computes F(x) and returns 
+        x_out = x + scale*F(x) 
+        as output.
+
+        :arg scale: parameter to scale the output by.
+        :arg x: :class:`.Function` object, the input Function.
+        :arg x_out: :class:`.Function` object, the output Function.
+        """
+        pass
```
wence-: If the forcing is always, compute f(x), and add it with some scaling, can the implementation of `apply` not live here?

colinjcotter: No, because x is the u+rho+theta, and we only solve for forcing for u here.
Other forcing might occur in theta or rho.

On 21 January 2016 at 16:47, Lawrence Mitchell <notifications@github.com>
wrote:

> In dcore/forcing.py
> <https://github.com/firedrakeproject/dcore/pull/9#discussion_r50427135>:
>
> > +
> > +    def __init__(self, state):
> > +        self.state = state
> > +
> > +    @abstractmethod
> > +    def apply(self,scale,):
> > +        """
> > +        Function takes x as input, computes F(x) and returns
> > +        x_out = x + scale*F(x)
> > +        as output.
> > +
> > +        :arg scale: parameter to scale the output by.
> > +        :arg x: :class:`.Function` object, the input Function.
> > +        :arg x_out: :class:`.Function` object, the output Function.
> > +        """
> > +        pass
>
> If the forcing is always, compute f(x), and add it with some scaling, can
> the implementation of apply not live here?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/firedrakeproject/dcore/pull/9/files#r50427135>.
>



-- 

colinjcotter: I should probably call it forcing-and-sourcing

On 21 January 2016 at 16:49, Colin Cotter <colin.cotter@imperial.ac.uk>
wrote:

> No, because x is the u+rho+theta, and we only solve for forcing for u
> here. Other forcing might occur in theta or rho.
>
> On 21 January 2016 at 16:47, Lawrence Mitchell <notifications@github.com>
> wrote:
>
>> In dcore/forcing.py
>> <https://github.com/firedrakeproject/dcore/pull/9#discussion_r50427135>:
>>
>> > +
>> > +    def __init__(self, state):
>> > +        self.state = state
>> > +
>> > +    @abstractmethod
>> > +    def apply(self,scale,):
>> > +        """
>> > +        Function takes x as input, computes F(x) and returns
>> > +        x_out = x + scale*F(x)
>> > +        as output.
>> > +
>> > +        :arg scale: parameter to scale the output by.
>> > +        :arg x: :class:`.Function` object, the input Function.
>> > +        :arg x_out: :class:`.Function` object, the output Function.
>> > +        """
>> > +        pass
>>
>> If the forcing is always, compute f(x), and add it with some scaling, can
>> the implementation of apply not live here?
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/firedrakeproject/dcore/pull/9/files#r50427135>.
>>
>
>
>
> --
> http://www.imperial.ac.uk/people/colin.cotter
>
> www.cambridge.org/9781107663916
>
>
>


-- 

PR #13, opened 2016-03-02
Linear advection

jshipton:
Many bugs have been fixed in this branch so I'd like to merge it back in to master, where the code does not currently run.


dcore/advection.py
```diff
@@ -0,0 +1,120 @@
...
+class NoAdvection(Advection):
+    """
+    An non-advection scheme that does nothing.
+    """
+
+    def __init__(self, state):
+        self.state = state
+
+        #create a ubar field even though we don't use it.
+        self.ubar = Function(state.V[0])
```
wence-: It's bad form not to call the superclass constructor.

dcore/advection.py
```diff
@@ -0,0 +1,120 @@
...
+        L = -p*dot(self.ubar,k)*dot(k,grad(qbar))*dx
+
+        aProblem = LinearVariationalProblem(a,L,self.dq)
+        if options == None:
+            options = {'ksp_type':'cg',
+                       'pc_type':'bjacobi',
+                       'sub_pc_type':'ilu'}
+            
+        self.solver = LinearVariationalSolver(aProblem,
+                                              solver_parameters = options)
```
wence-: ditto.

PR #19, opened 2016-03-08
setup testing and add one very minimal test

jshipton:


wence-:
I don't think you need the conftest.py file (it's sort of specific to the firedrake test suite).

tests/test_dcmip_setup.py
```diff
@@ -0,0 +1,161 @@
...
+    return stepper, dt
+
+def run_dcmip():
+
+    stepper, dt = setup_dcmip()
+    stepper.run(t=0, tmax=dt)
+
+def test_dcmip_runs():
+
+    run_dcmip()
```
wence-: Can you assert something about the state of the system when this has run?  Or is it only the case that you're testing for the stepper actually running?

wence-:
To wire this into travis, we'll need to install firedrake and then run the test suite in `.travis.yml`

tests/test_dcmip_setup.py
jshipton: Yes, I'm just testing that stepper runs - I really meant it when I said 'minimal'!

jshipton:
OK, I have no clue how this travis testing works - shall I delete the contest.py file? What next?

jshipton:
Thanks for looking at this Lawrence - I'm running for a train now but will carry on in the morning.

J


tests/test_dcmip_setup.py
```diff
@@ -0,0 +1,161 @@
+from __future__ import absolute_import
+from dcore import *
+from firedrake import IcosahedralSphereMesh, ExtrudedMesh, Expression, \
+    VectorFunctionSpace
+import numpy as np
+
+def setup_dcmip():
+
+    nlayers = 2         # 10 horizontal layers
```
wence-: This comment, it lies.

wence-:
You could look at the thetis travis file https://github.com/thetisproject/thetis/blob/master/.travis.yml for an example.  Note how we set the PETSc options we're going to need.  I think you don't need to build with mumps for this.  I also do "caching" of the installed firedrake instance.  We can try without that bit to see if things take too long.

PR #20, opened 2016-03-09
Tidy configuration

jshipton:


dcore/configuration.py
```diff
@@ -0,0 +1,81 @@
...
+Some simple tools for making model configuration nicer.
+"""
+
+
+class Configuration(object):
+    def __init__(self, **kwargs):
+
+        self.__setattr__ = self._proto__setattr__
+        for name, value in kwargs.iteritems():
+            print name, value
```
wence-: Do you really mean to print the values when initialising the configuration object.  This is pretty unfriendly in parallel.

dcore/configuration.py
```diff
@@ -0,0 +1,81 @@
...
+        self.g = 9.81
+        self.N = 0.01
+        self.cp = 1004.5
+        self.R_d = 287.
+        self.p_0 = 1000.0*1000.0
+        self.kappa = 2.0/7.0
+        self.k = None
+        self.Omega = None
+
+        super(CompressibleParameters, self).__init__(**kwargs)
```
wence-: If I understand this class correctly, I would use:

```python
class Configuration(object):

    def __init__(self, **kwargs):
        for name, value in kwargs.iteritems():
            self.__setattr__(name, value)

    def __setattr__(self, name, value):
        if not hasattr(self, name):
            raise AttributeError("'%s' object has no attribute '%s'" % (type(self).__name__, name))
        object.__setattr__(self, name, value)
```

Now the inherited configuration object don't have an init, they just provide a default value for each of the parameters you're allowed to set:

```python
class SubConfig(Configuration):
    a = None
    g = 9.81
```

Now, when I make a `SubConfig`, I can either just select the defaults:

```python
conf = SubConfig()
```

Or else override some defaults:
```python
conf = SubConfig(a=55)
```

If I try and set an "unsupported" parameter, then I get an attribute error:

```python
x.c = 14321

Truncated Traceback (Use C-c C-x to view full TB):
<ipython-input-31-d5c9fd37ecc0> in __setattr__(self, name, value)
      2     def __setattr__(self, name, value):
      3         if not hasattr(self, name):
----> 4             raise AttributeError("'%s' object has no attribute '%s'" % (type(self).__name__, name))
      5         object.__setattr__(self, name, value)
      6 

AttributeError: 'SubConfig' object has no attribute 'c'
```

I think this makes the code somewhat simpler, because you don't need init and so on.

Certainly, that's what the example below suggests the mechanism is.

dcore/configuration.py
```diff
@@ -0,0 +1,81 @@
...
+
+        #: Documentation for myparam
+        myparam = None
+
+        #: As in GEM, manual suggests 0.1
+        dt = 0.1
+
+    c = MyConfiguration(dt=2)
+
+    print c.dt
```
wence-: I would write this in the module documentation (rather than as an example at the bottom of the file).  You could test that things behave as you expect instead of this, I guess.

wence-: By which I mean, add a `tests/test_configuration.py` that makes assertions about how you expect the state of configuration objects to change.  e.g. One might be to assert that setting an invalid parameter raises an attribute error:


```python
import pytest
from dcore.configuration import *

class SubConfig(Configuration):
    a = 1
    b = None

def test_invalid_param_attributeerror():
    conf = SubConfig()
    with pytest.raises(AttributeError):
        conf.c = 10

    with pytest.raises(AttributeError):
        conf.c
```

dcore/configuration.py
jshipton: Hm, I hadn't really thought about parallel - but it is useful to have a record of which parameters are different from the default. I do tend to pipe the output to a log file but maybe there's a better way of doing this?


wence-: So the code documents the parameters used, right?  If you want to more formally record the parameter state used to run a simulation and dump it such that you can later recreate, then we could think about how to do that.  But printing out the values in the log file doesn't feel like the right way.

jshipton: I'm not sure what you mean when you say "the code" - the default parameters are there in the configuration class but can be overwritten when setup (e.g. in examples/dcmip_blah_blah_blah).  We should probably have the output going to a results directory whose name usually records some of the parameters used (e.g. dcmip_dx_dt) but it can't possibly record them all. I often use the same setup file to run with different parameters. I've used pickle to dump dictionaries of diagnostics - maybe we could dump a dictionary of parameters in the output directory?

wence-: So by "the code", I mean the file that you actually ran to produce the results.  I agree that the filename is the wrong place to put this.  The `State` object knows what all the user-specifiable configuration slots are.  So I think the correct thing to do is for `State` to gain a `dump` / `load` pair that you can either use to save the configuration used for a particular simulation, or else to load a configuration from a given save file.  I would be tempted to use JSON, rather than pickling, because then at least the dumped configuration is human readable.

jshipton: OK, that sounds good. I'll make a new branch for that.

PR #23, opened 2016-03-17
Dg advection

jshipton:


jshipton:
This is written with the shallow water code in mind but it would be nice if it also worked for future code (e.g. vertical slice). That's why I mention the convention that the DG space should be the last in the list of spaces. Is there a way of checking that this space is indeed DG (isinstance or something IIRC). Also, would you like me to call the field something more generic (like q?).

dcore/advection.py
```diff
@@ -122,3 +115,53 @@ def apply(self, x_in, x_out):
...
+class DGAdvection(Advection):
+
+    def __init__(self, state):
+
+        super(DGAdvection, self).__init__(state)
+
+        dt = state.timestepping.dt
+
+        # by convention the last space is the DG space
+        DGSpace = state.V[-1]
```
wence-: I would be tempted to redefine the `State` interface.  Currently, you access `state.V` which is an opaque lists of spaces which must conform to an implicit interface (note that this is not the only place where you do this, e.g. you assume that in `LinearAdvection_V3` the DG space is the second entry in `state.V`).

So, `State` could have a method (say):

```python
def get_space(self, name):
    return ...
```

You need to know the ordering (for the mixed function space), but build spaces should build the spaces with a descriptive name (e.g. "temperature", "pressure"), so you can get them again.  The state subclasses can document the available function space names (and say something about them).  This way you have fewer magic numbers.

wence-: You can assert that the space you got back is DG like so:

```python
element = DGSpace.fiat_element
assert element.entity_dofs() == element.entity_closure_dofs(), "Provided space is not discontinuous"
```

jshipton: I like the idea of a get_space method - much better than a list. I'm just hesitating because I don't know what to call the spaces… I don't want to name them after the field because that is different depending on the equations we are solving - for shallow water the DG field is the depth field, for vertical slice it is density (or pressure if you're incompressible). OK so I can call the DG space "DGspace" but what about the velocity space, which is RT if you're on quads but BD(F)M if you're on triangles… it's a bit inconsistent but I suppose that is always the velocity space so I can name that one after the field… maybe this ramble has helped me figure out an answer?!

dham: Could you not just have a list and put it in Form order. That would provide
a definitive reason why DG is space -1, and would provide places for the
other ones?

wence-: I note that at the moment, the fully DG space is not the last space in all cases (it's the CP-staggered temperature space for the 3D stuff).

Of course, another option, given that making function spaces is cheap if you've already built one (and all the data are shared), you could just *make* a DG space in this function, then you don't have to worry!

jshipton: >> +class DGAdvection(Advection):
>> +
>> +    def __init__(self, state):
>> +
>> +        super(DGAdvection, self).__init__(state)
>> +
>> +        dt = state.timestepping.dt
>> +
>> +        # by convention the last space is the DG space
>> +        DGSpace = state.V[-1]


>Could you not just have a list and put it in Form order. That would provide a definitive reason >why DG is space -1, and would provide places for the other ones?

Yes, this was my reasoning - but I didn't realise Colin had dome something different for the 3D code.

colinjcotter:
It's a good idea to have a get_space() method in State. 

However, for this pull request, I was actually thinking that the constructor for the Advection class should provide the space that should be used.

jshipton:
This won't pass tests as I wrote some of it forgetting that the latest commits to the organise_output branch hadn't been merged. Ignore for now - I'm playing around with adding embedded dg advection as most of the work is done already. One thing I wasn't sure of is how we want to add in extra fields, such as tracers, that aren't part of the mixed system...

colinjcotter:
Please can you clean this up and just have DG advection, not embedded DG in this PR.

jshipton:
I think this is good to go now (once tests pass on travis) - any thoughts anyone?

tests/test_DG_advection.py
```diff
@@ -0,0 +1,107 @@
...
+    refinements = 3  # number of horizontal cells = 20*(4^refinements)
+    R = 1.
+    dt = pi/3*0.001
+
+    mesh = IcosahedralSphereMesh(radius=R,
+                                 refinement_level=refinements)
+    global_normal = Expression(("x[0]/sqrt(x[0]*x[0]+x[1]*x[1]+x[2]*x[2])",
+                                "x[1]/sqrt(x[0]*x[0]+x[1]*x[1]+x[2]*x[2])",
+                                "x[2]/sqrt(x[0]*x[0]+x[1]*x[1]+x[2]*x[2])"))
+    mesh.init_cell_orientations(global_normal)
```
wence-: You know the normal doesn't have to be normalised, right?

dcore/advection.py
```diff
@@ -122,3 +115,56 @@ def apply(self, x_in, x_out):
         dt = self.state.timestepping.dt
         self.solver.solve()
         x_out.assign(x_in + dt*self.dq)
+
+
+class DGAdvection(Advection):
+
+    def __init__(self, state, V, continuity=False):
+
```
wence-: docstring for what these things mean?

tests/test_DG_advection.py
jshipton: Heh, oops.


tests/test_DG_advection.py
```diff
@@ -0,0 +1,107 @@
...
+def test_dgadvection_vector():
+
+    f_err = run(vector=True)
+    assert(abs(f_err.dat.data.max()) < 1.5e-2)
+
+
+def test_dgadvection_continuity_vector():
+
+    f_err = run(continuity=True, vector=True)
+    assert(abs(f_err.dat.data.max()) < 1.5e-2)
```
wence-: AKA:

```python
import pytest

@pytest.parametrize("vector", [False, True], ids=["scalar", "vector"])
@pytest.parametrize("continuity", [False, True])
def test_dgadvection(vector, continuity):
    f_err = run(continuity=continuity, vector=vector)
    assert abs(f_err.dat.data_ro.max()) < 1.5e-2
```

jshipton: Nice. I had a feeling you'd tell me there was an easier way.


tests/test_DG_advection.py
```diff
@@ -0,0 +1,107 @@
...
+    mesh = IcosahedralSphereMesh(radius=R,
+                                 refinement_level=refinements)
+    global_normal = Expression(("x[0]/sqrt(x[0]*x[0]+x[1]*x[1]+x[2]*x[2])",
+                                "x[1]/sqrt(x[0]*x[0]+x[1]*x[1]+x[2]*x[2])",
+                                "x[2]/sqrt(x[0]*x[0]+x[1]*x[1]+x[2]*x[2])"))
+    mesh.init_cell_orientations(global_normal)
+
+    fieldlist = ['u','D']
+    timestepping = TimesteppingParameters(dt=dt)
+    output = OutputParameters(dumpfreq=150, dirname='results/tests/'+dirname)
```
wence-: Is this ever used?  Can one pass `None` as output parameters?

tests/test_DG_advection.py
```diff
@@ -0,0 +1,107 @@
...
+        f_end = Function(state.V[1])
+        f_end_expr = Expression("exp(-pow(x[2],2) - pow(x[0],2))")
+
+    f.interpolate(fexpr)
+    f_end.interpolate(f_end_expr)
+
+    return state, u0, f, f_end
+
+
+def run(continuity=False, vector=False):
```
wence-: It's somewhat unpleasant that the test writes a load of output whereever it's run.  You can get pytest to provide you with a temporary directory like so:

```python
def test_foo(tmpdir, ...):
    # tmpdir is a per-test temporary directory.
    dirname = str(tmpdir.join("DGAdvection/continuity" + str(continuity)))
```

tests/test_DG_advection.py
```diff
@@ -13,9 +13,7 @@ def setup_DGadvection(dirname, vector=False):
 
     mesh = IcosahedralSphereMesh(radius=R,
                                  refinement_level=refinements)
-    global_normal = Expression(("x[0]/sqrt(x[0]*x[0]+x[1]*x[1]+x[2]*x[2])",
-                                "x[1]/sqrt(x[0]*x[0]+x[1]*x[1]+x[2]*x[2])",
-                                "x[2]/sqrt(x[0]*x[0]+x[1]*x[1]+x[2]*x[2])"))
+    global_normal = Expression(("x[0]", "x[1]", "x[2]"))
```
wence-: It's not just when R is one.  You *never* have to provide a normalised vector here.  We just use it for the direction.

jshipton: Oh, ok.


dcore/advection.py
```diff
@@ -122,3 +115,69 @@ def apply(self, x_in, x_out):
...
+    """
+    DG 3 step SSPRK advection scheme that can be applied to a scalar 
+    or vector field
+
+    :arg state: :class:`.State` object.
+    :arg V: function space of advected field - should be DG
+    :arg continuity: optional boolean
+         if True, the advection equation is of the form:
+         D_t + (u.\nabla)D = 0
+         if False, the advection equations in of the form:
```
wence-: s/in/is/.

Note as well you can use "latex" markup here (which will then format nicer when using sphinx).

```python
:arg continuity: optional boolean.  If ``True``, the advection equation is of the form
    :math:`D_t + (u \cdot \nabla)D = 0`.  If ``False``, use :math:`D_t + \nabla \cdot (uD)`.
```

colinjcotter:
Continuity flag is the wrong way around.

colinjcotter:
Once the flag is fixed then merge if it passes tests.

PR #24, opened 2016-03-17
make a new directory for dumping results, checking first that is does…

jshipton:
… not already exist

### DIFF
--- a/examples/sw_williamson2_triangle.py
+++ b/examples/sw_williamson2_triangle.py
@@ -16,7 +16,7 @@ global_normal = Expression(("x[0]/sqrt(x[0]*x[0]+x[1]*x[1]+x[2]*x[2])",
 mesh.init_cell_orientations(global_normal)
 
 timestepping = TimesteppingParameters()
-output = OutputParameters()
+output = OutputParameters(dumplist=(True,True), dirname='sw_williamson2')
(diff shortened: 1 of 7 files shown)


colinjcotter:
Is it a problem that the default for dumplist is None, and not an empty tuple, (,) ?

jshipton:
yes, I wasn't sure how to give a suitable default as I don't know  how long it should be: i.e dumplist=(T,T,T) for 3d compressible but (T,T) for shallow water.
You could have if dumplist is None then dumplist=list of Trues the same length as fieldlist.

colinjcotter:
ok, or just don't have a default argument for dumpiest

examples/sw_williamson2_triangle.py
```diff
@@ -16,7 +16,7 @@
 mesh.init_cell_orientations(global_normal)
 
 timestepping = TimesteppingParameters()
-output = OutputParameters()
+output = OutputParameters(dumplist=(True,True), dirname='sw_williamson2')
```
dham: Wouldn't it be much more intelligible to name the fields you are dumping?

jshipton: Yes.


jshipton:
If this passes tests I think it's good to go.

PR #25, opened 2016-03-23
New output

wence-:
Relies on firedrakeproject/firedrake#737.


wence-:
Any thoughts?  This will remove the warnings about deprecated "<<" syntax.  The other thing to consider is whether you want all fields dumped in the same VTU or in different VTUs.

jshipton:
Looks great to me. Very happy to have fields in the same vtu.

PR #30, opened 2016-04-25
use a temporary directory for all test output

jshipton:
@wence- it seems like pytest creates the tmpdir which causes a slight problem as the dcore code expects to be creating a new directory for output so that results don't get overwritten. i've got round it by making a subdirectory of the tmpdir. Is there a better way?

### DIFF
--- a/tests/test_sw_triangle.py
+++ b/tests/test_sw_triangle.py
@@ -3,7 +3,7 @@ from firedrake import IcosahedralSphereMesh, Expression, SpatialCoordinate, \
     Constant, as_vector
 
 
-def setup_sw():
+def setup_sw(dirname):
 
     refinements = 3  # number of horizontal cells = 20*(4^refinements)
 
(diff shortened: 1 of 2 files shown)


wence-:
That looks right to me.

tests/test_sw_triangle.py
```diff
@@ -19,7 +19,7 @@ def setup_sw():
 
     fieldlist = ['u', 'D']
     timestepping = TimesteppingParameters()
-    output = OutputParameters(dumplist=(True,True), dirname='tests/sw')
+    output = OutputParameters(dumplist=(True,True), dirname=dirname)
```
wence-: do you not mean dirname/sw here?

jshipton: Oh yes, good point. It doesn't break because it doesn't run yet, it just sets up spaces and initial conditions.... but that should change soon!

PR #31, opened 2016-04-26
Linear advection

jshipton:
This merges in a lot of changes from Colin's linear advection branch. In particular, we can now run a compressible vertical slice (hence Compressible3DState has been renamed to CompressibleState).

I have started to make a test for the linear Skamarock and Klemp non hydrostatic gravity wave. It runs on a smaller domain at lowish resolution for enough time for the wave to have started to spread out. I would appreciate some advice on how to make this into an actual test - should I commit a vtu of the final state and compare to that? If so, can someone remind me how to read in from a vtu please?!


dcore/state.py
```diff
@@ -72,11 +76,19 @@ def dump(self):
 
         funcs = self.xn.split()
         to_dump = []
+
         for name, f in zip(self.fieldlist, funcs):
             if name in self.output.dumplist:
                 to_dump.append(f)
             f.rename(name=name)
 
+        for field, meanfield in zip(funcs, self.output.meanfields):
```
jshipton: I think we're going to need a more flexible way of adding diagnostic fields soon but it this ok for now? I've just realised I could use the same setup to check steady state solutions but putting the initial field into the meanfields list (in which case maybe I should rename meanfields...)

examples/sk_linear_advection.py
```diff
@@ -0,0 +1,158 @@
...
+kappa = parameters.kappa
+
+# N^2 = (g/theta)dtheta/dz => dtheta/dz = theta N^2g => theta=theta_0exp(N^2gz)
+Tsurf = 300.
+thetab = Tsurf*exp(N**2*z/g)
+
+theta_b = Function(state.V[2]).interpolate(thetab)
+rho_b = Function(state.V[1])
+
+# Calculate hydrostatic Pi
```
jshipton: I think this hydrostatic initialisation should be moved to an initialise method in the CompressibleState class, I only hesitate because I'm not sure if it works for anything other than the vertical slice setup.

colinjcotter: It's a bit dependent on boundary conditions and things like that. I think we should eventually make a initialisation_tools class.

PR #32, opened 2016-04-26
Linear sw

jshipton:
This pull request adds a linear solver class for shallow water and a forcing class for shallow water. It is tested by running the linear Williamson 2 test case as documented in Weller et al 2012. In order to check the result I have added the option to compute error fields in steady state simulations. I have also added a very basic diagnostics class and setup a diagnostics dictionary that is dumped using json at the end of the simulation - very happy to hear if there's a better way of doing this!

### DIFF
--- a/dcore/configuration.py
+++ b/dcore/configuration.py
@@ -35,7 +35,7 @@ class OutputParameters(Configuration):
     """
 
     Verbose = False
-    dumpfreq = 10
+    dumpfreq = 1
     dumplist = None
     dirname = None
     #: Should the output fields be interpolated or projected to
@@ -63,5 +63,6 @@ class ShallowWaterParameters(Configuration):
     """
     Physical parameters for 3d Compressible Euler
     """
-    g = 9.806
+    g = 9.80616
     Omega = 7.292e-5  # rotation rate
+    H = None  # mean depth
--- a/dcore/configuration.py
+++ b/dcore/configuration.py
@@ -41,6 +41,8 @@ class OutputParameters(Configuration):
     #: Should the output fields be interpolated or projected to
     #: a linear space?  Default is interpolation.
     project_fields = False
+    #: Dump error fields for steady state simulation. Default is no.
+    steady_state_dump_err = False
 
 
 class CompressibleParameters(Configuration):
--- a/dcore/diagnostics.py
+++ b/dcore/diagnostics.py
@@ -4,15 +4,16 @@ from math import sqrt
 
 class Diagnostics(object):
 
-    def __init__(self, fieldlist=None):
+    def __init__(self, *fields):
 
-        if fieldlist is not None:
-            self.fields = fieldlist
-        else:
-            self.fields = []
+        self.fields = list(fields)
 
-    def register(self, fieldlist):
-        self.fields += fieldlist
+    def register(self, *fields):
+
+        fset = set(self.fields)
+        for f in fields:
+            if f not in fset:
+                self.fields.append(f)
 
     @staticmethod
     def l2(f):
--- a/dcore/diagnostics.py
+++ b/dcore/diagnostics.py
@@ -3,10 +3,16 @@ from math import sqrt
 
 
 class Diagnostics(object):
-    pass
+
+    def __init__(self):
+        self.fields = []
+
+    def register(self, fieldlist):
+        self.fields += fieldlist
 
 
 class ShallowWaterDiagnostics(Diagnostics):
 
-    def l2(self, f):
+    @staticmethod
+    def l2(f):
         return sqrt(assemble(dot(f, f)*dx))
--- a/dcore/diagnostics.py
+++ b/dcore/diagnostics.py
@@ -4,14 +4,15 @@ from math import sqrt
 
 class Diagnostics(object):
 
-    def __init__(self):
-        self.fields = []
+    def __init__(self, fieldlist=None):
+
+        if fieldlist is not None:
+            self.fields = fieldlist
+        else:
+            self.fields = []
 
     def register(self, fieldlist):
         self.fields += fieldlist
-
-
-class ShallowWaterDiagnostics(Diagnostics):
 
     @staticmethod
     def l2(f):
--- a/dcore/diagnostics.py
+++ b/dcore/diagnostics.py
@@ -1,0 +1,12 @@
+from firedrake import assemble, dot, dx
+from math import sqrt
+
+
+class Diagnostics(object):
+    pass
+
+
+class ShallowWaterDiagnostics(Diagnostics):
+
+    def l2(self, f):
+        return sqrt(assemble(dot(f, f)*dx))
--- a/dcore/state.py
+++ b/dcore/state.py
@@ -1,13 +1,12 @@
 from __future__ import absolute_import
 from os import path
 import itertools
+import json
 from sys import exit
 from abc import ABCMeta, abstractmethod
 from firedrake import FiniteElement, TensorProductElement, HDiv, \
     FunctionSpace, MixedFunctionSpace, interval, triangle, Function, \
-    Expression, File, CellNormal, grad, cross, solve, inner, dx, \
-    TestFunctions, TrialFunctions, div, assemble, dot
-from math import sqrt
+    Expression, File
 
 
 class State(object):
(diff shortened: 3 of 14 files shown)


dcore/state.py
```diff
@@ -71,24 +74,47 @@ def dump(self):
...
 
+            if self.diagnostics is not None:
+                self.diagnostic_data = {}
+                for name in field_dict.keys():
+                    self.diagnostic_data[name] = {"l2":[]}
+
         if (next(self.dumpcount) % self.output.dumpfreq) == 0:
             self.dumpfile.write(*to_dump)
 
+            if self.diagnostics is not None:
```
jshipton: @wence- is there a better way of handling the case where the user hasn't specified any diagnostics? I suppose we could make dumping the diagnostics the default behaviour - it's not too time consuming at the moment. What do you think?

wence-: So I would do the following.  Have a `register` method on the Diagnostics class.  So now I do:

```python
diags = MyDiagnostics()
diags.register(names, of, fields)

state = State(..., diags=diags, ...)
```

Then in `dump`.

```python
for name in diag.fields:
    data = diag.l2(field_dict[name])
   diag_data[name]["l2"].append(data)
```

wence-: You can make the diagnostic data dict easier like so:
```
from collections import defaultdict
from functools import partial

self.diagnostic_data = defaultdict(partial(defaultdict, list))

for name in diag.fields:
    self.diagnostic_data[name]["l2"].append(...)
```

No need to worry about initialising it with empty things.

dcore/diagnostics.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class Diagnostics(object):
+    pass
+
+
+class ShallowWaterDiagnostics(Diagnostics):
+
+    def l2(self, f):
+        return sqrt(assemble(dot(f, f)*dx))
```
wence-: This should either be an `@staticmethod` or an `@classmethod`.  See https://julien.danjou.info/blog/2013/guide-python-static-class-abstract-methods for deciding what to do.

colinjcotter:
Plan for diagnostics looks fine to me.

dcore/state.py
wence-: I guess you could have the field registration just happen in the constructor (if you're never going to register diagnostics after creating the Diagnostics object).

dcore/diagnostics.py
```diff
@@ -0,0 +1,19 @@
+from firedrake import assemble, dot, dx
+from math import sqrt
+
+
+class Diagnostics(object):
+
+    def __init__(self, fieldlist=None):
```
wence-: You could do:

```python
def __init__(self, *fields):
    self.fields = list(fields)

def register(self, *fields):
    self.fields += fields
```

That way the caller doesn't need to wrap fields up in a list.

dcore/state.py
```diff
@@ -71,24 +77,44 @@ def dump(self):
...
             f.rename(name=name)
 
-        dumpdir = path.join("results", self.output.dirname)
+        if self.output.steady_state_dump_err:
+            init_funcs = self.x_init.split()
+            for name, f, f_init in zip(self.fieldlist, funcs, init_funcs):
+                if name in self.output.dumplist:
+                    err = Function(f.function_space(), name=name+'err').assign(f-f_init)
+                    field_dict[name+"err"] = err
+                    self.diagnostics.register([name+"err"])
```
wence-: With my change, this just becomes `self.diagnostics.register(name + "err")`.  Which is bit friendlier.

dcore/diagnostics.py
wence-: You might want to check that the to be registered fields are not already in the existing fields.

```python
def register(self, *fields):
    fset = set(self.fields)
    for f in fields:
        if f not in fset:
            self.fields.append(f)
```

examples/dcmip_3-1_linear_advection_triangle_bdfm1.py
```diff
@@ -40,12 +40,14 @@
 timestepping = TimesteppingParameters(dt=10.0)
 output = OutputParameters(Verbose=True, dumpfreq=1, dirname='dcmip')
 parameters = CompressibleParameters(k=k, Omega=Omega)
+diagnostics = Diagnostics(fieldlist)
```
wence-: I think that these fields will be registered twice (once in initialisation here) and then again in State.dump?

jshipton: I don't think so... the register method is only called if steady_state_dump_err is True and in that case it registers the error field (which is f - f_init). At least, that's what I want to happen!


wence-: Oh, I see, yes.

dcore/state.py
```diff
@@ -71,24 +77,44 @@ def dump(self):
...
             f.rename(name=name)
 
-        dumpdir = path.join("results", self.output.dirname)
+        if self.output.steady_state_dump_err:
+            init_funcs = self.x_init.split()
+            for name, f, f_init in zip(self.fieldlist, funcs, init_funcs):
+                if name in self.output.dumplist:
+                    err = Function(f.function_space(), name=name+'err').assign(f-f_init)
+                    field_dict[name+"err"] = err
+                    self.diagnostics.register(name+"err")
```
wence-: Note that you don't actually need to allocate a new function for this guy if you're only going to use it in the diagnostics.  It suffices to put `f - f_init` in `field_dict`.  Because when you compute the l2 error you'll end up inserting `assemble(sqrt(dot(f - f_init, f-f_init))*dx)` which works fine.  If you're also going to output it for viz, you can't do this trick though.

jshipton: I need it for viz too at the moment but I'll bear that in mind, thanks.


dcore/state.py
```diff
@@ -75,32 +81,58 @@ def dump(self):
...
-        for name, f in zip(self.fieldlist, funcs):
+        for name, f in field_dict.iteritems():
             if name in self.output.dumplist:
                 to_dump.append(f)
             f.rename(name=name)
 
-        for field, meanfield in zip(funcs, self.output.meanfields):
+        steady_state_dump_err = self.output.steady_state_dump_err
+        for field in self.fieldlist:
+            steady_state_dump_err.setdefault(field, False)
```
jshipton: I think I can't use defaultdict here because I need specific keys in the dictionary - is that correct?

dcore/configuration.py
```diff
@@ -35,22 +35,22 @@ class OutputParameters(Configuration):
...
     #: a linear space?  Default is interpolation.
     project_fields = False
-    #: Can be a list of fields for computing perturbations.
-    #: Must be in the same order as fieldlist with entry None
-    #: if a particular perturbation field is not required.
-    meanfields = [None, None, None]
+    #: Dump error fields for steady state simulation. Default is False.
+    steady_state_dump_err = {}
+    #: Dictionary of fields for computing perturbations. Default is False.
+    meanfields = {}
```
jshipton: meanfields is now a dictionary - I didn't like the assumption that it should have 3 entries (corresponding to the 3 fields in the compressible model) although it's a little pedantic as shallow water doesn't have mean fields...

dcore/state.py
```diff
@@ -75,32 +81,58 @@ def dump(self):
...
+        for name, f, f_init in zip(self.fieldlist, funcs, self.x_init.split()):
+            if steady_state_dump_err[name]:
+                err = Function(f.function_space(), name=name+'err').assign(f-f_init)
+                field_dict[name+"err"] = err
+                self.diagnostics.register(name+"err")
+                to_dump.append(err)
+
+        meanfields = self.output.meanfields
```
jshipton: The meanfields and steady_state_dump_err loops can be combined if/when we write more general diagnostic fields - they are both just simply taking the difference of 2 fields.

dcore/state.py
wence-: `defaultdict` says nothing about what the keys are, what you are doing is providing a callable that constructs a default *value* if the key does not already exist in the dict.

```python

a = defaultdict(list)
a[10] => []
a[12] = 44
a[12] => 44
a[15].append(22)
a[15] => [22]
```

jshipton: I think I get it - I wasnt sure how to combine the defaultdict with setting specific values from a supplied dictionary but I've just remembered `update`... So this is better?
`steady_state_dump_err = defaultdict(lambda: False)
 steady_state_dump_err.update(self.output.steady_state_dump_err)`

wence-: That looks fine.  You can just use `defaultdict(bool)`

PR #33, opened 2016-05-04
Advection bugfixes

jshipton:
Bugfixes:
1) remove unused function space from EmbeddedDGAdvection
2) in DGAdvection define dS depending on whether mesh is extruded or not

Added an example file for the nonlinear SK gravity wave. Currently with no u advection as we're investigating stability issues.

### DIFF
--- a/dcore/advection.py
+++ b/dcore/advection.py
@@ -188,19 +188,17 @@ class DGAdvection(Advection):
         x_out.assign((1.0/3.0)*x_in + (2.0/3.0)*self.dD)
 
 
-class EmbeddedDGAdvection(Advection):
+class EmbeddedDGAdvection(DGAdvection):
 
     def __init__(self, state, Vdg, continuity):
 
-        super(EmbeddedDGAdvection, self).__init__(state)
-        self.dgadvection = DGAdvection(state, Vdg, continuity)
+        super(EmbeddedDGAdvection, self).__init__(state, Vdg, continuity)
 
         self.xdg_in = Function(Vdg)
         self.xdg_out = Function(Vdg)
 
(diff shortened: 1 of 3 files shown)


dcore/advection.py
```diff
@@ -185,7 +190,7 @@ def apply(self, x_in, x_out):
 
 class EmbeddedDGAdvection(Advection):
 
-    def __init__(self, state, V, Vdg, continuity):
+    def __init__(self, state, Vdg, continuity):
 
         super(EmbeddedDGAdvection, self).__init__(state)
         self.dgadvection = DGAdvection(state, Vdg, continuity)
```
wence-: While you're here, I think you should inherit from DGAdvection and then you can simplify things:

```python
class EmbeddedDGAdvection(DGAdvection):
    def __init__(self, state, Vdg, continuity):
        super(EmbeddedDGAdvection, self).__init__(state, Vdg, continuity)
        self.xdg_in = Function(Vdg)
        self.xdg_out = Function(Vdg)

    def apply(self, x_in, x_out):
        self.xdg_in.interpolate(x_in)
        # Call normal DG advection on DG fields via superclass
        super(EmbeddedDGAdvection, self).apply(self.xdg_in, self.xdg_out)
        x_out.project(self.xdg_out)
```

That way you're not carrying around an extra ubar that you need to move from this advection object to the dgadvection object inside it.

jshipton: That is what I'd originally intended... colinjcotter, can you remember why you wanted it done this way?

colinjcotter: yes, let's take @wence-'s suggestion.

PR #34, opened 2016-05-04
Courant number

jshipton:
This adds in the option to calculate diagnostic fields. At the moment just the Courant number is coded.

It feels a bit icky that the DiagnosticFields class is instantiated after State. I did this because it needs the mesh... I could pass the mesh in but I'm hesitating because I'm not sure it's future-proof.

I also need to think about what happens if there are no diagnostic fields - maybe just have an empty diagnostic_fields_dict.


dcore/state.py
```diff
@@ -120,6 +121,10 @@ def dump(self):
             self.diagnostic_data = defaultdict(partial(defaultdict, list))
 
         if (next(self.dumpcount) % self.output.dumpfreq) == 0:
+
+            if 'Courant' in self.diagnostic_fields.diagnostic_field_dict.keys():
+                self.diagnostic_fields.Courant(self)
+
```
wence-: This smells of bad design.  You shouldn't be attempting to enumerate the diagnostic fields here.

wence-: I would maybe do something like this:

```class DiagnosticField(object):
        __metaclass__ = ABCMeta
   
        @abstractmethod
        def compute(self, state, *args):
             """Compute the diagnostic field from the current state and any extra arguments."""
             pass

class CourantNumber(DiagnosticField):

    def area(self, mesh):
        if hasattr(self, "_area"):
            return self._area
        V = FunctionSpace(mesh, "DG", 0)
        self._area = assemble(TestFunction(V)*dx)
        return self._area

    def field(self, mesh):
        if hasattr(self, "_field"):
            return self._field
        self._field = Function(FunctionSpace(mesh, "DG", 0), name="CourantNumber")
        return self._field

    def compute(self, state):
        u = state.field_dict['u']
        dt = Constant(state.timestepping.dt)
        return self.field().project(sqrt(dot(u, u))/sqrt(self.area())*dt)
```

wence-: Now you can do:

```python

diagnostic_fields = [CourantNumber(), SomeOtherDiagnostic()]

state = State(..., diagnostic_fields)
```

And then in your dump method:
```python

def dump(self, ...):
    for diag in self.diagnostic_fields:
        to_dump.append(diag.compute())
```

jshipton: Thanks @wence- that's a zillion times better!

wence-: Note, if you find yourself writing this pattern:

```python
    def field(self, mesh):
        if hasattr(self, "_field"):
            return self._field
        self._field = Function(FunctionSpace(mesh, "DG", 0), name=self.name)
        return self._field
```

Over and over again, we can add a memoize decorator to simplify it to:

```python
@memoize
def field(self, mesh):
    return Function(FunctionSpace(mesh, "DG", 0))
```

dcore/diagnostics.py
```diff
@@ -18,3 +19,43 @@ def register(self, *fields):
...
+            return self._area
+        V = FunctionSpace(mesh, "DG", 0)
+        self._area = assemble(TestFunction(V)*dx)
+        return self._area
+
+    def field(self, mesh):
+        if hasattr(self, "_field"):
+            return self._field
+        self._field = Function(FunctionSpace(mesh, "DG", 0), name=self.name)
+        return self._field
```
wence-: I note in passing that these are sort of "unsafe" in that if you call the methods twice with the same instance on different meshes you'll get the wrong answer second time round.  Similarly if you move the mesh, the area will be wrong.  For the area, I guess what you would like to do is just use `CellVolume`, but that is only currently supported on affine cells (that's firedrakeproject/tsfc#1)

colinjcotter: ok, let's construct self._area on the first call, and the UFL expression 
self.exp = TestFunction(V)*dx
and then always do
self._area.assemble(self.exp)
to update

jshipton: This failed with 'Coordinateless Function object has no attribute assemble' so I've just made it return assemble(self.expr) - looks like we don't need to create self._area, we could check for the existence of self.expr instead, but I'm not sure what the protocol is with these funny underscores...

wence-: So the underscore "hides" the slot, it's a "private, internal" thing.

The idea is to create the function once, and then reuse it every time so:

```python
if not hasattr(self, '_area'):
    self._area = Function(...)
assemble(self.exp, tensor=self._area)
return self._area
```

PR #35, opened 2016-05-05
Only try to dump diagnostics if you have diagnostics.

colinjcotter:
Only try to dump diagnostics if you have diagnostics.

### DIFF
--- a/dcore/state.py
+++ b/dcore/state.py
@@ -122,9 +122,10 @@ class State(object):
         if (next(self.dumpcount) % self.output.dumpfreq) == 0:
             self.dumpfile.write(*to_dump)
 
-            for name in self.diagnostics.fields:
-                data = self.diagnostics.l2(field_dict[name])
-                self.diagnostic_data[name]["l2"].append(data)
+            if(self.diagnostics != None):
+                for name in self.diagnostics.fields:
+                    data = self.diagnostics.l2(field_dict[name])
+                    self.diagnostic_data[name]["l2"].append(data)
 
     def diagnostic_dump(self):
 


jshipton:
I think this will still break if you have meanfields or if steady_state_dump_err is True. I made having diagnostics the default behaviour to avoid having `if diagnostics is not None` in several places, and also because it seems useful to have in general. We could make a basic minimal set of diagnostics in a base diagnostics class if you'd prefer that to setting `diagnostics=None` (there is only one diagnostic at present anyway!)

colinjcotter:
OK, then we need to *actually* make having diagnostics the default
behaviour. My example didn't execute!

On 5 May 2016 at 22:01, jshipton <notifications@github.com> wrote:

> I think this will still break if you have meanfields or if
> steady_state_dump_err is True. I made having diagnostics the default
> behaviour to avoid having if diagnostics is not None in several places,
> and also because it seems useful to have in general. We could make a basic
> minimal set of diagnostics in a base diagnostics class if you'd prefer that
> to setting diagnostics=None (there is only one diagnostic at present
> anyway!)
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/firedrakeproject/dcore/pull/35#issuecomment-217278733>
>



-- 

jshipton:
Sorry, looks like I forgot to update all the examples files then. I was copying the way that the timestepping, output and parameters classes are set up in the setup file - i.e. you have to pass them in - but with diagnostics it does make sense to have a default class.

jshipton:
Superseded but diagnostics_default.

PR #36, opened 2016-05-06
set default diagnostics class so code works when diagnostics are not …

jshipton:
…specified. Tested in test_dcmip_setup


colinjcotter:
fine by me.

PR #39, opened 2016-05-24
Dump latlon

jshipton:
This provides the option to dump fields on a latlon coordinate mesh. The user specifies fieldnames in  output.dumplist_latlon, which is set by default to be an empty list. This option has been added to the sw_williamson_triangle test but this test doesn't currently do anything more than set up the initial conditions.

Concerns:
- the dump method is becoming rather long - would it be a good idea to move the creation of the latlon mesh to another place?
- is there a way to check if the original mesh is actually a sphere?! (or is this overkill?)

### DIFF
--- a/dcore/configuration.py
+++ b/dcore/configuration.py
@@ -37,6 +37,7 @@ class OutputParameters(Configuration):
     Verbose = False
     dumpfreq = 1
     dumplist = None
+    dumplist_latlon = []
     dirname = None
     #: Should the output fields be interpolated or projected to
     #: a linear space?  Default is interpolation.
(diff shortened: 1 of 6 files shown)


colinjcotter:
I don't think that there is any need to check if the original mesh is actually a sphere!

colinjcotter:
I think that the lat-lon calculation should indeed be in a separate function.

jshipton:
I think this is good to go...

jshipton:
Test failure due to problem gunzipping h5py.tar.gz on travis. Merging.

PR #42, opened 2016-06-07
A working DCMIP 3.1 example with linear advection, and no mean flow.

colinjcotter:
Added a working example as per the title.

### DIFF
--- a/dcore/linear_solvers.py
+++ b/dcore/linear_solvers.py
@@ -122,24 +122,24 @@ class CompressibleSolver(TimesteppingSolver):
         k = state.k             # Upward pointing unit vector
         theta = -dot(k,u)*dot(k,grad(thetabar))*beta + theta_in
 
-        #Only include theta' (rather than pi') in the vertical
-        #component of the gradient
+        # Only include theta' (rather than pi') in the vertical
+        # component of the gradient
 
-        #the pi prime term (here, bars are for mean and no bars are
-        #for linear perturbations
+        # the pi prime term (here, bars are for mean and no bars are
+        # for linear perturbations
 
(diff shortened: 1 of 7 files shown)


examples/dcmip_3-1_linear_advection_quads_rt0.py
```diff
@@ -1,11 +1,11 @@
 from dcore import *
-from firedrake import IcosahedralSphereMesh, ExtrudedMesh, Expression, \
+from firedrake import CubedSphereMesh, ExtrudedMesh, Expression, \
     VectorFunctionSpace
 from firedrake import exp, acos, cos, sin
 import numpy as np
 
-nlayers = 10  # 10 horizontal layers
-refinements = 3  # number of horizontal cells = 20*(4^refinements)
+nlayers = 25  # 10 horizontal layers
```
jshipton: comment is now incorrect

dcore/linear_solvers.py
```diff
@@ -122,14 +122,26 @@ def _setup_solver(self):
         k = state.k             # Upward pointing unit vector
         theta = -dot(k,u)*dot(k,grad(thetabar))*beta + theta_in
 
+        # Only include theta' (rather than pi') in the vertical
+        # component of the gradient
+
+        # the pi prime term (here, bars are for mean and no bars are
+        # for linear perturbations
```
jshipton: missing bracket in comment

jshipton:
1) should we remove the other dcmip example file and drop the 'new' on the working one?
2) can the dcmip test be updated, firstly so that the setup is the same as the working example, and secondly so that it actually runs a couple of time steps?

jshipton:
I think I satisfied all my requirements! Merging...

PR #44, opened 2016-06-14
fix parallel race condition on creation of dumpdir

jshipton:
@wence- This seems to work but it didn't fail consistently in the first place. Have I imported the right thing to get comm?


wence-:
That looks right.

PR #45, opened 2016-06-15
Complete rename

wence-:
Finishing renaming dcore->gusto by also renaming package.

Also add setup.py.


jshipton:
Thanks @wence- this looks great!

PR #46, opened 2016-06-16
More diagnostics

jshipton:
This pull request adds diagnostic fields Divergence, Vorticity and Potential Vorticity and also adds min and max of diagnostic fields to the diagnostics dictionary.

Concerns:

1) Relies on conventions for accessing the correct spaces. Currently the spaces are stored in a list called state.V with the convention that u lives in state.V[0] and the pressure or density like variable lives in state.V[-1]. This could be improved by making state.V a dictionary and coming up with appropriate names for the spaces or I think we previously discussed having a get_space function which again would require appropriate names for the spaces.

2) The vorticity diagnostics are only for shallow water and won't be correct for the vertical slice. I should probably ShallowWaterDiagnostics class - does that sound like the right thing to do?


dorugeber:
Closing this to reduce clutter, feel free to bring the branch up to date and open a new PR.

PR #50, opened 2016-07-06
Change advection_list to advection_dict

jshipton:
This should make it possible to add advected/diffused tracer fields.
Not sure of the best way to add forcing for tracers.


colinjcotter:
This can merge.

PR #51, opened 2016-07-08
3d advection

colinjcotter:
Add code to do velocity advection in 3D.

Also, automatically build broken spaces for Embedded DG.

### DIFF
--- a/gusto/advection.py
+++ b/gusto/advection.py
@@ -227,12 +227,16 @@ class EulerPoincareForm(Advection):
 
             # <w,curl(u) cross ubar + grad( u.ubar)>
             #=<curl(u),ubar cross w> - <div(w), u.ubar>
-            #=<u,curl(ubar cross w)> + <<u_upwind, [[(ubar cross w)cross n]]>>
+            #=<u,curl(ubar cross w)> - 
+            #      <<u_upwind, [[n cross(ubar cross w)cross]]>>
+
+            both = 2*avg
 
             Eqn = (
                 (inner(w, u-self.u0)
-                 + dt*inner(ustar, curl( cross(ubar, w)))*dx
-                 - dt*inner(Upwind*ustar, 2*avg(cross(cross(ubar, w), n)))*surface_measure
+                 + dt*inner(ustar, curl(cross(ubar, w)))*dx
+                 - dt*inner(both(Upwind*ustar),
+                            both(cross(n, cross(ubar, w))))*surface_measure
                  - dt*div(w)*inner(ustar, self.ubar))*dx
             )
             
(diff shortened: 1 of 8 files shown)


jshipton:
Does this need firedrake to be updated to pass tests? How do we do that on travis?

colinjcotter:
Oh yes, that might be what's needed.

On 13 July 2016 at 13:28, jshipton <notifications@github.com> wrote:

> Does this need firedrake to be updated to pass tests? How do we do that on
> travis?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/firedrakeproject/gusto/pull/51#issuecomment-232340177>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AF0q8spzPHRA-wFm7mcOgPCq1BURl4BDks5qVNoFgaJpZM4JIHzs>
> .
>



-- 

wence-:
No, these are actual bugs.  To make a VectorFunctionSpace you need the `VectorElement` modifier on the outside of the element.  To make the `BrokenElement` you need to unwrap the passed in element and wrap it up again, I think.

gusto/advection.py
```diff
@@ -191,18 +192,30 @@ def apply(self, x_in, x_out):
 
 class EmbeddedDGAdvection(DGAdvection):
 
-    def __init__(self, state, Vdg, continuity):
+    def __init__(self, state, V, Vdg=None, continuity=False):
```
wence-: You've changed the constructor signature, but haven't updated the embedded dg advection tests, which still use the old signature.

wence-:
I think I fixed this.  Note that firedrake is always rebuilt for gusto on travis, so you don't have to worry about firedrake being out of date.

jshipton:
Don't you need the function space, rather than the function? And anyway, I still get the same error.

wence-:
Oh, maybe.  I don't know what the return values of all these things are.

wence-:
I think I've now fixed the remaining embeddedDG constructor calls.

wence-:
Green and merged.

jshipton:
Hang on a minute! The new code isn't actually tested... I'm just sorting that out now.

wence-:
Too late :p.  Just make a new branch adding tests.

jshipton:
:P


I'm getting muddled using xfail with parameterise. At the moment I have:


@pytest.mark.parametrize("vector", [False, True])

@pytest.mark.parametrize("continuity", [False, True])

@pytest.mark.parametrize("broken", [False, True])


and I want to mark the ones with "broken" = True and "vector" = True as xfails.


I've tried amending the last parameterise to:


@pytest.mark.parametrize("broken", [False, pytest.mark.xfail("vector" is True, reason="not implemented")(True)])


but it seems to be ignoring the xfail mark - maybe it doesn't know what "vector" is yet?


Any ideas?


J

wence-:
There's no way to access the parameter value inside the parametrize decorator.  Here might be one way:

```python
import pytest

params = [False, True]

@pytest.fixture(params=params)
def vector(request):
    return request.param

@pytest.fixture(params=params)
def continuity(request):
    return request.param

@pytest.fixture(params=params)
def broken(request, vector):
    if request.param and vector:
        request.applymarker(pytest.mark.xfail(reason="foo"))
    return request.param

def test_foo(vector, continuity, broken):
    if vector and broken:
        assert False
    else:
        assert True
```

PR #58, opened 2016-09-14
Reset mu alpha to zero at beginning of step.

colinjcotter:
Reset mu alpha to zero at beginning of each step,
otherwise it will use the last value that was set.


colinjcotter:
I think this can be merged once the conflicts are resolved.

PR #59, opened 2016-09-17
label solvers

colinjcotter:
Label the main linear solvers to aid debugging.


colinjcotter:
I'm merging this since it passes and should be uncontroversial.

PR #60, opened 2016-09-19
Checkpointing for gusto

colinjcotter:
Checkpointing for gusto.

State fields, plus init fields and mean fields are all checkpointed.

Will pick up from checkpointing file in timestepper.run() if pick_up=True,
and will reset the time value from the checkpointing file.


colinjcotter:
I have finished tinkering with this now.

jshipton:
Looks good to me and doesn't break any tests so I'm happy.

colinjcotter:
cheers, closing.

PR #67, opened 2016-11-08
Test examples

jshipton:
Hooks up automatic testing for example files. Failing dcmip examples have been removed and all other adjusted to run for just one timestep when called from pytest.

With many thanks to @wence-


dorugeber:
Are there working DCMIP examples around still...?

jshipton:
Yeah, Colin has some which he'll add at some point. These examples are not what he's been running from anyway.

Although Colin, if you want to pick your favourite you can add it now and I'll sort out the test.

PR #69, opened 2016-11-17
Fix diffusion

jshipton:
This pull request fixes some issues with the InteriorPenalty diffusion method.

1) kappa and mu are now keyword arguments as they must be provided
2) added optional bcs argument so that we can apply to correct no normal flow boundary conditions for the dense bubble test
3) removed the directions argument that was used to specify the direction in which the field was discontinuous so that only the surface terms in that direction were added - instead we now assemble all the surface terms and those in the continuous direction are zero. This is less confusing for the user.

### DIFF
--- a/examples/dense_bubble.py
+++ b/examples/dense_bubble.py
@@ -80,7 +80,7 @@ for delta, dt in res_dt.iteritems():
     advection_dict["u"] = EulerPoincareForm(state, state.V[0])
     advection_dict["rho"] = DGAdvection(state, state.V[1], continuity=True)
     advection_dict["theta"] = SUPGAdvection(state, state.V[2], direction=[1])
-    #advection_dict["theta"] = EmbeddedDGAdvection(state, state.V[2], Vdg=Vtdg, continuity=False)
+    # advection_dict["theta"] = EmbeddedDGAdvection(state, state.V[2], Vdg=Vtdg, continuity=False)
 
     # Set up linear solver
     schur_params = {'pc_type': 'fieldsplit',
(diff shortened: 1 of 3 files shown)


tests/test_IP_diffusion.py
```diff
@@ -55,19 +55,14 @@ def run(dirname, vector, DG):
...
-    if DG:
-        direction = [1,2]
-    else:
-        direction = [1]
-
-    kappa = 0.05
+    kappa = Constant(0.05)
     if vector:
         kappa = as_tensor([[kappa, 0.],[0., kappa]])
     mu = 5.
```
dorugeber: any reason this isn't also a `Constant`?

examples/dense_bubble.py
```diff
@@ -112,8 +112,11 @@
     # Set up forcing
     compressible_forcing = CompressibleForcing(state)
 
-    diffusion_dict = {"u": InteriorPenalty(state, state.V[0],direction=[2], params={"kappa":Constant(75.), "mu":Constant(1./delta)}),
-                      "theta": InteriorPenalty(state, state.V[2],direction=[2], params={"kappa":75., "mu":0.005})}
+    V = state.V[0]
+    bcs = [DirichletBC(V, Expression(("0.0","0.0")), "bottom"),
+           DirichletBC(V, Expression(("0.0","0.0")), "top")]
```
dorugeber: I think `Expression(("0.0", "0.0"))` is misleading; I'd rather see 0.0 here (you're setting the DoF value, not the field)

tests/test_IP_diffusion.py
jshipton: kappa = Constant(as_tensor(...)) gives a ValueError: setting an array element with a sequence.

wence-: `kappa = Constant([[0.05, 0], [0, 0.05]])`

dorugeber: I was referring to `mu`.

Though, do you plan to use a non-diagonal `kappa` at some point?  If not, why not just multiply by a scalar kappa rather than by the matrix "kappa*I"?

jshipton:
Oh, it's the constant constants that's the problem, not the tensor - I see.

tests/test_IP_diffusion.py
jshipton: @dorugeber yes, possible that we will need non diagonal kappa at some point

PR #70, opened 2016-12-12
Code restructure

jshipton:
The point of this pull request is to separate the time stepping scheme used for advection from the spatial discretisation of the advection operator. Previously the code was set up to use Forward Euler for linear advection, SSPRK3 with a DG discretisation, and Implicit Midpoint for the momentum equation, which is fine but not very flexible. This setup also made mesh movement coding tricky as it would have to be coded separately in the apply method of each advection class.

The user still has to set up an advection_dict, containing the advection schemes for each field. Previously this had entries of the form:
`advection_dict["rho"] = DGAdvection(state, state.V[1], continuity=True)`
This is replaced by:
`rhoeqn = Advection(state, state.V[1], continuity=True)`
`advection_dict["rho"] = SSPRK3(state, rho0, rhoeqn)`
where Advection is a class defined in transport_equation.py

This makes the diff massive because every example and test needed changing!

I believe this pull request preserves the behaviour currently expected in master. In addition to the flexibility described above it also has the following new features:

- 'no_vort' discretisation of the vector invariant form of the momentum equation (as used in the old slicemodel code) which discretises the cross(curl(u), u) term directly i.e. without computing the vorticity, hence the name but better name suggestions appreciated - in the code this is the standard VectorInvariant class in transport_equation.py
- new time stepping class (in timeloop.py) that just does advection - this makes testing the advection schemes easier, which is why it is in this pull request.
- embedded DG advection can now be used with any advection time stepping scheme via the decorator in advection.py. Currently the decorator is only applied to the method for SSPRK3 advection because this is all that I have tested. Ideally I would like to apply it to the abstract apply method in the base class and @wence- kindly showed me a couple of ways to do this, but first I wanted to check if a decorator is an acceptable way to go - I'm very pleased with it as it is the first one I've ever coded but I understand that doesn't necessarily make it good coding practise!

Additional changes:

- slight reorganisation of forcing.py so that user can pass the arguments `linear` or `euler_poincare` because the discretisation of the momentum equation decides whether a term needs to be added to the forcing.
- clearer SUPG options (and consequent bug fixes!)
- new example (sk_nonlinear_3d) since dcmip tests were nonfunctional and I wanted to check that this restructure worked for 3d equations.
- reorganisation of advection tests into one file to avoid code duplication.

Notes:

- Tests are now a minute longer (taking about 16mins in total on my laptop) which may be too long - might be worth splitting them up in future.
- update_solver method in Advection base class might seem to have a strange name as it actually creates the solver - this is because I envisage using it in the very near future to update the solver at each timestep when the mesh is moving because this will cause the lhs and rhs to change.


gusto/transport_equation.py
```diff
@@ -0,0 +1,345 @@
...
+            space = Vdg
+        self.xdg_out = Function(space)
+        self.x_projected = Function(V)
+        parameters = {'ksp_type':'cg',
+                      'pc_type':'bjacobi',
+                      'sub_pc_type':'ilu'}
+        self.Projector = Projector(self.xdg_out, self.x_projected,
+                                   solver_parameters=parameters)
+
+        super(EmbeddedDGAdvection, self).__init__(state, space, ibp)
```
wence-: Miklos would say "call the superclass init first, so we're sure that there's no cross-contamination (in terms of needing to have set stuff up").  i.e. move this `super` call to after the space creation.

dorugeber:
Any particular reason the tests take longer?  Did something change?

examples/nh_mountain.py
```diff
@@ -23,7 +23,12 @@
 H = Constant(H)
 a = Constant(1000.)
 xc = Constant(L/2.)
-new_coords = Function(Vc).interpolate(as_vector([x[0], x[1]+(H-x[1])*a**2/(H*((x[0]-xc)**2+a**2))]))
+smooth_z = True
+if smooth_z:
+    xexpr = Expression(("x[0]","x[1] < zh ? x[1]+pow(cos(0.5*pi*x[1]/zh),6)*h*pow(a,2)/(pow(x[0]-xc,2)+pow(a,2)) : x[1]"), zh=5000., h=1., a=a, xc=xc)
```
wence-: May as well introduce ufl interpolation here too:

```python
x, y = SpatialCoordinate(mesh)
as_vector([x, conditional(y < zh, y + pow(cos(0.5*pi*y/zh), 6)*h*pow(a, 2) / (pow(x - xc, 2) + pow(a, 2)), y)])
```
I think.

examples/dense_bubble.py
```diff
@@ -76,12 +76,20 @@
...
+                                 continuity=False)
+    else:
+        thetaeqn = EmbeddedDGAdvection(state, state.V[2],
+                                       continuity=False)
     advection_dict = {}
-    advection_dict["u"] = EulerPoincareForm(state, state.V[0])
-    advection_dict["rho"] = DGAdvection(state, state.V[1], continuity=True)
-    advection_dict["theta"] = SUPGAdvection(state, state.V[2], direction=[1])
-    # advection_dict["theta"] = EmbeddedDGAdvection(state, state.V[2], Vdg=Vtdg, continuity=False)
+    advection_dict["u"] = ThetaMethod(state, u0, ueqn)
```
dorugeber: I assume 'theta' is hidden inside `state`?

gusto/__init__.py
```diff
@@ -3,6 +3,7 @@
 from gusto.advection import *       # noqa
 from gusto.diagnostics import *     # noqa
 from gusto.diffusion import *     # noqa
+from gusto.transport_equation import *     # noqa
```
wence-: advection.py and transport_equation.py both define `Advection` classes that are imported into the global namespace here.  That's a bit odd... (the latter overrides the former).

gusto/timeloop.py
```diff
@@ -42,7 +32,8 @@ def _set_ubar(self):
         unp1 = state.xnp1.split()[0]
 
         for field, advection in self.advection_dict.iteritems():
-            advection.ubar.assign(un + state.timestepping.alpha*(unp1-un))
+            if not isinstance(advection, NoAdvection):
+                advection.ubar.assign(un + state.timestepping.alpha*(unp1-un))
```
wence-: This kind of instance check is bad code smell.  The advection object should have a method to update `ubar` given some expression for the new `ubar`, and then `NoAdvection` can just implement a no-op there.  Then the caller doesn't need to know about what the object type is.

i.e., something like:

```python
class Advection(object):
    ...
    def update_ubar(self, expr):
        self.ubar.assign(expr)

class NoAdvection(Advection):
    def update_ubar(self, expr):
        return
```

gusto/timeloop.py
```diff
@@ -125,5 +147,50 @@ def run(self, t, tmax, pickup=False):
                 state.dump(t, pickup=False)
 
         state.diagnostic_dump()
-
         print "TIMELOOP complete. t= "+str(t-dt)+" tmax="+str(tmax)
```
wence-: In parallel, these kind of prints are pretty unfriendly.  If you do:

```python
from firedrake.petsc import PETSc

PETSc.Sys.Print("some string", comm=self.state.mesh.comm)
```

Then you'll only print on rank zero of the provided communicator.

gusto/transport_equation.py
```diff
@@ -0,0 +1,345 @@
...
+    :arg V: :class:`.FunctionSpace object. The function space that q lives in.
+    :arg qbar: The reference function that the equation has been linearised
+               about. It is assumed that the reference velocity is zero and
+               the ubar below is the nonlinear advecting velocity
+               0.5*(u'^(n+1) + u'(n)))
+    :arg ibp: string, stands for 'integrate by parts' and can take the value
+              None, "once" or "twice". Defaults to "once".
+    :arg continuity: is the equation in continuity form L(q) = div(u'*qbar)?
+                     If False, the equation is assumed to be in advective
+                     form L(q) = u' dot grad(qbar)
```
wence-: For ease of readability, I'd be tempted to change this flag into an argument:  `equation_form` which takes the value "continuity" or "advective".

gusto/transport_equation.py
```diff
@@ -0,0 +1,345 @@
...
+        super(Advection, self).__init__(state, V, ibp)
+        self.continuity = continuity
+
+    def advection_term(self, q):
+
+        if self.continuity:
+            if self.ibp is "once":
+                L = -inner(grad(self.test), outer(q, self.ubar))*dx
+            else:
+                L = inner(self.test, div(outer(q, self.ubar)))*dx
```
wence-: This branch can never be hit, right?  Because in the superclass constructor you ensure that ibp is `once` if continuity is True.

gusto/transport_equation.py
```diff
@@ -0,0 +1,345 @@
...
+        else:
+            if self.ibp is "once":
+                L = -inner(div(outer(self.test,self.ubar)),q)*dx
+            else:
+                L = inner(outer(self.test,self.ubar),grad(q))*dx
```
wence-: The once case can't happen here, right?  If I follow the logic in the superclass constructor, I think there are the following cases:

`continuity == True`: asserts `ibp == once`
`continuity == False`: asserts `ibp == None`

Hence the value of `continuity` completely specifies the value of `ibp`.

gusto/transport_equation.py
wence-: Oh wait, no, that was only in the `LinearAdvection` case.

gusto/transport_equation.py
wence-: Ignore me...

gusto/advection.py
```diff
@@ -1,34 +1,97 @@
...
+        # setup required functions
+        self.dq = Function(fs)
+        self.q1 = Function(fs)
+
+        # get ubar from the equation class if provided
+        self.equation = equation
+        if equation is not None:
+            self.ubar = self.equation.ubar
+
+    def update_solver(self):
```
wence-: Better called `create_solver`?

gusto/advection.py
```diff
@@ -1,34 +1,97 @@
...
+        # setup solver using lhs and rhs defined in derived class
+
+        problem = LinearVariationalProblem(self.lhs, self.rhs, self.dq)
+        solver_name = self.field.name()+self.equation.__class__.__name__+self.__class__.__name__
+        self.solver = LinearVariationalSolver(problem, solver_parameters=self.solver_parameters, options_prefix=solver_name)
```
wence-: I would be tempted to make this as a `cached_property`:

```python
from firedrake.utils import cached_property

class Foo(...):
    
    @cached_property
    def solver(self):
        problem = ...
        name = ...
        return LinearVariationalSolver(problem, ...)
```

At the first reference of `self.solver`, the solver is created.  It is then subsequently cached on the object, so subsequent references just return the same thing.  Then the subclasses don't need to call `update_solver` ever.  It is just made when needed.

gusto/advection.py
```diff
@@ -1,34 +1,97 @@
...
+
+        # get ubar from the equation class if provided
+        self.equation = equation
+        if equation is not None:
+            self.ubar = self.equation.ubar
+
+    def update_solver(self):
+        # setup solver using lhs and rhs defined in derived class
+
+        problem = LinearVariationalProblem(self.lhs, self.rhs, self.dq)
```
wence-: `self.lhs` and `self.rhs` come from subclasses, but it is not documented in the superclass.  So maybe:

```python
class Advection(...):

    ...
    @abstractproperty
    def lhs(self):
        ...

    @abstractproperty
    def rhs(self):
        ...

class ForwardEuler(Advection):

    @cached_property
     def lhs(self):
          return ...

    @cached_property
    def rhs(self):
        return ....
```

That way, except for the theta method, you don't even need to override the constructor.

gusto/advection.py
```diff
@@ -1,34 +1,97 @@
...
+        # equation class and we can get the correct function space from
+        # the output function.
+        if hasattr(equation, "Projector"):
+            fs = equation.xdg_out.function_space()
+            self.Projector = equation.Projector
+            self.xdg_out = equation.xdg_out
+            self.xdg_in = Function(fs)
+            self.x_projected = equation.x_projected
+        else:
+            fs = field.function_space()
```
wence-: There's a lot of magic "action-at-distance" involved in this determination.  You inspect some "internal" property of the advection equation object to determine if you're going to do embedded dg.  As commented below, I think the `equation` object shouldn't make the projector and so forth.  Instead the `Advection` object should.  So I would set, on the equation instances a property `embedded_dg` to True if it's `EmbeddedDG` and False if it's not.  Then (maybe here) you build the projector and the DG fields.

Right now, `SSPRK3` is the only scheme that implemented the embedded_dg advection.  I wonder as well whether this logic should move there.  I guess you'd need to do some refactoring to remove setting up the functions further down (turn them into `cached_property` s?).

gusto/advection.py
jshipton: yes, at the moment, but I expect to use it for moving meshes to update the solver (because moving the mesh changes lhs and rhs)- I'm happy to change the name here if you like but I'm only going to change it back pretty soon... (unless you can think of a name that covers both cases!)

gusto/timeloop.py
jshipton: cool - but I might make that a separate issue because it'll make me go through and check whether it'd be useful to have print statements elsewhere for new users

gusto/timeloop.py
jshipton: yes, I like that

examples/dense_bubble.py
jshipton: no, you can pass it in but it defaults to 0.5 because that's what I always use - would you rather it had to be explicitly set?

examples/nh_mountain.py
jshipton: yes, Andrew told me about this uff conditional yesterday - awesome! I will make a separate issue and convert all the Expressions to ufl interpolations - they appear all over the place at the moment.

gusto/advection.py
```diff
@@ -1,34 +1,97 @@
...
+        if solver_params is None:
+            if isinstance(equation, LinearAdvection):
+                self.solver_parameters = {'ksp_type':'cg',
+                                          'pc_type':'bjacobi',
+                                          'sub_pc_type': 'ilu'}
+            else:
+                self.solver_parameters = {'ksp_type':'preonly',
+                                          'pc_type':'bjacobi',
+                                          'sub_pc_type': 'ilu'}
+        else:
```
jshipton: nope, you're right, that should go in advection.py where the solver is created

gusto/transport_equation.py
```diff
@@ -0,0 +1,345 @@
...
+        # default assumes a continuous space
+        supg_params.setdefault('dg_direction', None)
+
+        # find out if we need to do DG upwinding in any direction and set
+        # self.dS accordingly
+        if supg_params["dg_direction"] is None:
+            # space is assumed to be continuous and we don't need
+            # any interior surface integrals
+            self.dS = None
+        elif supg_params["dg_direction"] is "horizontal":
```
jshipton: oh, thanks - will fix

gusto/transport_equation.py
jshipton: that make sense - thanks, will fix

gusto/__init__.py
jshipton: hm, maybe I should call one of them something different

examples/dense_bubble.py
dorugeber: From a maths perspective, I find it a bit weird, but meh, not too important.

gusto/advection.py
jshipton: And this is why I like having you review my pull requests - I find out new stuff! How would this work with mesh movement when the solver does need to change?

gusto/transport_equation.py
```diff
@@ -0,0 +1,345 @@
...
+            space = FunctionSpace(state.mesh, V_elt)
+        else:
+            space = Vdg
+        self.xdg_out = Function(space)
+        self.x_projected = Function(V)
+        parameters = {'ksp_type':'cg',
+                      'pc_type':'bjacobi',
+                      'sub_pc_type':'ilu'}
+        self.Projector = Projector(self.xdg_out, self.x_projected,
+                                   solver_parameters=parameters)
```
jshipton: yes, I think you're right

gusto/advection.py
wence-: so what does moving the mesh change?  Does it change the symbolic equation, or just the values in it (e.g. the coordinate field)?

gusto/advection.py
jshipton: Yes, I agree with your first paragraph. I don't want to move the embedded dg stuff to SSPRK3 because I want to be able to use it with the other advection schemes - I just haven't tested it yet (although see no reason why it shouldn't work). I'd prefer to decorate the abstract apply method but I'm a little wary of doing that without testing all the options.

gusto/advection.py
jshipton: ok

wence-: Again, in combination with mesh movement, if that changes the the lhs/rhs in some symbolic way, then make these `property` not `cached_property` (they're only accessed each time you make the solver, so that's OK).

gusto/advection.py
jshipton: it changes the coordinate field. I have something sketched out here: https://github.com/firedrakeproject/gusto/blob/move_mesh/gusto/advection.py

jshipton:
Separating the spatial discretisation from the advection scheme means there are more combinations - e.g. DG + ThetaMethod - that weren't possible before.

dorugeber:
Okay, so it's just testing more combinations, rather than some unexplained slowdown.

examples/dense_bubble.py
colinjcotter: I think it's fine.

PR #73, opened 2016-12-16
Incompressible Eady model works up to the front formation and blows up

hiroe-:
New classes EadyParameters and EadyForcing are added for the incompressible Eady test.

New example files using PeriodicRectangleMesh for the gravity wave test and the incompressible Eady test.

The Eady model runs up to day 5 where the result looks very similar to that from the old model using slicemodels. However, it blows up soon after the front formation starts.


gusto/forcing.py
```diff
@@ -228,6 +228,121 @@ def apply(self, scaling, x_in, x_nl, x_out, **kwargs):
...
+        self.u_forcing_solver.solve()  # places forcing in self.uF
+        self.b_forcing_solver.solve()  # places forcing in self.bF
+
+        u_out, p_out, b_out = x_out.split()
+
+        x_out.assign(x_in)
+        u_out += self.uF
+        b_out += self.bF
+
+        if 'incompressible' in kwargs and kwargs['incompressible']:
```
wence-: AKA `kwargs.get("incompressible", False)`

hiroe-:
It looks like something goes wrong at the front near the top and bottom boundaries in the u field.

horizontal-velocity
out-of-slice-velocity
vertical-velocity

examples/nonlinear_eady.py
```diff
@@ -0,0 +1,168 @@
...
+# set up all the other things that state requires
+##############################################################################
+# Spaces for initialising z, k and velocity
+W_VectorCG1 = VectorFunctionSpace(mesh, "CG", 1)
+W_CG1 = FunctionSpace(mesh, "CG", 1)
+
+# vertical coordinate and normal
+x = SpatialCoordinate(mesh)
+z = Function(W_CG1).interpolate(x[2])
+k = Function(W_VectorCG1).interpolate(Expression(("0.","0.","1.")))
```
wence-: Is there a reason to do this interpolation?  Why not just

```python
x, y, z = SpatialCoordinate(mesh)
k = Constant([0, 0, 1])
```

jshipton:
What boundary conditions were you applying in the slicemodels code?

examples/nonlinear_eady.py
```diff
@@ -0,0 +1,168 @@
...
+bref = x[2]*Nsq
+# interpolate the expression to the function
+b_b = Function(state.V[2]).interpolate(bref)
+
+# setup constants
+a = -7.5
+Bu = 0.5
+template_s = inifns.template_target_strings()
+b_exp = Expression(template_s,a=a,Nsq=Nsq,Bu=Bu,H=H,L=L)
+b_pert = Function(state.V[2]).interpolate(b_exp)
```
wence-: You can [interpolate UFL expressions](http://firedrakeproject.org/interpolation.html#ufl-expressions), so I think there's no need for this sympy stuff.   Something like:

```python
def coth(x):
    return cosh(x)/sinh(x)
x, y, z = SpatialCoordinate(mesh)
b_pert = Function(state.V[2]).interpolate(tanh(x) * coth(x))
```

for example (I think your sympy code just just translate directly.

examples/gw_incompressible_3d.py
```diff
@@ -0,0 +1,162 @@
+from gusto import *
```
jshipton: Do we need another example file? What does this test that sk_nonlinear_3d doesn't?

jshipton: Good point - it's on a branch! If you don't mind, I'd rather keep sk_nonlinear_3d and drop this one, purely because sk_nonlinear_3d is already set up to fit with the code_restructure whereas I'd have to change this one.

jshipton:
yes, this is something I've cleaned up a bit in the code_restructure branch. My plan is to get Hiroe's PR in first, then code_restructure, so I can sort it out there, if that's OK.

examples/gw_incompressible_3d.py
hiroe-: Of course! gw_incompressible_3d is removed

jshipton:
This passes tests and looks good to merge!

PR #75, opened 2017-01-16
State restructure

jshipton:
This pull request restructures the Gusto state. The main advantage is that there is now just one state class with no need to subclass for different equation sets or geometries. The function spaces are built depending on the user input `horizontal_degree`, `vertical_degree` and `family`.

Other features:

1.  Spaces and fields are defined using the new SpaceCreator and FieldCreator classes. This should fix checkpointing bugs and make it easier to get functions and spaces as they can be referred to by name. Note that:
- Spaces: I have tried to be consistent in accessing spaces via field.function_space() if the field is available, or using state.spaces.space_name otherwise. I'm not at all attached to the space names so if anyone can come up with something better that'd be great - I've used `HDiv` for the velocity space, `HDiv_v` for the scalar space with the same dofs as the vertical (hence the _v) component of the velocity space, and `DG` for the corresponding DG space (i.e. that used for pressure, density or height). 
- Fields: I have reserved state.xn for just those fields involved in the nonlinear iteration. These are created when the class is instantiated. Initial conditions and reference profiles are set by providing a dictionary of the form {name: function}, e.g. in the setup file:
```u0 = state.fields.u
u0.project(as_vector([10., 0., 0.]))
b0 = state.fields.b
b0.interpolate(some_expression)
state.initialise({'u': u0, 'b': b0})
```
The initialisation assumes the function is on the same function space. For reference profiles you can pass in a function on a different function space, only because we have had that situation in the past - I'm not sure it is still necessary - @colinjcotter ?
- each field has a `dump` attribute and a `pickup` attribute which decides whether it is dumped/picked up. Default is true in both cases.
- Additional fields (e.g. passive tracers) can be added in the setup file by calling the state.fields class:
`tracer = state.fields('tracer', some_function_space)`
and initialised by adding them to the dictionary of initial conditions. You can then define an advection or continuity equation for this field in the advection_dict and it will be picked up in the time loop, put in a list called passive_fieldlist, and advected outside of the nonlinear iteration loop. @colinjcotter @tommbendall you might want to call passive_fieldlist something different if you are planning to add forcing terms.

2. New setup_dump method which is called before the first call of dump. This sets up the diagnostic fields and lists of fields to dump and pickup.

3. New diagnostic field called Difference that calculates field1 - field2. Subclassed by Perturbation and SteadyStateError - makes these fields easier to setup.

4. perp is now defined in state (as it depends on the geometry) so that we don't have to define it in multiple places

5. I've deleted a couple of tests that were only checking setup - they were duplicating the examples so the setup is now tested when we test the examples.

6. I've updated all examples and tests.

To Do:
1. I need to update comments and docstrings.
2. @tommbendall, @colinjcotter I propose we remove examples/tracer.py (as it is so similar to the dense bubble example) and instead have a proper test that runs the tracer code for 'enough' time steps and checks that the tracer is the same as the potential temperature.
3. (@wence-) there are a couple of test failures that I don't understand at all.
- the shallow water tests (test_sw_linear_triangle.py and test_sw_triangle.py) complain that I am trying to dump fields that are not on the same mesh. Unfortunately this error doesn't always occur on my laptop so I'm having trouble pinning it down - I suspect it has something to do with the way I'm setting up the steady state error fields.
- the examples that use the IncompressibleSolver (gw_incompressible.py and nonlinear_eady.py) don't converge. This appears to be a problem with the spaces. I can make it go away by accessing the spaces in the old way (via a list). In particular it seems to be the velocity space that is the problem. See the branch weird_spaces_bug and run `python examples/gw_incompressible.py` and it should fail on the second linear solve. If you comment out line 253 of linear_solvers.py and uncomment line 256 it should work - all this is doing is switching between the new style space setup and the old style space setup. I don't understand this at all!!


jshipton:
@wence- just noticed something very strange about the shallow water test failures. In the output (pasted below) the functions are written out (I'm not sure which bit of code does this) and I noticed it says "ExtrudedMesh" and "quadrilateral" but these tests are run on IcosahedralSphere meshes so are neither extruded nor quadrilateral.

self = <firedrake.output.File object at 0x7f05feec3890>
functions = (Coefficient(WithGeometry(IndexedProxyFunctionSpace(<firedrake.mesh.MeshTopolog...ement(FiniteElement('Lagrange', Cell...ionSpace(<firedrake.mesh.ExtrudedMes...l, 1), cell=TensorProductCell(quadrilateral, interval)), dim=3), 3305)), 45524))
Function = <class 'firedrake.function.Function'>
f = Coefficient(WithGeometry(IndexedProxyFunctionSpace(<firedrake.mesh.ExtrudedMes...l, 1), cell=TensorProductCell(quadrilateral, interval)), dim=3), 3305)), 45524)
    def _write_vtu(self, *functions):
        from firedrake.function import Function
        for f in functions:
            if not isinstance(f, Function):
                raise ValueError("Can only output Functions, not %r" % type(f))
        meshes = tuple(f.ufl_domain() for f in functions)
        if not all(m == meshes[0] for m in meshes):
>           raise ValueError("All functions must be on same mesh")
E           ValueError: All functions must be on same mesh

wence-:
huh, weird.  Do you have any global variables hanging around?  Does this happen if you run *just that test* locally?

jshipton:
I haven't knowingly made any global variables! The test doesn't always fail on my laptop - even when run as part of make test.

jshipton:
@wence- I just had the experience that during running make test the shallow water tests failed with the 'All functions must be on same mesh' error, but then when I run them individually they pass. Fun. Are you around at all tomorrow to take a look at this with me?

gusto/linear_solvers.py
```diff
@@ -136,14 +139,18 @@ def _setup_solver(self):
         def V(u):
             return k*inner(u,k)
 
+        # specify degree for some terms as estimated degree is too large
+        dxp = dx(degree=(5,5))
```
colinjcotter: Please can we have a smarter way of choosing this?

jshipton: You mean make it an optional argument to the linear solver? I chose (5,5) based on the automatic quadrature degree that ufl was choosing, and also as it is what we used to use the slice models code. I'm just updating the docstrings in this file and I can't remember why degree=(5,5) and not just 5 - can you? or @wence- ?

wence-: Extruded mesh so vertical and horizontal degree. 

I think Colin means, check the degree of the spaces and pick an appropriate degree that way: this under integrates a Q3 mass matrix, for example

jshipton:
ah, ok, I was just thinking of the spaces we're currently using. Not sure exactly how to choose this for other spaces - it's the terms involving pi that bother me.

gusto/linear_solvers.py
jshipton: I don't use dxp for the mass matrices - only for the terms involving pi

wence-: the point is more that if you change gusto in the future to support higher than RT1 (say), this integration degree will be wrong.  Can you determine the value to put in programmatically by asking for the degree of the function spaces in the pi term?

jshipton:
Sure, I understand the point, but I don't know how to work out value given the degrees of the spaces. The pi function isn't very nice - I remember the decision of (5, 5) being "mumble mumble, probably enough" - am I missing something?

wence-:
So I think in the past in discussion it's been "if you can integrate a mass matrix exactly you don't negatively affect convergence".  I don't know if that would affect balance.

PR #77, opened 2017-01-24
incompressible hydrostatic initialisation code

jshipton:
Code for hydrostatic initialisation for the incompressible Euler equations. Seems to work - I ran the incompressible gravity wave without the initial perturbation and nothing happened, as expected.

### DIFF
--- a/gusto/initialisation_tools.py
+++ b/gusto/initialisation_tools.py
@@ -41,7 +41,7 @@ def incompressible_hydrostatic_balance(state, b0, p0, top=False, params=None):
     v, pprime = TrialFunctions(WV)
     w, phi = TestFunctions(WV)
 
-    bcs = [DirichletBC(WV[0], Expression(bc), bstring)]
+    bcs = [DirichletBC(WV[0], bc, bstring)]
 
     a = (
         inner(w, v) + div(w)*pprime + div(v)*phi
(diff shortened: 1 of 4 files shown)


gusto/initialisation_tools.py
```diff
@@ -5,10 +5,70 @@
...
+
+def incompressible_hydrostatic_balance(state, b0, p0, top=False, params=None):
+
+    # get F
+    v = TrialFunction(state.Vv)
+    w = TestFunction(state.Vv)
+
+    unp1 = state.xnp1.split()[0]
+    dim = unp1.ufl_element().value_shape()[0]
+    bc = ("0.0",)*dim
```
wence-: Can just use `zero(unp1.ufl_shape)`, then no need for `Expression(bc)` in the bcs below.

gusto/initialisation_tools.py
```diff
@@ -5,10 +5,70 @@
...
+                         'pc_fieldsplit_type': 'schur',
+                         'pc_fieldsplit_schur_fact_type': 'full',
+                         'pc_fieldsplit_schur_precondition': 'selfp',
+                         'fieldsplit_1_ksp_type': 'preonly',
+                         'fieldsplit_1_pc_type': 'gamg',
+                         'fieldsplit_1_mg_levels_pc_type': 'bjacobi',
+                         'fieldsplit_1_mg_levels_sub_pc_type': 'ilu',
+                         'fieldsplit_0_ksp_type': 'richardson',
+                         'fieldsplit_0_ksp_max_it': 4,
+                         'ksp_atol': 1.e-08,
```
wence-: Do you really want `atol` of 1e-8?

tests/test_gw_incompressible.py
```diff
@@ -1,15 +1,24 @@
...
-    VectorFunctionSpace, PeriodicIntervalMesh, ExtrudedMesh, \
-    sin
-import numpy as np
+from firedrake import SpatialCoordinate, \
+    VectorFunctionSpace, PeriodicRectangleMesh, ExtrudedMesh, \
+    sin, pi
+
+
+def max(f):
+    fmax = op2.Global(1, [-1000], dtype=float)
```
wence-: Is `f` guaranteed to be positive?  Safer to use `numpy.finfo(float).min`.

gusto/initialisation_tools.py
jshipton: @hiroe- @colinjcotter I'm not sure what settings you'd like here...

hiroe-: I don't think atol of 1e-8 is necessary. It is one of legacy settings in the old code.

PR #83, opened 2017-02-02
Drone CI support

wence-:
Since gusto is getting quite close to the 50 minute travis timeout, here's a configuration that tests on the local drone build system (no timeouts).


wence-:
Dead, since we don't have drone.

PR #88, opened 2017-03-02
Tracer test

tommbendall:
The intention of this branch was to create a test for the advection of inert tracers (those with no physics).

I have called the test "weightless_tracer" to emphasise that the concentration of this tracer does not appear in the momentum equation. The test is based on a simple rising bubble experiment, in which the tracer field is initially equal to the potential temperature (theta) field. The test is therefore that the difference between the advected tracer and theta fields is very small.

Because of this, I have also slightly extended the special diagnostics that are output in state.py.


jshipton:
I think this looks great. Note that you don't have to define a difference fields list - you can just define a single diagnostic field in your test file - but I'm happy to do it this way.

tommbendall:
Ah, I was looking to do it that way but couldn't see how to...

jshipton:
I think it should work if you set

diagnostic_fields = [Difference('theta', 'tracer')]

and pass that in to state.

jshipton:
P.S. looks like you need to pull from master before merging...

jshipton:
This looks perfect - thanks Tom!

Issue #78, opened 2017-01-26
Fix off-by-one error


jshipton:
As Andrew has pointed out, we are running for one too many time steps.

colinjcotter:
This is because we check t < tmax + 0.5*dt when it should be t < tmax - 0.5*dt

PR #94, opened 2017-05-07
fix off-by-one timestep error

dorugeber:
belatedly fixing #78


jshipton:
Cool - thanks Andrew!

PR #98, opened 2017-05-17
Hydrostatic Skamarock Klemp example

colinjcotter:
Added a Hydrostatic Skamarock Klemp wave example.

This example is a vertical slice but involves Coriolis and non-zero v so it uses the 1-element wide periodic mesh. I added capability to add an arbitary source expression as a balancing force is needed for this example.


jshipton:
@wence- any idea why this is still waiting for travis tests to run?

jshipton:
@wence- this seems to be waiting for jenkins to run tests - any idea why? (Side issue: I can't look at jenkins - safari says it can't establish a secure connection to the server - not sure if this is something I can fix or if jenkins is having problems...)

wence-:
Jenkins appears to be fallen over at the moment.  I have pinged tim.

PR #99, opened 2017-05-19
A hybridised solver for the compressible equations

colinjcotter:
Not ready yet but I made the PR so we can start discussing it.

When finished this will implement a hybridisable compressible solver. Unlike the Poisson example, it has no non-hybridisable equivalent, because the dS_h jump term contains the Lagrange multiplier l0 which doesn't get substituted for anything. So we need to explicitly form the hybridisable equations, solve them in hybridised form and then reconstruct.

### DIFF
--- a/gusto/linear_solvers.py
+++ b/gusto/linear_solvers.py
@@ -3,8 +3,9 @@ from firedrake import split, LinearVariationalProblem, \
     LinearVariationalSolver, TestFunctions, TrialFunctions, \
     TestFunction, TrialFunction, lhs, rhs, DirichletBC, FacetNormal, \
     div, dx, jump, avg, dS_v, dS_h, inner, MixedFunctionSpace, dot, grad, \
-    Function, Expression, MixedVectorSpaceBasis, VectorSpaceBasis, warning
-
+    Function, Expression, MixedVectorSpaceBasis, VectorSpaceBasis, warning, \
+    FunctionSpace, BrokenElement, mesh, ds_v, ds_h, Tensor, assemble, \
+    LinearSolver, Projector
 from gusto.forcing import exner, exner_rho, exner_theta
 from abc import ABCMeta, abstractmethod
 
@@ -220,6 +221,7 @@ class CompressibleSolver(TimesteppingSolver):
         self.theta_solver.solve()
         theta.assign(self.theta)
 
+
 class HybridisedCompressibleSolver(TimesteppingSolver):
     """
     Timestepping linear solver object for the compressible equations
@@ -281,7 +283,7 @@ class HybridisedCompressibleSolver(TimesteppingSolver):
 
         l0 = TrialFunction(Vtrace)
         dl = TestFunction(Vtrace)
-        
+
         n = FacetNormal(state.mesh)
 
         # Get background fields
@@ -318,9 +320,9 @@ class HybridisedCompressibleSolver(TimesteppingSolver):
         rbareqn = (l0 - avg(rhobar))*dl*(dS_vp + ds_vp + dS_hp + ds_hp)
         rhobar_prob = LinearVariationalProblem(lhs(rbareqn),rhs(rbareqn),rhobar_tr)
         self.rhobar_solver = LinearVariationalSolver(rhobar_prob,
-                                                solver_parameters={'ksp_type':'preonly',
-                                                                'pc_type':'bjacobi',
-                                                                   'pc_sub_type':'lu'})
+                                                     solver_parameters={'ksp_type':'preonly',
+                                                                        'pc_type':'bjacobi',
+                                                                        'pc_sub_type':'lu'})
 
         Aeqn = (
             inner(w, (u - u_in))*dx
@@ -339,18 +341,18 @@ class HybridisedCompressibleSolver(TimesteppingSolver):
 
         dl = dl('+')
         l0 = l0('+')
-        
+
         K = Tensor(beta*cp*inner(thetabar*w,n)*l0*(dS_vp + dS_hp)
                    + beta*cp*inner(thetabar*w,n)*l0*(ds_vp + ds_hp))
-        L = Tensor(dl*inner(u,n)*(dS_vp + dS_hp) + dl*inner(u,n)*(ds_vp + ds_hp)
+        L = Tensor(dl*inner(u,n)*(dS_vp + dS_hp) + dl*inner(u,n)*(ds_vp + ds_hp))
 
         #  U = A^{-1}(-Kl + U_r), 0=LU=-(LA^{-1}K)l, so (LA^{-1}K)l = LU
         # reduced eqns for l0
         S = assemble(L * Aop.inv * K)
         self.Rexp = L * Aop.inv * Arhs
-        #place to put the RHS (self.Rexp gets assembled in here)
+        # place to put the RHS (self.Rexp gets assembled in here)
         self.R = Function(M)
-                   
+
         # Set up the LinearSolver for the system of Lagrange multipliers
         self.lSolver = LinearSolver(S, solver_parameters=self.params)
         # a place to keep the solution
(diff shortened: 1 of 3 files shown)


gusto/linear_solvers.py
```diff
@@ -220,6 +220,182 @@ def solve(self):
...
+                           'pc_fieldsplit_schur_fact_type': 'FULL',
+                           'pc_fieldsplit_schur_precondition': 'selfp',
+                           'fieldsplit_0_ksp_type': 'preonly',
+                           'fieldsplit_0_pc_type': 'bjacobi',
+                           'fieldsplit_0_sub_pc_type': 'ilu',
+                           'fieldsplit_1_ksp_type': 'preonly',
+                           'fieldsplit_1_pc_type': 'gamg',
+                           'fieldsplit_1_mg_levels_ksp_type': 'chebyshev',
+                           'fieldsplit_1_mg_levels_ksp_chebyshev_estimate_eigenvalues': True,
+                           'fieldsplit_1_mg_levels_ksp_chebyshev_estimate_eigenvalues_random': True,
```
dorugeber: I suspect that the Chebyshev-related lines above are wrong.  An easy way to test this is to comment them out and see if the linear solver behaves differently.  http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/KSP/KSPCHEBYSHEV.html suggests that 'esteig' might be correct, although IIRC I tried this in the MA stuff and it didn't seem to do anything either.

gusto/linear_solvers.py
```diff
@@ -220,6 +220,182 @@ def solve(self):
...
+        Vtrace = FunctionSpace(mesh, "HDiv Trace", tdegree)
+
+        # Split up the rhs vector (symbolically)
+        u_in, rho_in, theta_in = split(state.xrhs)
+
+        # Build the reduced function space for u,rho
+        M = MixedFunctionSpace((Vu_broken, Vrho, Vtrace))
+        w, phi, dl = TestFunctions(M)
+        u, rho, l0 = TrialFunctions(M)
+
```
thomasgibson: Ah so you're defining a triple `(Vu_broken, Vrho, Vtrace)`. Due to the whole "pyop2 can't write into mixed data objects," you'll want to define the trace space separately and just leave `M = MixedFunctionSpace((Vu_broken, Vrho))`. So:

```python
M = MixedFunctionSpace((Vu_broken, Vrho))
w, phi = TestFunctions(M)
u, rho = TrialFunctions(M)
l0 = TrialFunction(Vtrace)
dl = TestFunction(Vtrace)
```

This is because computing the schur complement for the multiplier system will result in an expression whose arguments lie in a non-mixed space.

gusto/linear_solvers.py
```diff
@@ -220,6 +220,182 @@ def solve(self):
...
+        dS_hp = dS_v(degree=(self.quadrature_degree))
+
+        eqn = (
+            inner(w, (u - u_in))*dx
+            - beta*cp*div(theta*V(w))*pibar*dxp
+            # following does nothing but is preserved in the comments
+            # to remind us why (because V(w) is purely vertical.
+            # + beta*cp*jump(theta*V(w),n)*avg(pibar)*dS_v
+            - beta*cp*div(thetabar*w)*pi*dxp
+            + beta*cp*jump(thetabar*w,n)*l0*(dS_vp + dS_hp)
```
thomasgibson: From my previous comment above about the mixed space, now for `eqn` you only need to write the forms corresponding to `w, u` and `phi, rho`. So just drop the `l0, dl` terms as they will be handled later.

gusto/linear_solvers.py
```diff
@@ -220,6 +220,182 @@ def solve(self):
...
+
+        # Boundary conditions (assumes extruded mesh)
+        dim = M.sub(0).ufl_element().value_shape()[0]
+        bc = ("0.0",)*dim
+        bcs = [DirichletBC(M.sub(0), Expression(bc), "bottom"),
+               DirichletBC(M.sub(0), Expression(bc), "top")]
+
+        # Solver for u, rho
+        NEED TO BUILD THIS INTO A LAMBDA SOLVER
+
```
thomasgibson: Once the trace unknowns are added to the system, doing the Schur-complement to get the multiplier system is straightforward. Something like
```python
K0 = Tensor(beta*cp*dot(thetabar*w, n)*l0*(dS_vp + dS_hp))
K1 = Tensor(dl*dot(u,n0)*(dS_vp + dS_hp))
A = Tensor(aeqn) # It's already defined on the broken space at this point

S = assemble(K1 * A.inv * K0, bcs=...)
R = assemble(K1 * A.inv * Leqn)
lambdar = Function(Vtrace)
solve(S, lambdar, R, solver_parameters={...})
```

gusto/linear_solvers.py
colinjcotter: Yes, these defaults are just copy pasted and haven't been set for this class officially yet

gusto/linear_solvers.py
```diff
@@ -221,6 +222,207 @@ def solve(self):
...
+                                                                       'pc_sub_type':'ilu'},
+                                                    options_prefix='thetabacksubstitution')
+
+    def solve(self):
+        """
+        Apply the solver with rhs state.xrhs and result state.dy.
+        """
+
+        # Assemble the RHS for lambda into self.R
+        self.R.assemble(self.Rexp)
```
wence-: You presumably mean `assemble(self.R, tensor=self.Rexp)` here.

gusto/linear_solvers.py
```diff
@@ -221,6 +222,207 @@ def solve(self):
...
+        # Solve for lambda
+        self.lSolver.solve(self.lambdar, self.R)
+        # Assemble the RHS for uhat, rho reconstruction
+        self.Rurho.assemble(self.Rurhoexp)
```
wence-: Likewise.

gusto/linear_solvers.py
```diff
@@ -221,6 +222,214 @@ def solve(self):
...
+                                                     solver_parameters={'ksp_type':'preonly',
+                                                                        'pc_type':'bjacobi',
+                                                                        'pc_sub_type':'lu'})
+
+        Aeqn = (
+            inner(w, (u - u_in))*dx
+            - beta*cp*div(theta*V(w))*pibar*dxp
+            - beta*cp*div(thetabar*w)*pi*dxp
+            + (phi*(rho - rho_in) - beta*inner(grad(phi), u)*rhobar)*dx
+            + beta*inner(phi*u, n)*rhobar_tr*(dS_v + dS_h)
```
thomasgibson: A comment: the compile error you're experiencing is due to the form:
```
beta*inner(phi*u, n)*rhobar_tr*(dS_v + dS_h)
```
However, if you remove `rhobar_tr`, the tensor assembles without error. So the problem seems to be in `rhobar_tr`.

I can also verify that Slate can indeed assemble forms with trace coefficients in both extruded and non-extruded cases. I just verified this moments ago. I'm still looking into the issues and will update.

gusto/linear_solvers.py
```diff
@@ -221,6 +222,214 @@ def solve(self):
...
+        self.lSolver = LinearSolver(S, solver_parameters=self.params)
+        # a place to keep the solution
+        self.lambdar = Function(Vtrace)
+
+        # Place to put result of u rho solver
+        self.urho = Function(M)
+
+        # Reconstruction of broken u and rho
+        self.ASolver = LinearSolver(assemble(Aop),
+                                    solver_parameters={'ksp_type':'preonly',
```
thomasgibson: Here you want `lhs(Aeqn)` to perform the global reconstruction without Slate. However, you will need to modify `Aeqn` to include the jump terms as you normally would.

Probably better to just store all the cell integrals in something, say `A_dx`, and the facet terms in `A_dS` (with Slate-restricted syntax) and `A_jump_dS` (including jumps).

Then `Aop` above becomes `Aop = Tensor(lhs(A_dx + A_dS))` and the global reconstruct here would be `assemble(lhs(A_dx + A_jump_dS), ...)`. Just a thought.

thomasgibson:
Closing this. An updated branch with a working example (still in progress) can be found here: https://github.com/firedrakeproject/gusto/tree/hybrid-compressible-solver

PR #100, opened 2017-05-23
Compressible Eady test case

hiroe-:
This pull request adds

- a compressible Eady test case to examples. CompressibleEadyParameters and CompressibleEadyForcing are added for the test case.

- diagnostic fields HorizontalVelocity, MeridionalVelocity, KineticEnergy, CompressibleKineticEnergy, ExnerPi, ExnerPi_perturbation. Eady-related diagnostics are added in eady_diagnostics. This also adds min, max and rms of diagnostic fields to the diagnostics dictionary.  

- initialisation tools for incompressible and compressible Eady tests

- diagnostic_everydump option in the Timestepper to call state.diagnostic_dump() everytime dump() is called. 

This pull request also fixes

- time step restriction in the incompressible Eady test case.

- a parallel issue in 2d examples (see issue #90).

- maximum time step counter and end time.


jshipton:
By the way, looks like you need to merge from master to get the tests running.

gusto/configuration.py
```diff
@@ -93,20 +95,19 @@ class EadyParameters(Configuration):
...
 
 
-class CompressibleEadyParameters(Configuration):
+class CompressibleEadyParameters(CompressibleParameters):
 
     """
     Physical parameters for Compressible Eady
     """
+    eady_parameters = EadyParameters()
+    Nsq = eady_parameters.Nsq
```
jshipton: I think that rather than doing this, you can add EadyParameters in the class definition:
`class CompressibleEadyParameters(CompressibleParameters, EadyParameters)`

gusto/eady_diagnostics.py
```diff
@@ -1,50 +1,31 @@
...
         u = state.fields("u")
         rho = state.fields("rho")
-        kineticv = 0.5*rho*u[1]*u[1]
-        return self.field(state.mesh).interpolate(kineticv)
+        energy = self.kinetic(u[1], rho)
+        return self.field(state.mesh).interpolate(energy)
 
 
-class EadyPotentialEnergy(DiagnosticField):
+class EadyPotentialEnergy(Energy):
```
jshipton: Is this unique to the Eady problem? Isn't it just the incompressible Boussinesq potential energy?

gusto/eady_diagnostics.py
```diff
@@ -54,15 +35,9 @@ def compute(self, state):
...
 
-    def field(self, mesh):
-        if hasattr(self, "_field"):
-            return self._field
-        self._field = Function(FunctionSpace(mesh, "DG", 0), name=self.name)
-        return self._field
-
     def compute(self, state):
         x, y, z = SpatialCoordinate(state.mesh)
         g = state.parameters.g
```
jshipton: Ditto - isn't this just compressible potential energy?

gusto/eady_diagnostics.py
```diff
@@ -78,31 +53,19 @@ def compute(self, state):
         return self.field(state.mesh).interpolate(potential)
 
 
-class EadyTotalEnergy(DiagnosticField):
+class EadyTotalEnergy(Energy):
     name = "EadyTotalEnergy"
```
jshipton: If EadyPotentialEnergy is not Eady specific, then neither is this.

gusto/eady_diagnostics.py
```diff
@@ -78,31 +53,19 @@ def compute(self, state):
...
     def compute(self, state):
         kinetic = KineticEnergy()
         potential = EadyPotentialEnergy()
         total = kinetic.compute(state) + potential.compute(state)
         return self.field(state.mesh).interpolate(total)
 
 
-class CompressibleEadyTotalEnergy(DiagnosticField):
+class CompressibleEadyTotalEnergy(Energy):
     name = "CompressibleEadyTotalEnergy"
```
jshipton: It feels like there should be a nice way to compute the total energy, given the kinetic and potential energies - we shouldn't need a separate class... let me think...

jshipton:
OK, so the problem (I'm referring to my comment on the TotalEnergy class) is that we can't access other diagnostic fields in these diagnostic fields classes - so we can just use Sum(kinetic_energy+potential_energy). One way of dealing with this would be to change the Sum class so that it can take in other diagnostic field classes, rather than just fieldnames. Then amend the compute method so that it either extracts the fields from state or computes them. The only issue here is that if you want to dump both kinetic energy and total energy, then you will be computing the kinetic energy twice.

gusto/eady_diagnostics.py
hiroe-: This is unique to the Compressible Eady problem which needs the extra term - cp*Pi0*theta.

jshipton: What's Pi0?

hiroe-: Pi0 is in the extra term we added to the v equation to stop the mean flow in the compressible Eady test.

hiroe-: which is constant.

gusto/eady_diagnostics.py
hiroe-: I think non-Eady problem doesn't need -H/2.

jshipton: Ah yes, I think you're right

gusto/diagnostics.py
```diff
@@ -92,8 +140,72 @@ def field(self, mesh):
 
     def compute(self, state):
         u = state.fields("u")
-        uh = u[0]
-        return self.field(state.mesh).interpolate(uh)
+        v = u[1]
+        return self.field(state.mesh).interpolate(v)
```
wence-: So this always returns the second component y in 3D? Again, is this right?

gusto/diagnostics.py
```diff
@@ -92,8 +140,72 @@ def field(self, mesh):
...
+
+class Energy(DiagnosticField):
+
+    def field(self, mesh):
+        if hasattr(self, "_field"):
+            return self._field
+        self._field = Function(FunctionSpace(mesh, "DG", 0), name=self.name)
+        return self._field
+
+    def kinetic(self, u, rho=None):
```
wence-: Docstring?

gusto/eady_diagnostics.py
```diff
@@ -0,0 +1,73 @@
...
+    KineticEnergy, CompressibleKineticEnergy
+from gusto.forcing import exner
+
+
+class KineticEnergyV(Energy):
+    name = "KineticEnergyV"
+
+    def compute(self, state):
+        u = state.fields("u")
+        energy = self.kinetic(u[1])
```
wence-: So this is the kinetic energy of the y component?

gusto/transport_equation.py
```diff
@@ -58,8 +58,8 @@ def __init__(self, state, V, ibp="once"):
 
         # default solver options
         self.solver_parameters = {'ksp_type':'preonly',
-                                  'pc_type':'bjacobi',
-                                  'sub_pc_type': 'ilu'}
+                                  'pc_type':'lu',
+                                  'pc_factor_mat_solver_package': 'mumps'}
```
wence-: Do you need this for accuracy? Because the operator is no longer cell local? Would it be better then to use an iterative method?

gusto/state.py
```diff
@@ -253,9 +254,12 @@ def dump(self, t=0, pickup=False):
...
 
             # compute diagnostics
+            diagnostic_fns = ['min', 'max', 'rms', 'l2']
             for name in self.diagnostics.fields:
-                data = self.diagnostics.l2(self.field_dict[name])
-                self.diagnostic_data[name]["l2"].append(data)
+                for fn in diagnostic_fns:
+                    d = getattr(self.diagnostics, fn)
+                    data = d(self.field_dict[name])
+                    self.diagnostic_data[name][fn].append(data)
```
wence-: Should there be some way of controlling what diagnostics are computed?

gusto/configuration.py
```diff
@@ -85,9 +87,21 @@ class ShallowWaterParameters(Configuration):
...
-    geopotential = False  # use geopotential for gravity term
+    f = None
+
+
+class CompressibleEadyParameters(CompressibleParameters, EadyParameters):
+
+    """
+    Physical parameters for Compressible Eady
+    """
+    g = 10.
```
hiroe-: This is to compare the results with the previous Eady papers that use this value, although this would not affect the results much.

gusto/diagnostics.py
```diff
@@ -77,12 +121,16 @@ def field(self, mesh):
 
     def compute(self, state):
         u = state.fields("u")
-        w = u[1]
+        if(state.mesh.geometric_dimension() == 2):
+            w = u[1]
+        elif(state.mesh.geometric_dimension() == 3):
```
hiroe-: Fixed

gusto/diagnostics.py
hiroe-: This is what I meant. I would like to track v in Eady tests which uses a 3D mesh.

gusto/diagnostics.py
hiroe-: Added

gusto/eady_diagnostics.py
hiroe-: Exactly. Tracking the y component is useful to analyse Eady results.

gusto/transport_equation.py
hiroe-: This fixed the parallel issue #90 for the 2d_example sk_nonlinear.py. However I am not sure if this is good for accuracy etc.

gusto/diagnostics.py
```diff
@@ -17,6 +18,34 @@ def register(self, *fields):
                 self.fields.append(f)
 
     @staticmethod
+    def min(f):
+        fmin = op2.Global(1, [1000], dtype=float)
```
hiroe-: Thanks, this fixed a bug

gusto/diagnostics.py
```diff
@@ -66,6 +95,21 @@ def compute(self, state):
...
+    def field(self, mesh):
+        if hasattr(self, "_field"):
+            return self._field
+        self._field = Function(FunctionSpace(mesh, "CG", 1), name=self.name)
+        return self._field
+
+    def compute(self, state):
+        u = state.fields("u")
+        uh = u[0]
+        return self.field(state.mesh).interpolate(uh)
```
hiroe-: Renamed it VelocityX to avoid confusion. Similarly, Meridional and Vertical velocities are renamed as VelocityY and VelocityZ. KineticEnergyV is renamed as KineticEnergyY.

gusto/state.py
hiroe-: I think we could make diagnostic_fields a dictionary and specify diagnostic_fns for each diagnostics, e.g., diagnostic_fields = {CourantNumber():['max'], VelocityY():['max', 'rms']}. What do you think @jshipton?

wence-:
OK, so I think the comments I made on #104 were relevant here too.  Since that pulls in all of these changes, should we just review that one and close this?

jshipton:
@wence- close this, or merge this?

hiroe-:
Close this pull request as all the changes are pulled into #104.

PR #101, opened 2017-05-26
Boussinesq 2d lab

P-Burns:


gusto/timeloop.py
```diff
@@ -174,6 +191,15 @@ def run(self, t, tmax, x_end=None):
                 # advects field
                 advection.apply(field, field)
 
+            for physics in self.physics_list:
+                physics.apply()
+         
+	    #Added by PB, 15/05/2017
```
wence-: Don't do this.  `git blame` is much more correct in remembering what is added.

gusto/timeloop.py
```diff
@@ -174,6 +191,15 @@ def run(self, t, tmax, x_end=None):
...
+	    with timed_stage("Diffusion"):
+              for name, diffusion in self.diffusion_dict.iteritems():
+                 field = getattr(state.fields, name)
+                 diffusion.apply(field, field)
```
wence-: So now the advection time stepping scheme can include diffusion as well.  That seems... odd?

I wonder if we should think about how to redesign the way the time stepping is glued together.  Right now, if you want to do "advection + some stuff" you need to copy the full time loop, excise the bits you don't want, which is not a scalable method of writing code.

What if you want to compute diagnostics of various fields

gusto/timeloop.py
```diff
@@ -55,14 +56,18 @@ class Timestepper(BaseTimestepper):
     :arg forcing: a :class:`.Forcing` object
     """
 
-    def __init__(self, state, advection_dict, linear_solver, forcing, diffusion_dict=None):
+    def __init__(self, state, advection_dict, linear_solver, forcing, diffusion_dict=None, physics_list=None):
 
         super(Timestepper, self).__init__(state, advection_dict)
```
wence-: Given that you changed the superclass constructor, you should now pass `diffusion_dict` into the constructor here too.

gusto/timeloop.py
```diff
@@ -55,14 +56,18 @@ class Timestepper(BaseTimestepper):
...
         self.linear_solver = linear_solver
         self.forcing = forcing
         self.diffusion_dict = {}
         if diffusion_dict is not None:
             self.diffusion_dict.update(diffusion_dict)
```
wence-: And then this code should move there.

gusto/timeloop.py
```diff
@@ -152,6 +161,14 @@ def run(self, t, tmax, pickup=False):
 
 class AdvectionTimestepper(BaseTimestepper):
 
+    def __init__(self, state, advection_dict, diffusion_dict, physics_list=None):
+
+        super(AdvectionTimestepper, self).__init__(state, advection_dict, diffusion_dict)
+        if physics_list is not None:
+            self.physics_list = physics_list
+        else:
+            self.physics_list = []
```
wence-: This code is replicated in both the `TimeStepper` and `AdvectionTimeStepper`. If all time steppers need it, it should be initialised in `BaseTimeStepper`, like the diffusion_dict and advection_dict arguments.

wence-: When you do that, then you can remove the custom constructor here completely.

gusto/timeloop.py
```diff
@@ -152,6 +161,14 @@ def run(self, t, tmax, pickup=False):
...
+
     def run(self, t, tmax, x_end=None):
         state = self.state
```
wence-: I note this isn't your code, but below the `print ...` line should be replaced by proper logging (just to note for later).

examples/diffusion_2d.py
```diff
@@ -0,0 +1,195 @@
...
+diffusion_dict = {"u": InteriorPenalty(state, Vu, kappa=Constant(1.*10**(-6)),
+                                           mu=Constant(10./delta), bcs=bcs_u),
+                      "b": InteriorPenalty(state, Vb, kappa=Constant(1.4*10**(-7)),
+                                               mu=Constant(10./delta), bcs=bcs_b)}
+
+
+##############################################################################
+# build time stepper
+##############################################################################
+physics_list = {}
```
wence-: This is a dict, not a list, so it is probably wrong.

examples/diffusion_2d.py
```diff
@@ -0,0 +1,195 @@
...
+# Kinematic viscosity = 1.*10**(-6)
+# Heat diffusivity = 1.4*10**(-7)
+Vu = u0.function_space()
+Vb = b0.function_space()
+delta = L/columns 		#Grid resolution (same in both directions).
+
+
+bcs_u = [DirichletBC(Vu, 0.0, "bottom"), DirichletBC(Vu, 0.0, "top")]
+#bcs_b = [DirichletBC(Vb, 0.0, "bottom"), DirichletBC(Vb, 0.0, "top")]
+bcs_b = {}
```
wence-: bcs should be a list, not a dict.

examples/diffusion_2d.py
```diff
@@ -0,0 +1,195 @@
...
+
+diffusion_dict = {"u": InteriorPenalty(state, Vu, kappa=Constant(1.*10**(-6)),
+                                           mu=Constant(10./delta), bcs=bcs_u),
+                      "b": InteriorPenalty(state, Vb, kappa=Constant(1.4*10**(-7)),
+                                               mu=Constant(10./delta), bcs=bcs_b)}
```
wence-: indentation is screwy.

examples/diffusion_2d.py
```diff
@@ -0,0 +1,195 @@
...
+                               equation_form="advective")
+
+advection_dict = {}
+#advection_dict["u"] = ThetaMethod(state, u0, ueqn)
+#advection_dict["b"] = SSPRK3(state, b0, beqn)
+
+#For linear equations use:
+#advection_dict["u"] = NoAdvection(state, u0, None) 
+#beqn = LinearAdvection(state, Vb, qbar=b_b)
+#advection_dict["b"] = SSPRK3(state, b0, beqn)
```
wence-: Don't leave commented code like this in examples.

examples/diffusion_2d.py
```diff
@@ -0,0 +1,195 @@
...
+# advection_dict is a dictionary containing field_name: advection class
+ueqn = EulerPoincare(state, u0.function_space())
+supg = True
+if supg:
+    beqn = SUPGAdvection(state, Vb,
+                         supg_params={"dg_direction":"horizontal"},
+                         equation_form="advective")
+else:
+    beqn = EmbeddedDGAdvection(state, Vb,
+                               equation_form="advective")
```
wence-: None of this stuff is ever used.

examples/diffusion_2d.py
```diff
@@ -0,0 +1,195 @@
...
+##############################################################################
+# build time stepper
+##############################################################################
+physics_list = {}
+stepper = AdvectionTimestepper(state, advection_dict, diffusion_dict, physics_list)
+
+##############################################################################
+# Run!
+##############################################################################
+stepper.run(t=0, tmax=tmax)
```
wence-: This whole file looks rather cargo-culted.  Can you cut out all the stuff that isn't actually needed?

It is very confusing that you're doing `AdvectionTimestepper` with no advection, and only diffusion, for example :).

examples/boussinesq_2d_lab.py
```diff
@@ -0,0 +1,205 @@
...
+                            'fieldsplit_1_ksp_type':'preonly'}
+else:
+    tmax = 3600*6.
+    # use default linear solver parameters (i.e. mumps)
+    linear_solver_params = None
+
+##############################################################################
+# set up mesh
+##############################################################################
+# Construct 1d periodic base mesh for idealised lab experiment of Park et al. (1994)
```
wence-: It's a good idea to spell these references out.

examples/diffusion_2d.py
```diff
@@ -0,0 +1,195 @@
...
+                            'fieldsplit_1_ksp_type':'preonly'}
+else:
+    tmax = 3600*6.
+    # use default linear solver parameters (i.e. mumps)
+    linear_solver_params = None
+
+##############################################################################
+# set up mesh
+##############################################################################
+# Construct 1d periodic base mesh for idealised lab experiment of Park et al. (1994)
```
wence-: This example solves the heat equation, right?  Was that this paper?

PR #102, opened 2017-05-26
Fix is_cg determination

wence-:
EnrichedElement now has a defaultdict entity_dofs, so we never get a
KeyError when checking continuity.

### DIFF
--- a/gusto/transport_equation.py
+++ b/gusto/transport_equation.py
@@ -39,10 +39,14 @@ class TransportEquation(object):
         entity_dofs = V.finat_element.entity_dofs()
         # If there are as many dofs on vertices as there are vertices,
         # assume a continuous space.
-        try:
+        if 0 in entity_dofs:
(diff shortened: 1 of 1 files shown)


hiroe-:
Works here too, thanks!!

wence-:
Insufficient, a similar fix needs to be applied in Firedrake.

miklos1:
Maybe we could just fix FIAT instead?

wence-:
That would perhaps be safer.  defaultdicts are useful for internal usage, but when exposed through an API they can be dangerous it would seem.

miklos1:
Yes, just cast them to a `dict` once they are populated...

miklos1:
Does firedrakeproject/fiat@07a36ba fix this issue?

wence-:
Superseded by FIAT changes.

miklos1:
FIAT changes pushed.

PR #104, opened 2017-06-01
Incompressible Eady tools

hiroe-:
This pull request adds the following tools for the Incompressible Eady test case:

- GeostrophicImbalance, TrueResidualV, and SawyerEliassenU in eady_diagnostics.
- We can rescale the problem by changing beta in the test file.

The rest of the changes are the same as #100.


gusto/state.py
```diff
@@ -220,7 +225,7 @@ def setup_dump(self, pickup=False):
             fields_ll[name] = Function(functionspaceimpl.WithGeometry(f.function_space(), mesh_ll), val=f.topological, name=name+'_ll')
             self.to_dump_latlon.append(fields_ll[name])
 
-    def dump(self, t=0, pickup=False):
+    def dump(self, t=0, diagnostic_everydump=False, pickup=False):
```
wence-: docstring?

gusto/state.py
```diff
@@ -253,9 +258,12 @@ def dump(self, t=0, pickup=False):
                 self.dumpfile_ll.write(*self.to_dump_latlon)
 
             # compute diagnostics
+            diagnostic_fns = ['min', 'max', 'rms', 'l2']
```
wence-: This feels like the wrong place to enumerate these names.  Basically you're saying what diagnostic quantities the `Diagnostics` class can compute.  I think it would be better to add a field to the `Diagnostics` class that says what these are.

```python
class Diagnostics

    available_diagnostics = ["min", "max", "rms", "l2"]
    ...
```

And then here we do:
```python
for fn in self.diagnostics.available_diagnostics:
   ...
```

That way as well, you could later extend `State` to say which diagnostics you want to compute.

gusto/diagnostics.py
```diff
@@ -39,6 +69,10 @@ def compute(self, state):
         """ Compute the diagnostic field from the current state"""
         pass
 
+    def _setup_solver(self, state):
+        """Setup solvers for calculating this diagnostic field"""
+        pass
```
jshipton: Not ideal, but for now why can't the compute method just call the setup_solver method and have it set up the solver on the first call, or return the solver otherwise - just like the current field method does? Otherwise it is a bit confusing to have a method in the base DiagnosticField class that most diagnostic fields don't need.

gusto/state.py
jshipton: Yes, I like this. At the moment we compute all the diagnostics on all of the fields - which I don't mind because it doesn't seem to massively increase runtime and these things would be a pain to compute after running so I'd rather have them than not. However, that could change if we add a lot of diagnostics, some of which might be field specific.

gusto/diagnostics.py
jshipton: I don't understand the bit about passing the class constructor in to state - what would that look like? Also, your compute method takes ... args - do you mean to imply that different compute methods could take different arguments? How does that work?

hiroe-: _setup_solver is removed.

gusto/state.py
hiroe-: added

hiroe-:
Sorry for a long silence. All the changes apart from suggestions for rearranging diagnostics have been made. Not sure why tests do not begin... any ideas?

hiroe-:
Tests are successful. Are we working on rearranging diagnostics in this pull request or can I merge it? I think it would make sense to make a new branch for sorting diagnostics out as it would involve changes to all the diagnostic fields and examples which need diagnostics (not only the eady-related ones).

@wence- Merging this pull request is blocked as not all the requested changes are addressed. Can't I merge it until I make changes to diagnostics or how does it work?

jshipton:
I think we have to leave diagnostics restructuring for another branch. I'm not sure what to do with the setup_solver stuff though - the suggestion I made above wasn't explained very well. I meant to make the setup_solver method like the field method:
```
def setup_solver(self, state):
   if hasattr(self, "_solver"):
      return _solver
   else:
      do stuff to set up the solver and return it
```
Then in the compute method you can just use the solver and it will be created the first time you try to use it - just like the field. I'm not sure if this is better that what you've done though - might be nicer than the `try... except` stuff? @wence- is away this week though at Fenics so not sure what to do about the blocking.

wence-:
I agee we can't do everything here. Go ahead.

PR #105, opened 2017-06-05
Condensation bugs

tommbendall:
This pull request fixes two (embarrassing) bugs in the condensation code.

Firstly a missing pair of brackets, and secondly a change to the way the temperature and water mixing ratio fields are updated. Previously the temperature and water fields got updated by different condensation rates (it should of course be the same). This has been corrected.

### DIFF
--- a/gusto/physics.py
+++ b/gusto/physics.py
@@ -87,14 +87,15 @@ class Condensation(Physics):
                                    (cp * R_v * T ** 2.0)))))
 
         # adjust cond rate so negative concentrations don't occur
-        self.cond_rate = Interpolator(conditional(dot_r_cond < 0,
-                                conditional(-dot_r_cond > self.water_c / dt,
-                                            -self.water_c / dt,
-                                            dot_r_cond),
-                                conditional(dot_r_cond > self.water_v / dt,
-                                            self.water_v / dt,
-                                            dot_r_cond)), Vt)
-
+        self.lim_cond_rate = Interpolator(conditional(dot_r_cond < 0,
+                                                      conditional(-dot_r_cond > self.water_c / dt,
+                                                                  -self.water_c / dt,
+                                                                  dot_r_cond),
+                                                      conditional(dot_r_cond > self.water_v / dt,
+                                                                  self.water_v / dt,
+                                                                  dot_r_cond)), Vt)
+        
+        self.cond_rate = Function(Vt)
         self.water_v_new = Interpolator(self.water_v - dt * self.cond_rate, Vt)
         self.water_c_new = Interpolator(self.water_c + dt * self.cond_rate, Vt)
         self.theta_new = Interpolator(self.theta *
@@ -103,7 +104,7 @@ class Condensation(Physics):
                                         R_v * cv * c_pml / (R_m * cp * c_vml))), Vt)
 
     def apply(self):
-        self.cond_rate.interpolate()
+        self.cond_rate.assign(self.lim_cond_rate.interpolate())
         self.water_v.assign(self.water_v_new.interpolate())
         self.water_c.assign(self.water_c_new.interpolate())
         self.theta.assign(self.theta_new.interpolate())
--- a/gusto/physics.py
+++ b/gusto/physics.py
@@ -1,5 +1,5 @@
 from abc import ABCMeta, abstractmethod
-from firedrake import exp, Interpolator, conditional, interpolate
+from firedrake import exp, Interpolator, conditional, interpolate, Function
 
 
 class Physics(object):
@@ -42,7 +42,7 @@ class Condensation(Physics):
         rho = state.fields('rho')
 
         # declare function space
-        V = self.theta.function_space()
+        Vt = self.theta.function_space()
 
         param = self.state.parameters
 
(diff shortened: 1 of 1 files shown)


gusto/physics.py
```diff
@@ -78,31 +78,30 @@ def __init__(self, state):
...
-                                conditional(cond_rate > self.water_v / dt,
-                                            self.water_v / dt,
-                                            cond_rate)), V)
-
-        self.water_v_new = Interpolator(self.water_v - dt * cond_rate, V)
-        self.water_c_new = Interpolator(self.water_c + dt * cond_rate, V)
+        self.lim_cond_rate = Interpolator(conditional(dot_r_cond < 0,
+                                                      conditional(-dot_r_cond > self.water_c / dt,
+                                                                  -self.water_c / dt,
+                                                                  dot_r_cond),
```
miklos1: Is this just a really confusing way of writing `min_value` or `max_value`?

gusto/physics.py
```diff
@@ -78,31 +78,30 @@ def __init__(self, state):
...
+                                                      conditional(dot_r_cond > self.water_v / dt,
+                                                                  self.water_v / dt,
+                                                                  dot_r_cond)), Vt)
```
miklos1: Same here.

tommbendall: Yep, I think you're exactly right about both of these. I don't understand why I didn't do it that way originally...

gusto/physics.py
```diff
@@ -78,31 +79,26 @@ def __init__(self, state):
...
+                                      (1.0 + dt * self.cond_rate *
                                        (cv * L_v / (c_vml * cp * T) -
-                                        R_v * cv * c_pml / (R_m * cp * c_vml))), V)
+                                        R_v * cv * c_pml / (R_m * cp * c_vml))), Vt)
 
     def apply(self):
+        self.cond_rate.assign(self.lim_cond_rate.interpolate())
         self.water_v.assign(self.water_v_new.interpolate())
         self.water_c.assign(self.water_c_new.interpolate())
         self.theta.assign(self.theta_new.interpolate())
```
wence-: There seems to be no need to have these intermediate variables for `cond_rate` and `theta`.  Does it work to just do:

```python
cond_interp = Interpolator(..., self.cond_rate)
theta_interp = Interpolator(..., self.theta)

...
self.cond_interp.interpolate()
self.theta_interp.interpolate()
```

?

wence-: Oh, I see `theta` appears in the RHS of the theta expression, so probably not for that one, but cond_rate is purely diagnosed, so I don't think you need the intermediate variable?

tommbendall: Oh I didn't realise that `Interpolator` could take the function to interpolate to as an argument. In that case wouldn't we still need to have a `self.cond_rate` to interpolate into? I think I'm not sure which intermediate variable you're suggesting getting rid of!

gusto/physics.py
```diff
@@ -78,31 +79,26 @@ def __init__(self, state):
...
-                                            self.water_v / dt,
-                                            cond_rate)), V)
-
-        self.water_v_new = Interpolator(self.water_v - dt * cond_rate, V)
-        self.water_c_new = Interpolator(self.water_c + dt * cond_rate, V)
+        self.lim_cond_rate = Interpolator(conditional(dot_r_cond < 0,
+                                                      max_value(dot_r_cond, - self.water_c / dt),
+                                                      min_value(dot_r_cond, self.water_v / dt)), Vt)
+
+        self.cond_rate = Function(Vt)
```
wence-: So I think you can write here:

```python
self.cond_rate = Function(Vt)
self.lim_cond_rate = Interpolator(..., self.cond_rate)
```

gusto/physics.py
```diff
@@ -78,31 +79,26 @@ def __init__(self, state):
...
+        self.water_c_new = Interpolator(self.water_c + dt * self.cond_rate, Vt)
         self.theta_new = Interpolator(self.theta *
-                                      (1.0 + dt * cond_rate *
+                                      (1.0 + dt * self.cond_rate *
                                        (cv * L_v / (c_vml * cp * T) -
-                                        R_v * cv * c_pml / (R_m * cp * c_vml))), V)
+                                        R_v * cv * c_pml / (R_m * cp * c_vml))), Vt)
 
     def apply(self):
+        self.cond_rate.assign(self.lim_cond_rate.interpolate())
```
wence-: Then here you just do `self.lim_cond_rate.interpolate()`.

gusto/physics.py
```diff
@@ -78,31 +79,26 @@ def __init__(self, state):
...
-                                            cond_rate),
-                                conditional(cond_rate > self.water_v / dt,
-                                            self.water_v / dt,
-                                            cond_rate)), V)
-
-        self.water_v_new = Interpolator(self.water_v - dt * cond_rate, V)
-        self.water_c_new = Interpolator(self.water_c + dt * cond_rate, V)
+        self.lim_cond_rate = Interpolator(conditional(dot_r_cond < 0,
+                                                      max_value(dot_r_cond, - self.water_c / dt),
+                                                      min_value(dot_r_cond, self.water_v / dt)), Vt)
```
wence-: So this says:
```python
if dot_r_cond < 0:
   dot_r_cond = max(dot_r_cond, -self.water_c/dt)
else:
   dot_r_cond = min(dot_r_cond, self.water_v/dt)
```

OK, and then the condensation levels are in water_v and water_c and they're updating like:

```
water_c = water_c + dt*cond_rate
water_v = water_v - dt*cond_rate
```

So here you're arranging that if cond_rate is too negative, then water_c goes to zero.  And equally, if cond_rate is too positive, that water_v goes to zero.

What I wonder is if you need all these intermediate fields?  Can you just not put the expressions in?  Or do you somehow need to arrange that the updates to water_c and water_v and theta are evaluated with the same value of cond_rate.

tommbendall: Somehow I seem to have missed this comment! Yes, we need `cond_rate` to be the same for updating `water_v`, `water_c` and `theta` in a given time step

dorugeber:
What is the status of this?

jshipton:
Both Colin and Tom are on holiday...

tommbendall:
Ah sorry! I'd actually been hoping for someone else to merge this as I don't seem to have merge privileges any more!

I felt a bit awkward about asking someone else to merge it though so I probably shouldn't have expected anything to happen!

Issue #91, opened 2017-04-20
don't use preonly for solver if CG in transport equation


colinjcotter:
Transport equation sets preonly as default solver for advection but this is only suitable for DG.

thomasgibson:
Can you confirm whether or not conjugate gradient + bjacobi/ILU is sufficient? This is being addressed in #111

wence-:
It's fine, it's just a mass matrix.  If you care about reductions (I don't think you do), you could use cheby + ilu, but then you need eigenvalue bounds (can be explicitly calculated), but I don't think it's worth it.

thomasgibson:
Fixed in #111

PR #111, opened 2017-06-29
Change default solver parameters for advection schemes and add optional argument

thomasgibson:
The current default parameters for `TransportEquation` uses LU+MUMPS, and causes everything to choke for higher resolution problems. I believe conjugate gradient preconditioned with block-jacobi/ILU is a better compromise. Larger problems no longer freeze up and everything continues as normal.

If the default parameters are not sufficient for a particular problem, you can now pass an optional set of solver parameters to use instead (via `solver_params`). Just like the linear solvers.

### DIFF
--- a/gusto/transport_equation.py
+++ b/gusto/transport_equation.py
@@ -21,10 +21,12 @@ class TransportEquation(object):
     :arg V: :class:`.FunctionSpace object. The function space that q lives in.
     :arg ibp: string, stands for 'integrate by parts' and can take the value
               None, "once" or "twice". Defaults to "once".
+    :arg solver_params: (optional) dictionary of solver parameters to pass to the
+                        linear solver.
     """
     __metaclass__ = ABCMeta
 
-    def __init__(self, state, V, ibp="once"):
+    def __init__(self, state, V, ibp="once", solver_params=None):
         self.state = state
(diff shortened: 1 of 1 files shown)


dorugeber:
Any reason that you're not reverting back to the options pre-44e20e7 ?  Is the difference between CG+ILU and preonly+ILU negligible for DG problems?

thomasgibson:
If my understanding of block-jacobi is correct (@wence- can correct me if I'm wrong), then block-jacobi on a DG matrix is exact since it's just inverting the local diagonal blocks. So the previous options (`'ksp_type': 'preonly' ...`) works just fine.

However, do non-DG matrices come into play? If so, then preonly block-jacobi + ILU would be terribly inaccurate, no? That's why I added `cg` as the ksp type. I don't notice any significant slow downs. If the mass matrix is DG, then the conjugate gradient method should converge in 1 iteration.

dorugeber:
See #91 (remembering that CG = continuous Galerkin ;) )

If CG + ... is good enough for DG, leave it in place and that issue can probably be closed (let Colin and Hiroe confirm)

wence-:
CG+ILU for DG will work fine, it will do two more reductions than strictly necessary, but meh.

PR #112, opened 2017-06-30
Use preonly hybridization in the ShallowWaterSolver

thomasgibson:
This PR adds hybridization parameters for the shallow water solver.

### DIFF
--- a/gusto/linear_solvers.py
+++ b/gusto/linear_solvers.py
@@ -358,8 +358,33 @@ class IncompressibleSolver(TimesteppingSolver):
 
 class ShallowWaterSolver(TimesteppingSolver):
 
-    def _setup_solver(self):
+    def __init__(self, state, params=None):
 
+        if params is None:
+            # Use hybridization parameters
+            params = {'ksp_type': 'preonly',
+                      'mat_type': 'matfree',
+                      'pc_type': 'python',
+                      'pc_python_type': 'firedrake.HybridizationPC',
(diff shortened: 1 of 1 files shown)


dorugeber:
I'm sort of wary of these parameters being copy-pasted into 4 examples.  Some better-sounding alternatives, other ideas welcome:
1) copy one example (e.g. Williamson 5) and make a new `sw_williamson5_hybridized.py` example
2) rename one of the current examples to `..._hybridized.py` and use this
3) presumably the current SW code uses some solver options (it's surely not the default GMRES/ILU, probably some Lawrence fieldsplit magic).  If these options are clearly better, why not replace the default options within Gusto?  (can always rip out the current options and use them in one example as a 'backup')

thomasgibson:
I think 3 would be the cleanest way, as I don't want to bloat the examples code with even more duplication (unless people would rather have separate hybridization tests). Making the hybridization parameters default for the `ShallowWaterSolver` is an option.

dorugeber:
Then you should get this past @jshipton and @colinjcotter!

dorugeber:
What is the status of this?

jshipton:
Final decision on solver parameters to be made by Colin (currently in Tuscany)

jshipton:
I think this is now superseded by https://github.com/firedrakeproject/gusto/pull/131 - @thomasgibson, is that correct?

thomasgibson:
Yes, we can close this.

PR #113, opened 2017-07-03
use GMRES rather than CG in asymmetric advection cases

dorugeber:
Turns out these options get used for nonsymmetric systems (e.g. via `ThetaMethod`), and CG is a Bad Idea for those.  At this point, I'm wondering if these options should be here at all... GMRES/BJacobi/ILU is just the Firedrake -- rather, PETSc -- default, no?


thomasgibson:
Is there a reason why we shouldn't just overwrite `solver_params` in the `ThetaMethod` constructor? Just curious.

colinjcotter:
Yes, we should just revert it when SUPG is used (or semi-implicit timestepping).

dorugeber:
Okay, I've switched over the solver parameters in those two cases (I think...) rather than changing the global default.  Review plz.

PR #114, opened 2017-07-03
fix w6 example to work

dorugeber:
Get the Williamson 6 example working.  @jshipton, you forgot to convert the velocity field out of spherical components!  It was just `as_vector((zonal, meridional, 0))`.

Relies on #113.


dorugeber:
Silly me.  Thanks, should be working now.

PR #115, opened 2017-07-04
no longer ignore E231, and some v minor fixes

dorugeber:


dorugeber:
By the power vested in me...

PR #116, opened 2017-07-11
added option to output data at points

jshipton:
This PR adds the option to output fields at points. This is done by passing `point_data` to the OutputParameters class. `point_data` is a dictionary of the form `{field_name: points}` where `points` is a list of coordinate points acceptable to the Firedrake `at` function. The data is output to a point_data.json file in the results/dirname directory, in the form `{field_name:{time: list_of_data_at_points}}`. There is a corresponding option `output.pointwise_everydump` to dump this dictionary at every timestep, rather than waiting until the end of the simulation. I have also moved the `diagnostic_everydump` option to the OutputParameters class.


jshipton:
Note: this doesn't work on the sphere is `at` doesn't work on immersed manifolds - maybe I should check for that...

jshipton:
I'm not sure I understand the output of `at` - I needed to use `tolist` because json couldn't cope with arrays...

wence-:
You get one value per input point as a list of zero-dimensional `numpy.array` objects.

jshipton:
So is using `tolist` the right thing to do?

wence-:
I think so.

jshipton:
Is there a nicer way of dumping the data as we go along? It'd be good if we could just append to the file rather than writing the whole thing but IIRC that isn't something json can deal with...

colinjcotter:
If json files can't be appended to (brief Google search suggests they can't) then we should just rewrite this file every dump I think.

wence-:
Merged master into this branch so that testing might run.

Jenkinsfile
```diff
@@ -51,8 +52,6 @@ make lint
                 timestamps {
                     sh '''
 . build/firedrake/bin/activate
-export PYOP2_CACHE_DIR=${VIRTUAL_ENV}/pyop2_cache
-export FIREDRAKE_TSFC_KERNEL_CACHE_DIR=${VIRTUAL_ENV}/tsfc_cache
```
wence-: By default firedrake now puts the caches in the virtualenv anyway.

gusto/state.py
```diff
@@ -274,17 +338,46 @@ def dump(self, t=0, diagnostic_everydump=False, pickup=False):
...
+        if self.output.point_data is not None:
+            data = Dataset(self.pointdata_filename, "a")
+            time = data.dimensions["time"]
+            idx = len(time)
+            times = data.variables["time"]
+            times[idx:idx+1] = self.t
+            for fname in self.output.point_data.keys():
+                grp = data.groups[fname]
+                field = grp.variables[fname]
+                print np.array(point_data[fname]).shape
```
wence-: This debugging should go (also it's now a syntax error).

gusto/configuration.py
```diff
@@ -48,6 +48,8 @@ class OutputParameters(Configuration):
     steady_state_error_fields = []
     #: List of fields for computing perturbations
     perturbation_fields = []
+    #: Dictionary of fields and points at which to dump them
+    point_data = {}
```
wence-: You should describe the format of these points.  It's not the format the `Function.at` wants (which is a `(N, geometric_dimension)` shape array of points).  Is there a reason you didn't use that format?

gusto/state.py
```diff
@@ -238,31 +306,27 @@ def dump(self, t=0, diagnostic_everydump=False, pickup=False):
...
-                    self.diagnostic_data[name]["total"].append(data)
+                    self.diagnostic_data[name][fn] = data
+            self.diagnostic_dump()
+
+            # calculate pointwise data
+            point_data = {}
+            for name, plist in self.output.point_data.iteritems():
+                # get points in the right format for the at function
+                points = [p for p in itertools.product(*plist)]
+                point_data[name] = [x.tolist() for x in self.field_dict[name].at(points)]
```
wence-: Field evaluation is currently collective across the MPI communicator.  So you need to arrange that this iteration happens in a consistent order.  I think rather than having a dict, you should have a list of pairs `(name, points)`.  Given that you always iterate over everything.

wence-: I think now that you're using netcdf, this could be: `np.asarray(self.field_dict[name].at(points))`

gusto/state.py
```diff
@@ -274,17 +338,46 @@ def dump(self, t=0, diagnostic_everydump=False, pickup=False):
                         chk.store(field)
                     chk.write_attribute("/", "time", t)
 
-            if diagnostic_everydump:
-                self.diagnostic_dump()
+            if(next(self.dumpcount) % self.output.dumpfreq) == 0:
```
wence-: Space after `if`.

gusto/state.py
```diff
@@ -219,12 +223,76 @@ def setup_dump(self, pickup=False):
...
+                    dim_names.append(name)
+                    grp.createDimension(name, npts[i])
+                    var = grp.createVariable(name, "f8", (name,))
+                    var[:] = plist[i]
+                # get tuple of dimensions that the output field varies over
+                dims = tuple(d for d in dim_names)
+                # finally, create field variable
+                grp.createVariable(field, "f8", dims)
+            # close the file
+            point_data.close()
```
wence-: This function is now getting very long.  Is it worthwhile splitting this setup part out into a separate method?

gusto/configuration.py
jshipton: A couple of reasons:
1) I think it is marginally easier for the user to specify the points as `list_of_xcoords`, `list_of_ycoords` (for example)
2) If the user gave an array of points I would have to think about numpy arrays when setting up the spatial dimensions in the netcdf file - this is definitely doable but my brain was starting to hurt at this point and lists seemed easier.

Happy to change if you think it best...

gusto/state.py
jshipton: If I have a list of pairs, what is the best way to replace 
`for name in self.output.point_data.keys()`? 
i.e. I only want the first item of each pair.

gusto/diagnostics.py
```diff
@@ -53,7 +55,10 @@ def l2(f):
 
     @staticmethod
     def total(f):
-        return assemble(f * dx)
+        if len(f.ufl_shape) == 0:
+            return assemble(f * dx)
+        else:
+            pass
```
wence-: I would prefer to `raise ValueError` here.

gusto/state.py
wence-: `for name, _ in self.output.point_data`

gusto/diagnostics.py
jshipton: I would definitely like to do something else with this diagnostic, but we can't do that because when the diagnostics are calculated they are calculated for every field (see line 323 of state.py) - so `total` will be passed a vector field at some point. I would rather just not call `total` on vector fields but I couldn't decide on the nicest way to do this - note it also affects the creation of variables in the netcdf file.

jshipton: I did think of having a separate list (in addition to available_diagnostics) for diagnostics that only apply to scalar fields but I wondered if it would generalise nicely to other exceptions (not that I can think of any examples!)

wence-: Hmm, so you could explicitly require that the user states which diagnostics they would like for each field name.

Or you could (sketch, don't take this as gospel), say something like:


```python
class DiagnosticBase(object, metaclass=...):

    @staticmethod
    @abstractmethod
    def computable(field):
        "Is this diagnostic computable on the provided field?"
        pass

class Total(DiagnosticBase):
    def computable(field):
        return len(field.ufl_shape) == 0

...

for diag in available_diagnostics:
    for field in fields:
        if not diag.computable(field)
            continue
       diagnostic_data.add(diag(field))
```

jshipton: I like that - I'll do something along those lines... thanks!

gusto/state.py
```diff
@@ -210,12 +214,74 @@ def setup_dump(self, pickup=False):
...
+        point_data.createDimension("time", None)
+        # create time variable so that we can set the values time values
+        times = point_data.createVariable("time", "f8", ("time",))
+        times.units = "seconds"
+        # create a group for each field - each group will have dimensions
+        # set according to the information in plist
+        for field, points in self.output.point_data:
+            grp = point_data.createGroup(field)
+            # start list of dimensions that point data will have
+            dim_names = ["time"]
```
wence-: Unused variable.

gusto/state.py
```diff
@@ -210,12 +214,74 @@ def setup_dump(self, pickup=False):
...
+            # get number of points in each direction
+            npts, dim = points.shape
+            grp.createDimension("points", npts)
+            grp.createDimension("geometric_dimension", dim)
+            var = grp.createVariable("points", points.dtype,
+                                     ("points", "geometric_dimension"))
+            var[:] = points
+            # finally, create field variable
+            grp.createVariable(field, self.fields(field).dat.data.dtype,
+                               ("time", "points"))
```
wence-: Use `self.fields(field).dat.dtype`.  That way you don't access the data (forcing evaluation).

gusto/state.py
```diff
@@ -229,31 +295,26 @@ def dump(self, t=0, diagnostic_everydump=False, pickup=False):
...
-                    self.diagnostic_data[name]["total"].append(data)
+                    self.diagnostic_data[name][fn] = data
+            self.diagnostic_dump()
+
+            # calculate pointwise data
+            point_data = {}
+            for name, points in self.output.point_data:
+                # get points in the right format for the at function
+                point_data[name] = np.asarray(self.fields(name).at(points))
+            self.pointwise_dump(point_data)
```
wence-: It seems that one could move the evaluation of the fields into the `pointwise_dump` function (similarly with the `diagnostic_dump`).  That way you have less "action at a distance".  I would also be tempted to pull these out as free functions.  Let me push a commit showing what I mean.

gusto/state.py
```diff
@@ -210,12 +316,26 @@ def setup_dump(self, pickup=False):
...
+
+        self.pointdata_output = PointDataOutput(pointdata_filename,
+                                                self.output.point_data,
+                                                self.output.dirname,
+                                                self.fields,
+                                                create=not pickup)
+        self.diagnostic_output = DiagnosticsOutput(diagnostics_filename,
+                                                   self.diagnostics,
+                                                   self.output.dirname,
+                                                   create=not pickup)
```
wence-: This is basically just attempting to make a smaller handbag...

gusto/state.py
```diff
@@ -52,6 +53,109 @@ def __iter__(self):
...
+        with Dataset(self.filename, "a") as dataset:
+            idx = dataset.dimension["time"].size
+            dataset.variables["time"][idx:idx + 1] = t
+            for name in self.diagnostics.fields:
+                field = mapping[name](state)
+                group = dataset.groups[name]
+                for dname in self.diagnostics.available_diagnostics:
+                    diagnostic = getattr(self.diagnostics, dname)
+                    var = group.variables[dname]
+                    var[idx:idx + 1] = diagnostic(field)
```
wence-: This is mostly just code movement from the previous commits.  But I think it is marginally clearer.  There's less action at a distance for one.

gusto/state.py
```diff
@@ -52,6 +53,109 @@ def __iter__(self):
...
+                    group.createVariable(diagnostic, np.float64, ("time", ))
+
+    def dump(self, diagnostic_fields, state, t):
+        """Dump diagnostics.
+
+        :arg diagnostic_fields: Iterable of :class:`DiagnosticField` objects.
+        :arg state: The :class:`State` at which to compute the diagnostic.
+        :arg t: The current time.
+        """
+        mapping = dict((df.name, df) for df in diagnostic_fields)
```
wence-: Convert the diagnostic fields list into a map from name to diagnostic_field object.

gusto/state.py
```diff
@@ -52,6 +53,109 @@ def __iter__(self):
...
+        with Dataset(self.filename, "a") as dataset:
+            idx = dataset.dimension["time"].size
+            dataset.variables["time"][idx:idx + 1] = t
+            for name in self.diagnostics.fields:
+                field = mapping[name](state)
```
wence-: OK, so now we ask for a `DiagnosticField` and immediately evaluate it to get the resulting `Function`.  Previously we evaluated the diagnostic fields up front, and relied on them magically being up to date in `State` when actually computing the diagnostics.

gusto/state.py
```diff
@@ -210,12 +316,26 @@ def setup_dump(self, pickup=False):
             field = Function(functionspaceimpl.WithGeometry(f.function_space(), mesh_ll), val=f.topological, name=name+'_ll')
             self.to_dump_latlon.append(field)
 
-    def dump(self, t=0, diagnostic_everydump=False, pickup=False):
+        # we create new netcdf files to write to, unless pickup=True, in
+        # which case we just need the filenames
+        diagnostics_filename = self.dumpdir+"/diagnostics.nc"
+
+        pointdata_filename = self.dumpdir+"/point_data.nc"
```
wence-: It's unclear to me why some of these directories live in `self.output`, others live in `self` (i.e. State) and so on.  So I haven't touched this here.

jshipton: The directory name is provided by the user and lives in ```self.output```. We automatically create this directory in a ```results``` subdirectory (see line 288) so the difference is that ```self.dumpdir``` has ```results/``` prepended to it - just saves the user typing it repeatedly and keeps the directory tidy. Happy to change if it doesn't make sense to other people though!

gusto/diagnostics.py
jshipton: I'm a bit confused - would these classes be defined within the current `Diagnostics` class?

PR #117, opened 2017-07-14
update estimate_eigenvalues -> esteig

dorugeber:
Per https://www.mcs.anl.gov/petsc/documentation/changes/36.html

This change is trivial and I could merge it straight in.  However, I want to raise a couple of points for discussion:

The first is that Chebyshev has been the default smoother in GAMG for some years now, so that line is superfluous... should we take it out, or leave it in, in case the default changes again in the future?

More seriously, this looks like a _lot_ of copy-paste solver-option code, and these are similar to what's inside `linear_solvers.py`, so some refactoring may be in order.


wence-:
I agree that there is a lot of copy-paste here.

Providing good default solver options is slightly hard, and I'm not sure where the status of all the different preconditioner setups is.  

Perhaps there should be a `solver_parameters` module one can import that provides default canned options for the various solvers.  The `linear_solvers` module can then use those as defaults: so that you can access them when creating and modifying.

PR #119, opened 2017-07-16
Add PV diagnostic

dorugeber:
Cherry-pick of @thomasgibson's commit (rather, his rewrite of @jshipton's commit), which has made its way into several branches, and some fixups.

This doesn't quite make the right FEEC PV (for RT or BDFM on triangles), but it's only a diagnostic variable, not prognostic, so probably doesn't matter.

Would prefer to see a proper initialisation rather than this slightly convoluted code, but that can come in a future diagnostics rewrite.


gusto/diagnostics.py
```diff
@@ -289,3 +290,62 @@ def __init__(self, state, name):
...
     def name(self):
         return self.field1+"_perturbation"
+
+
+class PotentialVorticity(DiagnosticField):
+    """Diagnostic field for potential vorticity."""
+
+    @property
+    def name(self):
+        return "potential_vorticity"
```
wence-: This could be a class attribute `name = "potential_vorticity"`

PR #121, opened 2017-07-17
Perturbed diagnostics

tommbendall:
This pull request is intended to provide a facility for making diagnostic fields that are perturbations of other diagnostic fields.

There are two sections of code for setting the base field. Either you can set a chosen reference profile, or the base field will be automatically calculated from the initial model state.

I've also edited the compressible eady test to use this feature.

### DIFF
--- a/examples/compressible_eady.py
+++ b/examples/compressible_eady.py
@@ -165,7 +165,7 @@ state.initialise({'u': u0, 'rho': rho0, 'theta': theta0})
 
 # set the background profiles
 state.set_reference_profiles({'rho': rho_b, 'theta': theta_b})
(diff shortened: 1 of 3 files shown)


jshipton:
I think this is superseded by https://github.com/firedrakeproject/gusto/pull/123 but please let me know if that PR doesn't do everything you'd like...

tommbendall:
There's still actually a facility I'd like to add: I want to be able to output an arbitrary diagnostic field as a perturbation from an arbitrary reference state. (Maybe it does this already and I've just misunderstood?)

For example I might want diagnostics that depend on moisture variables, and but I want to see these perturbed from a reference field that I have chosen.

Not sure whether this PR is a neat solution though! I'll close this and have a rethink.

PR #123, opened 2017-07-27
Diagnostic initialisation

jshipton:
This PR adds a `setup_diagnostics` method to state. It is called at the start of the time loop because this is when we know everything else has been set up. The `setup_diagnostics` method loops over the diagnostic fields classes and calls their setup method - they all have one because they all need to set up the actual diagnostic field. The base `DiagnosticFields` class has a `setup` method that sets up `self.field` on a DG0 function space - this is the default. Any fields requiring a different function space can set this up in their own `setup` method (e.g. see `VelocityX`). Any additional fields and solvers can also be set up here.

This has the bonus that now diagnostic fields are created through the state `FieldsCreator` class which means that they are available for other diagnostic fields (such as `Sum`) to act on. There is some basic checking to see that the fields exist.

Remove state.field_dict as no longer necessary now diagnostic fields are added to state.fields (and I remember how to iterate over a class).

Fix up ExnerPi diagnostic so that it can either calculate the full Exner pressue (i.e. based on the rho and theta fields) or the reference Exner pressure (i.e. based on the reference rho and theta fields). As long as you have both of these in your diagnostic_fields list, you can add "ExnerPi" to your perturbation_fields list.

Outstanding issues:
- diagnostic fields are only calculated when they are dumped out - this means that the data in the diagnostics_dict is incorrect if for diagnostic fields if you are not dumping at every timestep.


jshipton:
@tommbendall it would be great if you could see if this works ok for your perturbed ExnerPi diagnostic - I think it does as I have tried it in the compressible_eady example...

gusto/state.py
```diff
@@ -160,6 +160,21 @@ def __init__(self, mesh, vertical_degree=None, horizontal_degree=1,
...
+            f = Perturbation(name)
+            self.diagnostic_fields.append(f)
+
+        for name in self.output.steady_state_error_fields:
+            f = SteadyStateError(self, name)
+            self.diagnostic_fields.append(f)
+
+        for diagnostic in self.diagnostic_fields:
+            diagnostic.setup(self)
+            f = diagnostic(self)
```
dorugeber: what is this doing?

gusto/state.py
```diff
@@ -260,10 +251,10 @@ def dump(self, t=0, diagnostic_everydump=False, pickup=False):
             for name in self.diagnostics.fields:
                 for fn in diagnostic_fns:
                     d = getattr(self.diagnostics, fn)
-                    data = d(self.field_dict[name])
+                    data = d(self.fields(name))
                     self.diagnostic_data[name][fn].append(data)
-                if len(self.field_dict[name].ufl_shape) is 0:
-                    data = self.diagnostics.total(self.field_dict[name])
+                if len(self.fields(name).ufl_shape) is 0:
```
dorugeber: surely `=`, not `is`?

miklos1: Haha. `==`

dorugeber: closenough

PR #124, opened 2017-07-28
Python3

wence-:
Changes to make gusto Python 3 compatible once the Firedrake update lands.  I've opened this now because there's a design change as well as mindless code fixing.

To whit, I change the arguments in the Timestepper to take a list of `advected_fields` and `diffused_fields`.  Each entry is a pair `(field_name, scheme)`.

Previously, these were dicts, but I noticed they are always just iterated over (no lookups are done).  By changing to a list, the iteration order is consistent.

Python3 has hash-randomisation turned on by default: that means you're guaranteed that dicts on different MPI processes will have different iteration orders.

In some sense, this bit of the code was always broken, but you were just lucky.

The other option, to keep the same interface, is to use an `OrderedDict`.  But now users writing examples have to remember to use an `OrderedDict`, and terrifically bad things happen if they don't (or you have to sort and so forth internally).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,7 @@ pipeline {
     environment {
         PATH = "/usr/local/bin:/usr/bin:/bin"
         CC = "mpicc"
+        PYTHONHASHSEED="1243123"
     }
     stages {
         stage('Clean') {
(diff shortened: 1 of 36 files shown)


jshipton:
That's absolutely fantastic - thanks @wence-

PR #128, opened 2017-08-01
expunge Expression

jshipton:
Remove deprecated Expression.
test_dcmip_setup has been removed completely as it is out of date with Colin's current dcmip setup scripts

I would like to test this a bit more before merging but made this PR just incase anyone else had a fit of tidying and decided to tackle this!


examples/sw_williamson5.py
```diff
@@ -49,13 +49,23 @@
...
     uexpr = as_vector([-u_max*x[1]/R0, u_max*x[0]/R0, 0.0])
+    theta, lamda = latlon_coords(mesh)
     h0 = Constant(H)
     Omega = Constant(parameters.Omega)
     g = Constant(parameters.g)
-    Dexpr = Expression("h0 - ((R0 * Omega * u0 + pow(u0,2)/2.0)*(x[2]*x[2]/(R0*R0)))/g - (2000 * (1 - sqrt(fmin(pow(pi/9.0,2),pow(atan2(x[1]/R0,x[0]/R0)+1.0*pi/2.0,2)+pow(asin(x[2]/R0)-pi/6.0,2)))/(pi/9.0)))", h0=5960, R0=R0, Omega=Omega, u0=20.0, g=g)
-    bexpr = Expression("2000 * (1 - sqrt(fmin(pow(pi/9.0,2),pow(atan2(x[1]/R0,x[0]/R0)+1.0*pi/2.0,2)+pow(asin(x[2]/R0)-pi/6.0,2)))/(pi/9.0))", R0=R0)
+    Rsq = R0**2
+    RR = pi/9.
+    Rsq = RR**2
```
dorugeber: Something looks wrong here

tests/test_weightless_tracer.py
```diff
@@ -57,11 +56,12 @@ def setup_tracer(dirname):
...
-        Expression("sqrt(pow(x[0]-xc,2)+pow(x[1]-zc,2))" +
-                   "> rc ? 0.0 : 0.25*(1. + cos((pi/rc)*" +
-                   "(sqrt(pow((x[0]-xc),2)+pow((x[1]-zc),2)))))",
-                   xc=500., zc=350., rc=250.))
+    xc = Constant(500.)
+    zc = Constant(350.)
+    rc = Constant(250.)
+    x = SpatialCoordinate(mesh)
+    r = sqrt((x[0]-xc)**2 + (x[1]-zc)**2)
+    theta_pert = conditional(r > rc, Constant(0.), 0.25*(1. + cos((pi/rc)*r)))
```
dorugeber: If `0.0` is unlikely to change to anything else, I don't think you need to wrap this in `Constant`.

jshipton: TBH I'm never sure when it's worth using Constant...

wence-: If you expect the *value* ever to change, use `Constant`.  If it's always just a literal that is unchanging, then just put it in "bare".

jshipton: Aha, in that case I will get rid of a lot of Constants...

examples/dense_bubble.py
```diff
@@ -118,10 +122,10 @@
...
     bcs = [DirichletBC(Vu, 0.0, "bottom"),
            DirichletBC(Vu, 0.0, "top")]
-    diffused_fields = [("u", InteriorPenalty(state, Vu, kappa=Constant(75.),
-                                             mu=Constant(10./delta), bcs=bcs)),
-                       ("theta", InteriorPenalty(state, Vt, kappa=Constant(75.),
-                                                 mu=Constant(10./delta)))]
+    diffused_fields = [("u", InteriorPenalty(state, Vu, kappa=75.,
+                                             mu=10./delta, bcs=bcs)),
+                       ("theta", InteriorPenalty(state, Vt, kappa=75.,
+                                                 mu=10./delta))]
```
wence-: Here's an example where you would use `Constant`, because `delta` changes through the loop.

jshipton: Good spot - thanks!

jshipton:
I'm happy with this now...

jshipton:
It passes tests and I've checked all the initial conditions of the examples by eye...

PR #103, opened 2017-05-30
tidy definition of zero BCs

dorugeber:
Replace various expressions for 0 with just 0.0.

PR #130, opened 2017-08-04
Moist dynamics

tommbendall:
This pull request fully integrates moisture into the dynamics. 

One main addition is changes to the momentum equations in `CompressibleForcing` and `CompressibleSolver`, which provide the option to include the effect of moisture upon the air density.

The other major change is the addition of a `moist_hydrostatic_balance` initialisation scheme, which sets up a saturated and hydrostatically balanced atmosphere for given profiles of the total water content and `theta_e`, the wet equivalent potential temperature.

I have also added an example, which replicates the moist rising bubble test case of Bryan and Fritsch, and two relevant diagnostics.


examples/moist_bf_bubble.py
```diff
@@ -0,0 +1,196 @@
...
+thetaeqn = EmbeddedDGAdvection(state, Vt, equation_form="advective")
+
+advected_fields = [('u', ThetaMethod(state, u0, ueqn)),
+                   ('rho', SSPRK3(state, rho0, rhoeqn)),
+                   ('theta', SSPRK3(state, theta0, thetaeqn)),
+                   ('water_v', SSPRK3(state, water_v0, thetaeqn)),
+                   ('water_c', SSPRK3(state, water_c0, thetaeqn))]
+
+# Set up linear solver
+schur_params = {'pc_type': 'fieldsplit',
```
jshipton: I think you probably don't want these solver options - the default ones are better.

examples/moist_bf_bubble.py
```diff
@@ -0,0 +1,196 @@
...
+compressible_forcing = CompressibleForcing(state, moisture=moisture)
+
+# diffusion
+bcs = [DirichletBC(Vu, 0.0, "bottom"),
+       DirichletBC(Vu, 0.0, "top")]
+
+diffused_fields = []
+
+if diffusion:
+    diffusion_fields = [('u', InteriorPenalty(state, Vu, kappa=Constant(60.),
```
jshipton: diffused_fields rather than diffusion_fields and also you can append

gusto/initialisation_tools.py
```diff
@@ -6,14 +6,14 @@
 from firedrake import MixedFunctionSpace, TrialFunctions, TestFunctions, \
     TestFunction, TrialFunction, SpatialCoordinate, \
     FacetNormal, inner, div, dx, ds_b, ds_t, ds_tb, DirichletBC, \
-    Function, Constant, assemble, \
+    Function, Constant, assemble, Expression, \
```
jshipton: noooooo! No More Expressions!

gusto/initialisation_tools.py
```diff
@@ -279,3 +287,153 @@ def calculate_Pi0(state, theta0, rho0):
...
+         exp(L_v * w_v / ((cp + c_pl * water_t) * T)) * dxp
+         - phi * w_v * dxp
+         + phi * w_sat * dxp
+         + cp * inner(v, w) * dxp
+         - cp * div(w * theta_v / (1.0 + water_t)) * pi * dxp
+         + psi * div(theta_v * v / (1.0 + water_t)) * dxp
+         + cp * inner(w, n) * pi_boundary * theta_v / (1.0 + water_t) * ds_b
+         + g * inner(w, state.k) * dxp)
+
+    bcs = [DirichletBC(Z.sub(3), Expression(("0.", "0.")), "top")]
```
jshipton: See PR https://github.com/firedrakeproject/gusto/pull/103 for how to avoid using Expression

gusto/linear_solvers.py
```diff
@@ -156,14 +159,26 @@ def V(u):
         dxp = dx(degree=(self.quadrature_degree))
         dS_vp = dS_v(degree=(self.quadrature_degree))
 
+        # introduce density potential temperatures
+        theta_rho = theta
```
jshipton: what are theta_rho and theta_rho_bar?

tommbendall: `theta_rho` is the "density potential temperature", θ_ρ defined as θ_v / (1 + total_water). When there is no water, this is just the same as the dry θ. The name is definitely confusing, I don't know what else to call it!

gusto/initialisation_tools.py
tommbendall: Ahhhh sorry!

examples/moist_bf_bubble.py
tommbendall: Thanks! I just copied from an old rising bubble case without knowing what I was doing...

gusto/linear_solvers.py
jshipton: I think it is especially confusing because elsewhere we use ```exner_rho``` to mean d(exner)/d(rho) i.e. the derivative of the Exner pressure w.r.t. rho so this kind of looks like d(theta/d(rho).

Can you not just keep it ```theta``` and then expand your comment "add moisture fields" to explain that when there is moisture, theta becomes blah...?

tommbendall: Sure, thanks!

tommbendall: Actually it isn't as simple as that, as in `forcing.py` we still need to use the original `theta` so don't want it overwritten by `theta/(1+water_total)`. How about just changing the name `theta_rho` to `theta_weighted` or `theta_water`? Variable naming is hard...

jshipton: In ```forcing.py```, can't you have
```
if self.moisture is not None:
   theta = whatever
else
   theta = theta0
```

jshipton: and yes, variable naming is a pain

tommbendall: haha of course, plain `theta` does work as well

gusto/initialisation_tools.py
```diff
@@ -107,6 +108,7 @@ def compressible_hydrostatic_balance(state, theta0, rho0, pi0=None,
     it at the bottom.
     :arg pi_boundary: a field or expression to use as boundary data for pi on
     the top or bottom as specified.
+    :arg moisture: the initial total water mixing ratio field.
```
jshipton: this doesn't match the name in the interface (water_t)

gusto/initialisation_tools.py
```diff
@@ -120,9 +122,15 @@ def compressible_hydrostatic_balance(state, theta0, rho0, pi0=None,
 
     cp = state.parameters.cp
 
+    # introduce density potential temperature
```
jshipton: is this another place where we can replace `theta_rho` with `theta` again?

gusto/initialisation_tools.py
```diff
@@ -279,3 +287,153 @@ def calculate_Pi0(state, theta0, rho0):
...
+    R_v = param.R_v
+    L_v0 = param.L_v0
+    T_0 = param.T_0
+    w_sat1 = param.w_sat1
+    w_sat2 = param.w_sat2
+    w_sat3 = param.w_sat3
+    w_sat4 = param.w_sat4
+    g = param.g
+
+    quadrature_degree = (5, 5)
```
jshipton: does this hard coded quadrature degree need a warning, similar to the one in the compressible linear solver?

jshipton: I think you should check the degree of the element, as in linear 99-101 of linear_solver.py, otherwise this warning will be triggered for all elements.

gusto/initialisation_tools.py
```diff
@@ -279,3 +287,156 @@ def calculate_Pi0(state, theta0, rho0):
...
+    L_v0 = param.L_v0
+    T_0 = param.T_0
+    w_sat1 = param.w_sat1
+    w_sat2 = param.w_sat2
+    w_sat3 = param.w_sat3
+    w_sat4 = param.w_sat4
+    g = param.g
+
+    VDG = state.spaces("DG")
+    if any(deg > 2 for deg in VDG.ufl_element().degree()):
```
tommbendall: Is this any better? I've tried to replicate the check from the compressible linear solver but I'm not sure I understand exactly what it does...

gusto/initialisation_tools.py
jshipton: Yes, that's better. The point is that you have hard coded a quadrature degree (because the one that Firedrake generates automatically is absurdly high).  This quadrature degree is presumably OK for the problems you're currently running but at some point in the future someone might run this code with higher order elements, in which case the quadrature degree you've hardcoded won't be sufficient. So what this does is grab the degree of the DG space and check if it is greater than 2 - this will trigger the warning.

PR #131, opened 2017-08-09
Solver params

jshipton:
This PR deals with the solver parameters for the linear solve. A lot of the vertical slice examples and tests were using outdated solver parameters when they should have been using the defaults provided in their linear_solver class. To make it easier to tweak these options, I have rejigged the linear solver classes a bit:

- the child classes now all call the base class __init__ method
- the base class defines an abstract property called default_solver_parameters - a dictionary of default solver parameters
- the __init__ method has optional args solver_parameters (default None) and overwrite_solver_parameters (default False). If the user provides solver_parameters then the defaults are either updated or replaced, depending on the value of overwrite_solver_paramters.

All vertical slice tests have been updated to use this. I have left the shallow water examples and tests as they are (so they currently fail) - I propose adding the hybridised preconditioner options from https://github.com/firedrakeproject/gusto/pull/112 as the default options - @thomasgibson what do you think?

### DIFF
--- a/gusto/linear_solvers.py
+++ b/gusto/linear_solvers.py
@@ -5,7 +5,7 @@ from firedrake import split, LinearVariationalProblem, \
     Function, MixedVectorSpaceBasis, VectorSpaceBasis, warning
 
 from gusto.forcing import exner, exner_rho, exner_theta
-from abc import ABCMeta, abstractmethod
+from abc import ABCMeta, abstractmethod, abstractproperty
 
 
 __all__ = ["CompressibleSolver", "IncompressibleSolver", "ShallowWaterSolver"]
@@ -18,37 +18,33 @@ class TimesteppingSolver(object, metaclass=ABCMeta):
     This is a dummy base class.
 
     :arg state: :class:`.State` object.
-    :arg params (optional): solver parameters
+    :arg solver_params (optional): solver parameters
+    :arg overwrite_solver_parameters: boolean, if True use only the
+    solver_parameters that have been passed in, if False then update
+    the default solver parameters with the solver_parameters passed in.
     """
 
-    def __init__(self, state, params=None):
+    def __init__(self, state, solver_parameters=None,
+                 overwrite_solver_parameters=False):
 
         self.state = state
 
-        if params is None:
-            self.params = {'pc_type': 'fieldsplit',
-                           'pc_fieldsplit_type': 'schur',
-                           'ksp_type': 'gmres',
-                           'ksp_max_it': 100,
-                           'ksp_gmres_restart': 50,
-                           'pc_fieldsplit_schur_fact_type': 'FULL',
-                           'pc_fieldsplit_schur_precondition': 'selfp',
-                           'fieldsplit_0_ksp_type': 'preonly',
-                           'fieldsplit_0_pc_type': 'bjacobi',
-                           'fieldsplit_0_sub_pc_type': 'ilu',
-                           'fieldsplit_1_ksp_type': 'preonly',
-                           'fieldsplit_1_pc_type': 'gamg',
-                           'fieldsplit_1_mg_levels_ksp_type': 'chebyshev',
-                           'fieldsplit_1_mg_levels_ksp_chebyshev_esteig': True,
-                           'fieldsplit_1_mg_levels_ksp_chebyshev_esteig_random': True,
-                           'fieldsplit_1_mg_levels_ksp_max_it': 1,
-                           'fieldsplit_1_mg_levels_pc_type': 'bjacobi',
-                           'fieldsplit_1_mg_levels_sub_pc_type': 'ilu'}
+        if solver_parameters is not None:
+            if overwrite_solver_parameters:
+                self.solver_parameters = solver_parameters
+            else:
+                self.default_solver_parameters.update(solver_parameters)
+                self.solver_parameters = self.default_solver_parameters
         else:
-            self.params = params
+            self.solver_parameters = self.default_solver_parameters
 
+        print(self.solver_parameters)
         # setup the solver
         self._setup_solver()
+
+    @abstractproperty
+    def default_solver_parameters(self):
+        pass
 
     @abstractmethod
     def solve(self):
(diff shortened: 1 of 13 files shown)


gusto/linear_solvers.py
```diff
@@ -18,38 +18,34 @@ class TimesteppingSolver(object, metaclass=ABCMeta):
...
+            if overwrite_solver_parameters:
+                self.solver_parameters = solver_parameters
+            else:
+                self.default_solver_parameters.update(solver_parameters)
+                self.solver_parameters = self.default_solver_parameters
         else:
-            self.params = params
+            self.solver_parameters = self.default_solver_parameters
 
+        print(self.solver_parameters)
```
wence-: Debugging print.

gusto/linear_solvers.py
```diff
@@ -18,38 +18,34 @@ class TimesteppingSolver(object, metaclass=ABCMeta):
...
-                           'fieldsplit_1_mg_levels_ksp_chebyshev_esteig_random': True,
-                           'fieldsplit_1_mg_levels_ksp_max_it': 1,
-                           'fieldsplit_1_mg_levels_pc_type': 'bjacobi',
-                           'fieldsplit_1_mg_levels_sub_pc_type': 'ilu'}
+        if solver_parameters is not None:
+            if overwrite_solver_parameters:
+                self.solver_parameters = solver_parameters
+            else:
+                self.default_solver_parameters.update(solver_parameters)
+                self.solver_parameters = self.default_solver_parameters
```
wence-: Careful, this doesn't do what you want (since `default_solver_parameters` is a class property).

I think you could do this:

```python
class Solver(...):
     def __init__(self, ..., solver_parameters=None):
        if solver_parameters is not None:
            if not overwrite:
               p = dict(self.solver_parameters)
               p.update(solver_parameters)
               solver_parameters = p
            
    @abstractproperty
    def solver_parameters(self):
        """"Solver parameters for this solver"""

class FooSolver(Solver):
    solver_parameters = {...}
```

The lookup order is (simplified)

```
getattr(instance, "solver_parameters")
getattr(instance.__class__, "solver_parameters")
```

So the instance setting overrides the class instance, but if you didn't set something on the instance then it gets looked up on the class.

jshipton: OK, cool - I'll also update the docstrings now I know this is OK - thanks!

gusto/linear_solvers.py
```diff
@@ -238,27 +231,23 @@ class IncompressibleSolver(TimesteppingSolver):
...
+        'ksp_type': 'gmres',
+        'pc_type': 'fieldsplit',
+        'pc_fieldsplit_type': 'additive',
+        'fieldsplit_0_pc_type': 'lu',
+        'fieldsplit_1_pc_type': 'lu',
+        'fieldsplit_0_pc_factor_mat_solver_package': 'mumps',
+        'fieldsplit_1_pc_factor_mat_solver_package': 'mumps',
+        'fieldsplit_0_ksp_type': 'preonly',
+        'fieldsplit_1_ksp_type': 'preonly'
+    }
```
thomasgibson: This is also something @wence- may have an opinion on. Should we be pushing to use the nested-dict syntax for solver options? For example, the options here can be written as
```(python)
default_solver_parameters = {'ksp_type': 'gmres',
                             'pc_type': 'fieldsplit',
                             'pc_fieldsplit_type': 'additive',
                             'fieldsplit_0': {'ksp_type': 'preonly',
                                              'pc_type' : 'lu',
                                              'pc_factor_mat_solver_package': 'mumps'},
                             'fieldsplit_1': {'ksp_type': 'preonly',
                                              'pc_type' : 'lu',
                                              'pc_factor_mat_solver_package': 'mumps'}}
```
Every sub-ksp (associated with `fieldsplit_x`) takes another dict of parameters. This, in my opinion, is a nice way to write out solver options as it seems a bit more readable. Not crucial, just a suggestion.

wence-: This is only a readability change.  I agree that it is probably easier to read.

gusto/linear_solvers.py
```diff
@@ -72,9 +68,29 @@ class CompressibleSolver(TimesteppingSolver):
...
+        'fieldsplit_1_mg_levels_ksp_type': 'chebyshev',
+        'fieldsplit_1_mg_levels_ksp_chebyshev_esteig': True,
+        'fieldsplit_1_mg_levels_ksp_chebyshev_esteig_random': True,
+        'fieldsplit_1_mg_levels_ksp_max_it': 1,
+        'fieldsplit_1_mg_levels_pc_type': 'bjacobi',
+        'fieldsplit_1_mg_levels_sub_pc_type': 'ilu'
+    }
+
+    def __init__(self, state, quadrature_degree=None, solver_parameters=None,
+                 overwrite_solver_parameters=False):
```
wence-: This is painful.  I wish ipython would just fix this.

gusto/linear_solvers.py
jshipton: yes, I agree - easier to read

gusto/linear_solvers.py
jshipton: Yes, I'll document the argument now the general idea is approved. Not sure what your last sentence of @wence- 's comment means though...

jshipton: I'll grab the options from your PR if that's OK - shallow water examples are currently broken on this branch because I haven't set any default solver options and it'd be silly to set old ones that I'm joust going to replace...

wence-: So in the built docs you inherit, but not with built in `help`

jshipton: Ah, I see - that is a bit annoying.

jshipton:
I think I've responded to everything... poke me if not!

thomasgibson:
Looks good. I'll check again on the Jenkins build finishes.

gusto/linear_solvers.py
```diff
@@ -69,12 +63,35 @@ class CompressibleSolver(TimesteppingSolver):
...
+        'pc_fieldsplit_schur_fact_type': 'FULL',
+        'pc_fieldsplit_schur_precondition': 'selfp',
+        'fieldsplit_0': {'ksp_type': 'preonly',
+                         'pc_type': 'bjacobi',
+                         'sub_pc_type': 'ilu'},
+        'fieldsplit_1': {'ksp_type': 'preonly',
+                         'pc_type': 'gamg',
+                         'mg_levels': {'ksp_type': 'chebyshev',
+                                       'ksp_chebyshev_esteig': True,
+                                       'ksp_chebyshev_esteig_random': True,
```
wence-: This is now spelt `ksp_chebyshev_esteig_noisy` and is `True` by default.

gusto/linear_solvers.py
```diff
@@ -18,38 +18,32 @@ class TimesteppingSolver(object, metaclass=ABCMeta):
...
-                           'fieldsplit_1_mg_levels_pc_type': 'bjacobi',
-                           'fieldsplit_1_mg_levels_sub_pc_type': 'ilu'}
-        else:
-            self.params = params
+        if solver_parameters is not None:
+            if not overwrite_solver_parameters:
+                p = self.solver_parameters
+                p.update(solver_parameters)
+                solver_parameters = p
+            self.solver_parameters = self.solver_parameters
```
wence-: `self.solver_parameters = solver_parameters`

jshipton: whoops!

gusto/linear_solvers.py
dorugeber: PETSc changing their parameter names and not advertising it well?  Well I never...

dorugeber:
This looks ready to go?

PR #133, opened 2017-08-10
Remove unused code

jshipton:
This PR removes two lots of unused code:

1) the NoForcing class, which is also broken.

2) the geopotential_form option - originally introduced because we thought it might help with the non hydrostatic mountain test case but it didn't. Please shout if you use this option! @colinjcotter @gwimmer @tommbendall


jshipton:
Approved by Colin via Slack.

PR #134, opened 2017-08-10
first draft of default models

jshipton:
WARNING: This is a fake PR! I would just like a bit of feedback (or just tell me to stop if this is a dreadful way of doing things...).

The point of this code is to make setting up a simulation much easier by reducing the amount of copy and pasting. As a bonus, we can potentially remove some stuff from `state` so that it really is just the current state.

I have written a `model` class that has a load of `abstractproperties` that the child classes define. Instantiating the model class updates any model parameters (e.g. if you want to run using a different gravitational acceleration) and sets up the model state. You then have the chance to amend the default prognostic equations for your fields, your advection schemes, your forcing and your linear solver before passing all this to the model `setup` method. This method either uses what you've passed in or instantiates the default classes specified - this is the bit that I'm not sure whether it is icky or not.

Incase I'm wasting my time, I have not wired everything in (notably the `diffused_fields`, `physics_list` and options such as the sponge layer. I have fixed up 2 tests: test_sw_triangle.py and test_sk_nonlinear.py so you can see that this does indeed make the simulation easier to set up and also that you can fairly easily amend the defaults. I have not fully taken advantage of moving things like `perp` and `k` out of state and I also have not updated the docs but there are some comments in the model class that might help describe what is going on.

Thoughts?


gusto/model.py
```diff
@@ -0,0 +1,228 @@
...
+
+    eqn_opts = defaultdict(dict)
+
+    def __init__(self,
+                 mesh,
+                 timestepping=None,
+                 output=None,
+                 parameters=None,
+                 diagnostics=None,
+                 diagnostic_fields=None):
```
wence-: OK, so a model creates a `State` object and then ties that together with a load of other stuff: advection schemes, diagnostic fields, etc...  The timestepper then uses the model to run the problem.  As commented below, I think with some judicious refactoring, there should be no need to inherit from `Model` for different model types.  Instead, one should be able to instantiate a model, poke properties of it into shape, and then hand that off to the Timestepper.

Then your `CompressibleEulerModel` and `ShallowWaterModel` can just be free functions that set up a `Model` instance appropriately.

One needs to be a little careful, because right now there's all sorts of action at a distance in the instantiation of a `Model`.  It makes `State`, but requires properties to be set in the instance that are not set in the constructor.  So either you are expecting that subclass constructors are called first (bad practice) or else that these are somehow class properties, rather than instance properties (it appears that the latter is the case).

I think this would make it also less confusing as to when the class instance has to provide instantiated objects (e.g. the parameters), and when it has to provide the constructor (e.g. `linear_solver`).

gusto/model.py
```diff
@@ -0,0 +1,228 @@
...
+    @abstractproperty
+    def default_field_equations(self):
+        pass
+
+    @abstractproperty
+    def default_advection_schemes(self):
+        pass
+
+    def setup(self, linear_solver=None, forcing=None, field_equations=None,
+              advection_schemes=None):
```
wence-: In general, these `setup` functions smell bad to me.  What appears to be going on is:

1. Create a `Model` instance (calls `__init__`)
1a. You can't use it for everything yet!
2a. Poke some things on the instance
2b. Poke some things on the member variables of the instance (e.g. `state`, to set up more stuff).
2c. You still can't use it!
3. Call `setup` on the model instance, in the process poking it some more. (e.g. providing advection schemes).
4. Now you can use it!


I think you could take advantage of `firedrake.utils.cached_property` to simplify a lot of stuff, and make it clearer what is going on.  By which I mean, `__init__` takes all things necessary to set up the instance, and then use `@cached_property` to provide "on-demand" computed properties (like the `linear_solver` and so forth).

That way you don't shove all of the complexity into this second `setup` phase, that I have to remember to call!

jshipton: Yes, I don't particularly like this structure either but I couldn't think of another way of achieving the following 2, slightly conflicting, aims:

1) I'd like a user (an MSc student, say) to be able to write a really simple example file that looks something like:
```
do stuff to set up geometry, timestepping and basic output

model = SomeModel(geometry, timestepping, output, ...)

do some stuff to initial conditions

stepper = timestepper(model)
stepper.run(t=0, t=tmax)
```

This is different from what we currently do because they don't have to setup a `fieldlist` and advection equations, advection schemes, forcing or linear solvers - we know what standard defaults should be for all that (depending on which equations we're solving).

2) Realistically, most people won't be running default model settings for _everything_ so I need to provide a way for them to change, say, the advection scheme for one field, or the form of the equation for a field. This is why there is a separate `setup` function - this enables changing just the bits you need to care about. The aim is to reduce the copy-pasting currently occurring in example file setup.

If I understand correctly you are suggesting I pass the equations, advection schemes, linear solver etc etc into `Model.__init__` and this can be done via some functions like this:

`ShallowWaterModel = Model(stuff)`

where `stuff` is a bunch of default arguments for the field list, linear_solver, forcing etc (i.e. replacing the class properties I currently have). So, I think that solves aim 1) above - but I don't see how to easily change just one thing (say I want to use MyFancyNewScheme as an advection scheme for the depth field, instead of the default SSPRK3 scheme) - how do I do that without just copy-pasting the whole interface again?

I don't see how the `firedrake.utils.cached_property` helps so maybe that's the bit of the puzzle that will help me understand what you mean?

wence-: > I don't see how the firedrake.utils.cached_property helps so maybe that's the bit of the puzzle that will help me understand what you mean?

This is just a way of "delaying" evaluation of properties (create on first use, rather than create in `__init__`).

> If I understand correctly you are suggesting I pass the equations, advection schemes, linear solver etc etc into `Model.__init__` and this can be done via some functions like this:

> `ShallowWaterModel = Model(stuff)`

> where stuff is a bunch of default arguments for the field list, linear_solver, forcing etc (i.e. replacing the class properties I currently have).

OK, I see.  The problem is that if I want to poke some part of the initialisation (say linear_solver) then I need a lot of stuff already set up (from the Model).

But actually, I sort of only need `state` to do this setup phase.  So would it make sense to split into two bits:

```python
def ShallowWaterState(mesh, ...):
    return State(...)

def CompressibleEulerState(mesh, ...):
    return State(...)

def ShallowWaterModel(state, ...):
    return Model(state, ...)

state = ShallowWaterState(mesh)

advected_fields = ...
model = ShallowWaterModel(state, advected_fields=advected_fields)

...
```
?

jshipton: yes, I see that makes things nicer because there is no need to call `model.setup`... let me try something like this...

jshipton:
P.S. I haven't updated any documentation - but I will if this is approved! Let me know if you need any comments/docs to understand what is actually going on though!

gusto/linear_solvers.py
```diff
@@ -37,8 +41,9 @@ def __init__(self, state, solver_parameters=None,
                 solver_parameters = p
             self.solver_parameters = solver_parameters
 
-        # setup the solver
-        self._setup_solver()
+    @cached_property
+    def _solvers(self):
+        pass
```
wence-: I think you can write this as an `@abstractproperty` to indicate the the subclass must provide it?

gusto/model.py
```diff
@@ -0,0 +1,136 @@
...
+                      forcing=None,
+                      advected_fields=None,
+                      diffused_fields=None,
+                      physics_list=None):
+
+    if parameters is None:
+        raise ValueError("Default parameters cannot be set for shallow water model. You must at least provide the mean depth via the ShallowWaterParameters configuration class.")
+
+    if timestepping is None:
+        raise ValueError("Default timestepping parameters cannot be set. You must at least provide the timestep, dt, via the TimesteppingParameters configuration class.")
```
wence-: If both of these things are required, best to make them required arguments, rather than optional.

gusto/transport_equation.py
```diff
@@ -311,28 +311,37 @@ class VectorInvariant(TransportEquation):
...
+                self.perp_u_upwind = (
+                    lambda q: Upwind('+')*perp(q('+')) + Upwind('-')*perp(q('-'))
+                )
+            elif physical_domain.on_sphere:
+                outward_normals = CellNormal(physical_domain.mesh)
+                self.perp_u_upwind = (
+                    lambda q: Upwind('+')*cross(outward_normals('+'), q('+'))
+                    + Upwind('-')*cross(outward_normals('-'), q('-'))
+                )
+            self.gradperp = lambda u: perp(grad(u))
```
wence-: I sort of don't like setting all these properties on the class in `__init__` that are basically just used as local variables in one other function.  The reason is that it adds to the cognitive load when reading and understanding the code.  I have to worry, is `self.gradperp` required at a later point many times (for example)?  In this case, it turns out no, I think that `gradperp`, `perp` and `perp_u_upwind` and `Upwind` are basically only used in the `advection_term` method.  So why not make them local variables there?

tests/test_advection.py
```diff
@@ -1,167 +1,224 @@
...
+    # (advective and continuity) and integrating by parts once and
+    # twice for each option - this give us 4 combinations.
+    eqnlist = []
+    eqnlist.append(AdvectionEquation(physical_domain, fspace, uspace))
+    eqnlist.append(AdvectionEquation(physical_domain, fspace, uspace, ibp="twice"))
+    eqnlist.append(AdvectionEquation(physical_domain, fspace, uspace, equation_form="continuity"))
+    eqnlist.append(AdvectionEquation(physical_domain, fspace, uspace, ibp="twice", equation_form="continuity"))
+
+    # setup the model with the above equations and the time_discretisation
+    field_eqns = {fname: eqn for (fname, eqn) in zip(fieldlist, eqnlist)}
```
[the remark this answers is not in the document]
jshipton: argh, dicts and parallel - one day I will remember this pitfall!

gusto/transport_equation.py
jshipton: ok, yes, that sounds nicer

tests/test_advection.py
```diff
@@ -1,167 +1,224 @@
...
-    f = f_dict["f"]
+def check_errors(f_dict, f_end, field_eqns, tol):
+    f_err = Function(f_end.function_space())
+    errors = {}
+    for name, field in f_dict.items():
+        f_err.assign(f_end - field)
+        errors[name] = max(abs(f_err.dat.data.min()), abs(f_err.dat.data.max()))
+    for name, err in errors.items():
+        if err > tol:
+            print("Test fails for field %s with options ibp = %s, continuity = %s" % (name, field_eqns[name].ibp, field_eqns[name].continuity))
```
jshipton: @wence- How about here if field_eqns isn't a dict? Is it safe to use `dict(field_eqns)` [note: I do this in `model.py` - see e.g. line 95] or is there a better way to search a list of tuples?

I am trying to print out some useful information because there are multiple asserts here and I want to know exactly which one failed. Maybe this is the wrong way to go about it - these used to be separate tests but they take a fairly long time and I wondered if testing all the advection options by using different fields within one simulation might speed things up by avoiding repeating the model setup.

gusto/linear_solvers.py
jshipton: instead of `@cached_property`?

thomasgibson: You can document this as `@abstractproperty`, and then all subclass implementations can mark the method as `@cached_property`. The only thing `@abstractproperty` will do is raise an exception if a subclass fails to implement a `_solvers` method.

jshipton: ah, I see - thanks @thomasgibson !

tests/test_advection.py
```diff
@@ -1,167 +1,224 @@
...
+    f_err = Function(f_end.function_space())
+    print(field_eqns)
+    errors = {}
+    for name, field in f_dict.items():
+        f_err.assign(f_end - field)
+        errors[name] = max(abs(f_err.dat.data.min()), abs(f_err.dat.data.max()))
+    for name, err in errors.items():
+        if err > tol:
+            print("Test fails for field %s with options ibp = %s, continuity = %s" % (name, dict(field_eqns)[name].ibp, dict(field_eqns[name]).continuity))
+    assert(all([err < tol for err in errors.values()]))
```
wence-: > @wence- How about here if field_eqns isn't a dict? Is it safe to use dict(field_eqns) [note: I do this in model.py - see e.g. line 95] or is there a better way to search a list of tuples?

dictifying the list of tuples for searching is fine.  I note as per usual that the input `f_dict` does not have consistent iteration order.

> I am trying to print out some useful information because there are multiple asserts here and I want to know exactly which one failed. Maybe this is the wrong way to go about it - these used to be separate tests but they take a fairly long time and I wondered if testing all the advection options by using different fields within one simulation might speed things up by avoiding repeating the model setup.

Does it?  I guess it probably does because advecting 3 fields is probably not 3x as expensive as advecting one.

Why not write:

```python
f_err = Function(f_end.function_space())
fields = dict(field_eqns)
failed = False
for name, field in f_list:
    f_err.assign(f_end - field)
    maxerr = numpy.abs(f_err.dat.data_ro).max()
    if maxerr > tol:
        print(...)
        failed = True
assert not failed
```

tests/test_advection.py
```diff
@@ -1,167 +1,224 @@
...
+    eqnlist.append(AdvectionEquation(physical_domain, fspace, uspace, ibp="twice"))
+    eqnlist.append(AdvectionEquation(physical_domain, fspace, uspace, equation_form="continuity"))
+    eqnlist.append(AdvectionEquation(physical_domain, fspace, uspace, ibp="twice", equation_form="continuity"))
+
+    # setup the model with the above equations and the time_discretisation
+    field_eqns = list(zip(fieldlist, eqnlist))
+    model = setup_model(state, physical_domain, timestepping, field_eqns, time_discretisation)
+
+    # run the model
+    f_dict = run(model, pi/2., fieldlist)
```
wence-: So `Timestepper.run` returns a dict of the fields you asked for in `fieldlist` if its not `None`.  That seems really odd.  Why not just do:
```python
fields = [(name, gettattr(state.fields, field)) for name in fieldlist]
```
here.

I note in passing that this syntax could be made nice with a small change to `FieldCreator` to something like:

```
class FieldCreator(OrderedDict):
    def __init__(self, fieldlist=None, xn=None, dumplist=None, pickup=True):
        super().__init__()
        for name, func in zip(fieldlist, xn.split()):
            self.add_field(func, name, name in dumplist, pickup)

    def add_field(self, field, name, dump, pickup):
        self[name] = field
        field.dump = dump
        field.pickup = pickup
        field.rename(name)
        return field
        
    def __getattr__(self, name):
        try:
            return self[name]
        except KeyError as e:
            raise AttributeError("Unknown field %s" % name) from e
    
    def __call__(self, name, space=None, dump=True, pickup=True):
        try:
            return self[name]
        except KeyError:
            f = Function(space)
            return self.add_field(f, name, dump, pickup)
    
    def __iter__(self):
        return iter(self.values())
```

now you can either do `state.fields[name]` or `getattr(state.fields, name)` or `state.fields.foo` or `state.fields["foo"]`

jshipton:
whoops - didn't mean to push to this branch! @wence- I think I will make a new branch and try to introduce some changes along these lines incrementally. I've just written a new time stepper so am rethinking this a bit in light of that.

wence-:
Ok, shall we close this PR?

PR #136, opened 2017-09-21
Added vorticity diagnostic field

jshipton:
- amended degree of function space for both vorticity and pv
- added test for vorticity and pv diagnostic fields (in the current test_sw_triangle test)


thomasgibson:
Perhaps we can take this opportunity to now let `PotentialVorticity` subclass from `Vorticity`, since the only thing that really needs to change is the `setup` method. Thoughts?

dorugeber:
Clearly `Vorticity` should be a base class that `RelativeVorticity` and `PotentialVorticity` inherit from :D :D :D

PR #137, opened 2017-09-28
Theta limiter

tommbendall:
This pull request provides a limiter that can be used on temperature space variables (such as moisture species).

It adds `limiters.py`, in which new limiters could be added, and a test that the limiter works.


gusto/advection.py
```diff
@@ -185,10 +187,17 @@ def solve_stage(self, x_in, stage):
     @embedded_dg
     def apply(self, x_in, x_out):
 
+        if self.limiter is not None:
+            self.limiter.apply(x_in)
         self.q1.assign(x_in)
         for i in range(3):
             self.solve_stage(x_in, i)
+            if i < 2:
```
jshipton: I think this would be tidier if we amended the `solve_stage` method. For stage==2 you could assign to q1 as in the other stages and then do `x_out.assign(self.q1)` on line 198. You could either put the application of the limiter in this loop (without the need for the `if i < 2`) or put it in the `solve_stage` method, whichever you think is nicest.

gusto/limiters.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def __init__(self, space):
+        """
+        Initialise limiter
+
+        :param space : FunctionSpace instance
+        """
+
+        self.Vt = space
+        self.Q1DG = FunctionSpace(self.Vt.mesh(), 'DG', 1)  # space with only vertex DOF
```
jshipton: Do you think we should check that the space is DG1xCG2?

tommbendall: I'm not sure how I would do that!

jshipton: There's an `is_cg` test somewhere in `transport_equation.py` which I just copied from somewhere in the Firedrake source, which checks the number of DOFs is as expected - we could move that to a `utility_functions.py` file so we can use it in both places. I think @colinjcotter doesn't like it, but at least if we have it defined in just one place we can change it if we think of a better way to do it! You can get the degree of the space using `space.ufl_element().degree()`.

thomasgibson: There is an alternative way to check the element continuity. If `W` is your DG1xCG2 `MixedElement`, then `W.ufl_element().sub_elements()` gives you the individual component elements. Each of those elements has a `sobolev_space()` method which tells you the Sobolev space the element is associated with. That is, If you write `DG1.ufl_element().sobolev_space().name`, you'll get `L2`. Similarly, `CG2.ufl_element().sobolev_space().name` returns `H1`.

If `W` is a `TensorProductElement`, this is simplified a bit. Calling `sobolev_space()` on the TPE gives you a `DirectionalH` sobolev space, which you can pass a spatial index to see the continuity in a particular direction. For example, `W.sobolev_space()[i].name` returns `L2` for i a horizontal index (0, 1 in 3D), and the vertical index (2 in 3D) returns `H1`.

Just an alternative you may want to experiment with!

gusto/limiters.py
```diff
@@ -0,0 +1,117 @@
...
+    weight[i][0] += 1.0;
+    }"""
+
+_average_kernel = """
+for (int i=0; i<vrec.dofs; ++i) {
+        vrec[i][0] += v_b[i][0]/weight[i][0];
+        }"""
+
+
+class ThetaLimiter(object):
```
thomasgibson: You might want to have a look at `limiter.py` and `vertex_based_limiter.py` in `firedrake/firedrake/slope_limiters`. With some modifications, your `ThetaLimiter` could inherit from either the `Limiter` class (or possibly `VertexBasedLimiter`). That way, your limiter conforms to the existing limiter setup in current Firedrake.

tommbendall: It used to do this in previous commits, but I ended up thinking that it wouldn't be appropriate. The Firedrake `Limiter` class has several `@abstractmethod`s that we don't use in this limiter.

PR #138, opened 2017-10-05
Checkpoint bug bypassed

gwimmer:
To avoid the h5 viewer error that appears after repeatedly opening a checkpoint file: move Dumbcheckpoint to setup_dump, so that it is only created once. Note that we can't use the "with" statement in dump in this case since it will close the checkpoint (in which case we would have to open it again each time,  leading to the original error).


gusto/state.py
```diff
@@ -267,16 +267,22 @@ def setup_diagnostics(self):
...
         if self.mesh.comm.rank == 0 and "pytest" not in self.output.dirname \
            and path.exists(self.dumpdir) and not pickup:
             raise IOError("results directory '%s' already exists" % self.dumpdir)
         self.dumpcount = itertools.count()
         self.dumpfile = File(outfile, project_output=self.output.project_fields, comm=self.mesh.comm)
+        self.chkpts = ["chkpt", "chkptbk"]
+        for name in self.chkpts:
+            chkpt_path = path.join(self.dumpdir, name)
+            chkpt = DumbCheckpoint(chkpt_path, mode=FILE_CREATE)
+            self.chkpts[self.chkpts.index(name)] = chkpt
```
wence-: You need to remember to close these files manually when you're done.  The garbage collector might do it, but it's safer to do so "by hand".

At the end of the timeloop do something like:
```
for chk in self.chkpts:
    chk.close()
```

gusto/state.py
```diff
@@ -267,16 +267,22 @@ def setup_diagnostics(self):
...
+        # setup checkpoint files
         self.dumpdir = path.join("results", self.output.dirname)
         outfile = path.join(self.dumpdir, "field_output.pvd")
         if self.mesh.comm.rank == 0 and "pytest" not in self.output.dirname \
            and path.exists(self.dumpdir) and not pickup:
             raise IOError("results directory '%s' already exists" % self.dumpdir)
         self.dumpcount = itertools.count()
         self.dumpfile = File(outfile, project_output=self.output.project_fields, comm=self.mesh.comm)
+        self.chkpts = ["chkpt", "chkptbk"]
+        for name in self.chkpts:
```
wence-: I'm confused about the two different checkpoint files.  It just looks you write the same data twice.  I know that's what was happening previously, but perhaps you can explain.

gusto/state.py
wence-: FWIW, I would write this loop as:
```python
self.chkpts = []
for name in ["chkpt", "chkptbk"]:
    fullpath = path.join(self.dumpdir, name)
    self.chkpts.append(DumbCheckpoint(fullpath, mode=FILE_CREATE)
```

gusto/state.py
```diff
@@ -345,15 +351,12 @@ def dump(self, t=0, pickup=False):
...
-                    # Dump all the fields to a checkpoint
-                    for field in self.to_pickup:
-                        chk.store(field)
-                    chk.write_attribute("/", "time", t)
-
+            for file in self.chkpts:
+                # Dump all the fields to a checkpoint
+                for field in self.to_pickup:
+                    file.store(field)
+                file.write_attribute("/", "time", t)
```
wence-: Prefer not to use builtin names (file).  Write `for chk in self.chkpts`?

wence-: I notice that you've changed the order in which the checkpoint files are written (previously it was `chkptbk` then `chkpt`, now the other way round).

gusto/state.py
gwimmer: As far as I remember this was done for the case if something goes wrong and the test run stops while writing to the checkpoint. If the checkpoint doesn't work one would then still have the other checkpoint to recover the data.

gusto/state.py
gwimmer: I wrote this so that one can guess right away what's supposed to be in the chkpts list (compared to just having an empty list), but then again looking at your loop I guess it's still completely clear what chkpts is supposed to contain, so I'll replace my loop by yours.

gusto/state.py
wence-: OK, this is a really bad way to do file rotation, FWIW.  One should do something like:

```
os.rename("current", "backup")
dump_checkpoint("current")
```

gwimmer: I'm not sure how to use os.rename without having to build a new checkpoint each time (leading to the original error). How about if I build one checkpoint in setup_dump, then just copy the file each time in dump after having written to it, using something like
''''
from shutil import copyfile
copyfile(path.join(self.dumpdir,"chkpt.h5"), path.join(self.dumpdir,"chkptbk.h5"))
''''
I can't really think of a scenario where this wouldn't produce at least one functioning backup file while the version of writing to two files in a row (as it is now) would.

wence-: So if I understand the problem, what you're trying to avoid is the simulation crashing overwriting some "good" checkpoint file.  So think about the time loop:

At the beginning you have no checkpoint files.
After step one, you either have one checkpoint file (`current`) or none (if the simulation crashed).

Now you come to step two and want to do a checkpoint.  So you first move `current` to `backup`, ensuring that if you now crash there is a backup checkpoint file from the previous dump.

Now you write to `current`.  If this succeeds, great, if not, then you still have the backup file.

This is similar to your suggestion, but doesn't require as much filesystem movement (`os.rename` doesn't move data, just metadata).

So the logic is like:

```python
while t < tend:
    if dump:
        if os.path.exists("checkpoint.h5"):
            os.rename("checkpoint.h5", "backup.h5")
        dump_checkpoint("checkpoint.h5")
```

gwimmer: Thanks for the clarification! As I understand it, I will run into the following issue with this:
Suppose I have one checkpoint file (```current```) after step one. Then os.rename moves the metadata, but not the data, renaming the file to backup.h5.

But then with ```dump_checkpoint``` I will have to create a new data set, since I do not want to write to the existing data set (labelled as backup.h5). Hence, I create a new data set, naming it "checkpoint.h5", which is then renamed using ```os.rename``` in the next time step and so forth.

This means, if I understand this right, that if I want to use ```os.rename``` I will have to create a new data set using ```dump_checkpoint``` each time, which is what I want to avoid in the first place (since ```PETSc.ViewerHDF5().create``` in  ```Dumbcheckpoint.new_file()``` produces an error after 240 or so dump steps).

This is why I thought I should use ```copyfile``` instead, since this way I can avoid using ```dump_checkpoint``` each time.

Out of interest (this only makes sense if I understood this right): when using ```copyfile```, I will only ever have two data sets, since the backup one is replaced each time ```copyfile``` is used (according to their documentation). What happens in the ```os.rename``` scenario? I don't really see any data sets being deleted or replaced in this case.

wence-: OK, I see.  Yes, what you're proposing makes sense I think.  So you always write to `checkpoint.h5` and intermittently copy it to `backup.h5`.

jshipton:
Whatever you decide to do with the default checkpoint option, the lines in timpeloop.py where you do `state.chkpt.close()` will have to be preceded by `if state.output.checkpoint:`

gusto/state.py
```diff
@@ -356,12 +359,13 @@ def dump(self, t=0, pickup=False):
...
-            # Create second backup file
-            copyfile(path.join(self.dumpdir, "chkpt.h5"),
-                     path.join(self.dumpdir, "chkptbk.h5"))
+            if self.output.checkpoint:
+                for field in self.to_pickup:
+                    self.chkpt.store(field)
+                self.chkpt.write_attribute("/", "time", t)
+                # Create second backup file
+                copyfile(path.join(self.dumpdir, "chkpt.h5"),
+                         path.join(self.dumpdir, "chkptbk.h5"))
```
wence-: This copy should be done on a single rank.  Also, I'm not sure you're guaranteed filesystem consistency when you call this copy.

.gitignore
```diff
@@ -55,3 +55,4 @@ docs/_build/
 
 # PyBuilder
 target/
+.DS_Store
```
wence-: You're ignoring this file, but also you added it in your initial commit!  You should remove that add from the commit.  (In general, don't do git add -A).

gwimmer: Yes, sorry, I got a bit confused about this. I'm using github desktop at the moment, which seems to alway do git add -A unless I specifically remove files I don't want to commit.

gusto/state.py
gwimmer: I'm not sure I understand filesystem consistency. Do you mean this as in "what happens if the system crashes while the copying is taking place, could both files then be lost"? If this is the case then I don't really know how to rewrite the code other than how it was originally, with two files being written to seperatly.

wence-:
You'll need to either rebase this on top of current master, or else merge master into this branch (and fix conflicts) before merging.  I think the latter will probably be easier.

Prod when that is done, because that will invalidate the review.

PR #139, opened 2017-10-10
Output options

jshipton:
This PR does the following:

- Add option to output diagnostics - previously this was automatically done, and the default is for it to happen, but it might be handy to be able to turn it off (we needed to do this for debugging purposes and also it might be necessary if anyone has trouble with netcdf output).

- Add option to disable checkpointing - maybe we don't need this after Golo's PR goes through but it might still be useful.

- Make creation of point_data.nc file depend on whether the user specifies some points at which to dump data. This means that we don't unnecessarily create an empty file.

- The unlimited time dimension in the point_data.nc file seemed to be causing problems with parallel output so I have just set its length to be `int(tmax/dt) + 1` and this fixes things. Netcdf should be able to do this in parallel and it doesn't seem to be a problem with the diagnostics.nc file...

- Fix some minor bugs in plot_1D_profile.py.

### DIFF
--- a/gusto/configuration.py
+++ b/gusto/configuration.py
@@ -44,6 +44,7 @@ class OutputParameters(Configuration):
     dumplist = None
     dumplist_latlon = []
     dump_diagnostics = True
+    checkpoint = False
     dirname = None
     #: Should the output fields be interpolated or projected to
     #: a linear space?  Default is interpolation.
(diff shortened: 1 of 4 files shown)


jshipton:
P.S. This seems to massively reduce the time taken to run tests.

PR #140, opened 2017-10-10
Fix dq warnings

jshipton:
I wondered why there were so many warnings:
`UFL:WARNING Discontinuous Lagrange element requested on interval * interval, creating DQ element.`

Turns out we're creating a function space each timestep in order to calculate the rms value of field for the diagnostics file.

I'm sure there's a better way of fixing this... here I've just added a `setup` method to the `Diagnostics` class so that we can calculate the `area` field once. Then, if I understand correctly, `rms` can no longer be a `@staticmethod`.

Suggestions? (@wence- )

### DIFF
--- a/gusto/diagnostics.py
+++ b/gusto/diagnostics.py
@@ -17,6 +17,12 @@ class Diagnostics(object):
 
         self.fields = list(fields)
 
+    def setup(self, state):
+        # set up area computation
+        V = state.spaces("DG0", state.mesh, "DG", 0)
+        c = Function(V).assign(1.)
+        self.area = assemble(c*dx)
+
     def register(self, *fields):
 
         fset = set(self.fields)
@@ -44,12 +50,8 @@ void maxify(double *a, double *b) {
 """, "maxify"), f.dof_dset.set, fmax(op2.MAX), f.dat(op2.READ))
         return fmax.data[0]
 
-    @staticmethod
-    def rms(f):
-        V = FunctionSpace(f.function_space().mesh(), "DG", 1)
-        c = Function(V)
-        c.assign(1)
-        rms = sqrt(assemble(dot(f, f)*dx)/assemble(c*dx))
+    def rms(self, f):
+        rms = sqrt(assemble(dot(f, f)*dx)/self.area)
         return rms
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


gusto/diagnostics.py
```diff
@@ -44,12 +50,8 @@ def max(f):
...
         return fmax.data[0]
 
-    @staticmethod
-    def rms(f):
-        V = FunctionSpace(f.function_space().mesh(), "DG", 1)
-        c = Function(V)
-        c.assign(1)
-        rms = sqrt(assemble(dot(f, f)*dx)/assemble(c*dx))
+    def rms(self, f):
+        rms = sqrt(assemble(dot(f, f)*dx)/self.area)
```
wence-: Can you not just use `CellVolume`?

```python
def rms(f):
    return sqrt(assemble(dot(f, f)/CellVolume(f.ufl_domain()) * dx))
```

jshipton: Awesome - knew you'd have the answer! I did look at the `CellVolume` function but couldn't figure out what to give it. I'll close this and make a new PR to avoid all the junk commits...

jshipton: TypeError: unsupported operand type(s) for *: 'MeshGeometry' and 'Measure'

jshipton: ah, bracketing issue - sorted!

jshipton: OK, now I'm confused by the answer - see ipython output below:
```
In [9]: mesh=RectangleMesh(100, 100, 10, 1)

In [10]: V=FunctionSpace(mesh, "DG", 1)

In [11]: f=Function(V)

In [12]: c=Function(V).assign(1.)

In [13]: assemble(c*dx)
Out[13]: 10.000000000000732

In [14]: assemble(CellVolume(f.ufl_domain())*dx)
Out[14]: 0.0049999999999998362
```
The first seems to give the total area (which is what I want); the second seems to give the area of an element(?). What am I not understanding?!

dorugeber: Every cell has area 0.5*10/10,000 = 5e-4.  The area of the mesh is 10.

You have integrated the "function" 5e-4 over a domain of area 10, giving you 5e-3.

jshipton: Looks like `CellVolume` is not what I want then.

wence-: Oh wait, I was confused.  You're trying to compute the volume of the mesh.  Just do `assemble(1*dx(domain=f.ufl_domain())`

PR #143, opened 2017-10-16
Hydrostatic switch

gwimmer:
The hydrostatic switch is built as a horizontal projection P in state and appears in transport_equation, forcing, and linear_solvers. In transport equation, it is implemented for the VectorInvariant and EulerPoincare schemes (except for ibp /= “once”).

In the hydrostatic case, the forcing differs in the explicit and implicit stages. To incorporate this, I generalised the sponge layer term ```mu_alpha``` to ```stage```.

The hydrostatic discretisation initially produces spurious waves in the presence of orography, which can be avoided by an initial fully implicit time step. To do this, I built a method called ```parameter_update``` in state, which can then be defined accordingly in the example file (in my case 67-71 in h_mountain) and is called at each time step in the time loop.

In the hydrostatic case, the default linear solver behaves badly, and so I am currently using an LU factorisation as my solver parameters. This will hopefully be replaced by the hybridised solver.

Finally, to check for hydrostatic balance, I built a “HydrostaticImbalance” diagnostic.


gusto/diagnostics.py
```diff
@@ -270,6 +270,42 @@ def compute(self, state):
...
+
+            bcs = [DirichletBC(space, 0.0, "bottom"),
+                   DirichletBC(space, 0.0, "top")]
+
+            imbalanceproblem = LinearVariationalProblem(a, L, self.field, bcs=bcs)
+            self.imbalance_solver = LinearVariationalSolver(imbalanceproblem)
+
+    def compute(self, state):
+        self.imbalance_solver.solve()
+        return self.field[1]
```
wence-: The y-component of the imbalance?

gusto/state.py
```diff
@@ -194,6 +195,7 @@ def __init__(self, mesh, vertical_degree=None, horizontal_degree=1,
 
         self.Omega = Coriolis
         self.mu = sponge_function
+        self.h = hydrostatic
```
wence-: `h` feels like a very bad name for this attribute.  A naive reader might think that `h` is somehow related to mesh spacing?  Why not just call it `hydrostatic`?

gusto/state.py
```diff
@@ -249,9 +251,19 @@ def __init__(self, mesh, vertical_degree=None, horizontal_degree=1,
             if dim == 2:
                 self.perp = lambda u: as_vector([-u[1], u[0]])
 
+        # project test function for hydrostatic case
+        if self.h:
+            self.P = lambda u: u - self.k*inner(u, self.k)
+        else:
+            self.P = lambda u: u
```
wence-: Again, `P` is a pretty impenetrable name.

gusto/transport_equation.py
```diff
@@ -366,18 +367,18 @@ def advection_term(self, q):
             both = lambda u: 2*avg(u)
 
             L = (
-                inner(q, curl(cross(self.ubar, self.test)))*dx
-                - inner(both(self.Upwind*q),
+                inner(self.P(q), curl(cross(self.ubar, self.test)))*dx
+                - inner(both(self.Upwind*self.P(q)),
```
wence-: If `q` should always be projected, better to apply the projection when you define `q`.

gusto/forcing.py
```diff
@@ -89,7 +94,10 @@ def forcing_term(self):
         L = self.scaling * L
         # sponge term has a separate scaling factor as it is always implicit
         if self.sponge:
-            L -= self.mu_scaling*self.sponge_term()
+            L -= self.stage*self.state.timestepping.dt*self.sponge_term()
+        # hydrostatic term has no scaling factor
+        if self.hydrostatic:
+            L += self.hydrostatic_term()
```
wence-: OK, so in the hydrostatic case, in the first step you add `-<u0 . k, test . q>` and then in the next step you add `<u0 . k, test . q>`.

Whereas the sponge term is only added on the second step.

wence-: Is it clear that you want the implicitness of the sponge term to be identical to the hydrostatic term?

gusto/forcing.py
```diff
@@ -120,9 +128,9 @@ def apply(self, scaling, x_in, x_nl, x_out, **kwargs):
         """
         self.scaling.assign(scaling)
         self.x0.assign(x_nl)
-        mu_scaling = kwargs.get("mu_alpha")
-        if mu_scaling is not None:
-            self.mu_scaling.assign(mu_scaling)
+        stage = kwargs.get("stage")
+        if stage is not None:
+            self.stage.assign(stage)
```
wence-: `stage` is a bad name for this, I think.

gusto/state.py
```diff
@@ -249,9 +251,19 @@ def __init__(self, mesh, vertical_degree=None, horizontal_degree=1,
...
+
         #  Constant to hold current time
         self.t = Constant(0.0)
 
+    def parameter_update(self):
+        # method to update parameters if needed
```
wence-: This is not how you write a docstring, you write a docstring like:
```python
def parameter_update(self):
    """Do foo to the parameters.

More long-winded explanation here."""
    implementation_here
```

In particular, this comment gives me no indication as to which parameters should be updated, when it might be needed, or anything else.

gusto/timeloop.py
```diff
@@ -110,12 +111,13 @@ def run(self, t, tmax, pickup=False):
...
                 print("STEP", t, dt)
 
+            state.parameter_update()
             t += dt
             state.t.assign(t)
 
             with timed_stage("Apply forcing terms"):
                 self.forcing.apply((1-alpha)*dt, state.xn, state.xn,
-                                   state.xstar, mu_alpha=mu_alpha[0])
+                                   state.xstar, stage=stage[0])
```
wence-: So this flag is actually selecting whether you are doing an implicit or explicit step of forcing?  Why not do that:

```
self.forcing.apply(...., explicit=True)
```

And then below

`self.forcing.apply(..., explicit=False)`

Then you don't need any of this action at a distance with magic "stage" variables saying how terms are modified.

examples/h_mountain.py
```diff
@@ -0,0 +1,179 @@
...
+              diagnostic_fields=diagnostic_fields)
+
+
+# Off-centring parameter is set to alpha=1.0 for first time step
+# to omit spurious orography waves, then alpha=0.5
+def implicit_first_step():
+    if state.t.dat.data_ro[0] == dt:
+        state.timestepping.alpha = 0.5
+
+state.parameter_update = implicit_first_step
```
wence-: Rather than closing over `state` in the `parameter_update` method, you should change the API so that it is clear what `parameter_update` can change.  You could do this by making this method take an argument.

A better way to do this would be to pass the update function in as an argument to the `State` constructor.

Then you could have something like:
```
class State(...):

    def __init__(self, ..., parameter_update=None):
       ...
       if parameter_update is not None:
           self.parameter_update = parameter_update
       else:
           self.parameter_update = lambda state: pass

      ...
     def run(self, ...):
        while ...:
             self.parameter_update(self)
             ...
```

examples/h_mountain.py
```diff
@@ -0,0 +1,179 @@
...
+              parameters=parameters,
+              diagnostics=diagnostics,
+              fieldlist=fieldlist,
+              diagnostic_fields=diagnostic_fields)
+
+
+# Off-centring parameter is set to alpha=1.0 for first time step
+# to omit spurious orography waves, then alpha=0.5
+def implicit_first_step():
+    if state.t.dat.data_ro[0] == dt:
```
wence-: Careful comparing floating point values exactly.  You might be OK here, but it could be incidental.

examples/h_mountain.py
wence-: Although having looked at this.  This update function is not a property of the `State`, but rather the time loop object.  So the update callback should live there.

examples/h_mountain.py
```diff
@@ -0,0 +1,179 @@
...
+          "fieldsplit_1_ksp_monitor_true_residual": True,
+          'fieldsplit_1_pc_type': 'bjacobi',
+          'fieldsplit_1_sub_pc_type': 'ilu'}
+Pi = Function(Vr)
+rho_b = Function(Vr)
+compressible_hydrostatic_balance(state, theta_b, rho_b, Pi, top=True, pi_boundary=0.5, params=params)
+
+
+def min(f):
+    fmin = op2.Global(1, [1000], dtype=float)
```
wence-: OK, this kind of bad code suggests we should implement these callbacks in PyOP2.  Because this is so fragile.  If all the values in `f` are larger than 1000, you get the wrong answer, for example.

examples/h_mountain.py
```diff
@@ -0,0 +1,179 @@
...
+          'fieldsplit_1_ksp_max_it': 5,
+          "fieldsplit_1_ksp_monitor_true_residual": True,
+          'fieldsplit_1_pc_type': 'bjacobi',
+          'fieldsplit_1_sub_pc_type': 'ilu'}
+Pi = Function(Vr)
+rho_b = Function(Vr)
+compressible_hydrostatic_balance(state, theta_b, rho_b, Pi, top=True, pi_boundary=0.5, params=params)
+
+
+def min(f):
```
wence-: Prefer not to override builtin methods.

gusto/diagnostics.py
wence-: Note this is almost certainly not a `Function`, but rather a ufl expression?

gusto/transport_equation.py
```diff
@@ -407,5 +408,5 @@ class EulerPoincare(VectorInvariant):
 
     def advection_term(self, q):
         L = super().advection_term(q)
-        L -= 0.5*div(self.test)*inner(q, self.ubar)*dx
+        L -= 0.5*div(self.test)*inner(self.P(q), self.ubar)*dx
```
wence-: So you only need to project the test function for this one term, and not any of the others?

wence-: Would it make sense for the *caller* of these term constructors to ensure the test function is projected?

examples/h_mountain.py
gwimmer: I think this should be ok. After the first time step we have ```state.t.dat.data_ro[0] = float(dt)```, and python interprets ```float(dt) == dt``` as true both for dt as an integer and as a float.

gusto/diagnostics.py
gwimmer: The imbalance lives in the space of vector fields with zero horizontal component. The way it is written now only works in a 2D setting, so I should probably write something like ```self.field[-1]``` or ```self.field[dimension_of_problem -1]```

```self.field``` is the function I am writing my solution of the variational problem into (like in the ```Vorticity``` diagnostic), and I thought ```self.field[1]``` would then simply give me the vertical component of this function (like in the ```VelocityZ``` diagnostic)

This works, except of the output is a discontinuous field, but I thought we are more interested in the values than the field's shape (like in the ```CourantNumber``` diagnostic), so I left it like this.

File: gusto/forcing.py (hunk 5 above)
Comment (gwimmer) [reply gusto/forcing.py]: The hydrostatic terms here are a result of the missing time derivative in the vertical velocity equation, and so aren't really part of the forcing but rather part of what function the forcing should be added to (that's why they have no scaling factor). I then thought that since an explicit/implicit switch already existed, using ```(2*self.stage-1)``` would be the least intrusive way to implement this.

File: gusto/forcing.py (hunk 6 above)
Comment (gwimmer) [reply gusto/forcing.py]: Yes, Colin wasn't very happy with it either, I thought I'd keep it like this for now and wait for other suggestions here (such as yours of simply doing ```explicit=True```)

File: gusto/state.py (hunk 7 above)
Comment (gwimmer) [reply gusto/state.py]: Sorry for the bad docstring, how about 
```
def parameter_update(self, t):
       """
       Update parameters such as off-centring parameter,
       dumpfrequency at specific times during model run
        
       :arg t: the current model time
       """
       pass
```
with ``` state.parameter_update(t)``` in timeloop and 
```
def implicit_first_step(time):
    if time == dt:
        state.timestepping.alpha = 0.5
```
in the example file? I'll also rewrite the current docstrings of ```setup_diagnostics``` and ```setup_dump```.

Comment (wence-) [reply gusto/state.py]: See my comment in the test file about not closing over state.

File: gusto/transport_equation.py (hunk 4 above)
Comment (gwimmer) [reply gusto/transport_equation.py]: It's not projected in the case of integration by parts twice, because I have not tested this case. In theory the projection still should be applied to the all ```q```s in this case, so it's probably fine to leave all ```q```s as they are and do something like
```q = self.P(q)```
at the beginning of ```advection_term```.

File: gusto/transport_equation.py (hunk 13 above)
Comment (gwimmer) [reply gusto/transport_equation.py]: You are right, one could for example simply project the test and trial functions when ```eqn.advection_term``` is called in ```advection.py```, but I thought for now I should only implement the projection where I have tested it, i.e. in the Euler Poincare transport equation form.

File: examples/h_mountain.py (hunk 10 above)
Comment (wence-) [reply examples/h_mountain.py]: That's not what I mean.  Consider:

```python
dt = 0.1
x = 0
for i in range(10):
    x += dt
print(x == dt*10) => False
```

gwimmer: Hmm interesting, thanks for pointing this out! I wasn't aware of this issue before. In this case here it should still be ok since I think ```0+dt == dt``` is true for any time increment value dt. If the initial time isn't 0 this need not be the case anymore, so maybe in this case one should write something like
```
t_init = 0

def implicit_first_step(time):
    if time >= t_init + dt:
        state.timestepping.alpha = 0.5

stepper.run(t=t_init, tmax=tmax)
```

examples/h_mountain.py
```diff
@@ -0,0 +1,170 @@
...
+columns = res*12  # number of columns
+L = 240000.
+m = PeriodicIntervalMesh(columns, L)
+
+# build volume mesh
+H = 50000.  # Height position of the model top
+ext_mesh = ExtrudedMesh(m, layers=nlayers, layer_height=H/nlayers)
+Vc = VectorFunctionSpace(ext_mesh, "DG", 2)
+coord = SpatialCoordinate(ext_mesh)
+x = Function(Vc).interpolate(as_vector([coord[0], coord[1]]))
```
tommbendall: Do you still need lines 23 and 24 since below you have 
`x, z = SpatialCoordinate(ext_mesh)`?

gwimmer: Hmm you are probably right. I didn't really have a proper look at this part since I simply copied it from nh_mountain.

gusto/forcing.py
```diff
@@ -45,13 +45,14 @@ def __init__(self, state, euler_poincare=True, linear=False, extra_terms=None, m
...
         self.sponge = state.mu is not None
+        self.hydrostatic = state.hydrostatic
         self.topography = hasattr(state.fields, "topography")
         self.extra_terms = extra_terms
         self.moisture = moisture
 
         # some constants to use for scaling terms
         self.scaling = Constant(1.)
-        self.mu_scaling = Constant(1.)
+        self.impl = Constant(1.)
```
colinjcotter: Please can you add comment fields defining scaling and impl

wence-:
Can you remove the `.DS_store` file?

wence-:
Merged.

PR #144, opened 2017-10-16
Tidy test advection

jshipton:
The advection test is currently embarrassingly messy - to the point where some bits weren't even doing what they said they were. This PR sorts this out and, as a bonus, runs much faster.

- use pytest fixtures to provide things that are dependent on the `geometry` parameter
- reduce the number of parameters and instead advect many fields, each with different discretisation settings, for each test

It's still not the most beautiful code but I think it's better (and faster).


jshipton:
Some numbers:
On my laptop `test_advection.py` now takes 116.22s. Previously it was over 300s.

Issue #72, opened 2016-12-15
printing


jshipton:
Make printing more friendly in parallel and think about whether we need any other print statements  to help new users.

jshipton:
I'm already using the Firedrake warning function - should I be using the Firedrake logging module to print out user friendly logging statements?

jshipton:
thoughts? @wence- @dham

wence-:
So what you should do is make a `gusto` logger, and then set it to only print on rank 0.

```python
import logging

logger = logget.getLogger("gusto")

handler = logging.StreamHandler()
handler.setFormatter(logging.Formatter(fmt="%(name)s:%(levelname)s %(message)s"))
if COMM_WORLD.rank == 0:
    logger.addHandler(handler)
else:
    logger.addHandler(logging.NullHandler())
```

jshipton:
Thank you!

PR #147, opened 2017-10-30
Logging

jshipton:
This PR does 2 things:

1) creates a logger so that we can have different levels of output, and so that we only print on one processor when running in parallel.

2) rename `Verbose` in `OutputParameters` class to `log_level` for more flexibility. `log_level` can be `warning` (the default), `info` or `debug`. When set to `info`, basic information on the user defined inputs in printed out. When set to `debug`, `ksp_monitor_true_residual` is set to `True`.

Things I don't like about this code:

1) the output from the solvers is not controlled by the logger - does that matter?

2) I have passed `state` in to the `Limiter` class - it's a bit annoying to do this just to get the logger, but I'd rather users didn't have to think about passing the logger around. I could get the logger via `equation.state.logger` but that seems a bit convoluted. Advice?

### DIFF
--- a/gusto/advection.py
+++ b/gusto/advection.py
@@ -35,10 +35,11 @@ class Advection(object, metaclass=ABCMeta):
     :arg field: field to be advected
     :arg equation: :class:`.Equation` object, specifying the equation
     that field satisfies
-    :arg solver_params: solver_parameters
+    :arg solver_parameters: solver_parameters
     """
 
-    def __init__(self, state, field, equation=None, solver_params=None, limiter=None):
+    def __init__(self, state, field, equation=None, solver_parameters=None,
+                 limiter=None):
 
         if equation is not None:
 
@@ -50,10 +51,12 @@ class Advection(object, metaclass=ABCMeta):
             self.dt = self.state.timestepping.dt
 
             # get default solver options if none passed in
-            if solver_params is None:
+            if solver_parameters is None:
                 self.solver_parameters = equation.solver_parameters
             else:
-                self.solver_parameters = solver_params
+                self.solver_parameters = solver_parameters
+                if state.output.Verbose:
+                    self.solver_parameters["ksp_monitor_true_residual"] = True
 
             self.limiter = limiter
 
@@ -205,16 +208,16 @@ class ThetaMethod(Advection):
     Class to implement the theta timestepping method:
     y_(n+1) = y_n + dt*(theta*L(y_n) + (1-theta)*L(y_(n+1))) where L is the advection operator.
     """
-    def __init__(self, state, field, equation, theta=0.5, solver_params=None):
+    def __init__(self, state, field, equation, theta=0.5, solver_parameters=None):
 
-        if not solver_params:
+        if not solver_parameters:
             # theta method leads to asymmetric matrix, per lhs function below,
             # so don't use CG
-            solver_params = {'ksp_type': 'gmres',
-                             'pc_type': 'bjacobi',
-                             'sub_pc_type': 'ilu'}
+            solver_parameters = {'ksp_type': 'gmres',
+                                 'pc_type': 'bjacobi',
+                                 'sub_pc_type': 'ilu'}
 
-        super(ThetaMethod, self).__init__(state, field, equation, solver_params)
+        super(ThetaMethod, self).__init__(state, field, equation, solver_parameters)
 
         self.theta = theta
 
--- a/gusto/advection.py
+++ b/gusto/advection.py
@@ -151,12 +151,14 @@ class ExplicitAdvection(Advection):
     :arg equation: :class:`.Equation` object, specifying the equation
     that field satisfies
     :arg subcycles: (optional) integer specifying number of subcycles to perform
-    :arg solver_params: solver_parameters
+    :arg solver_parameters: solver_parameters
     :arg limiter: :class:`.Limiter` object.
     """
 
-    def __init__(self, state, field, equation=None, *, subcycles=None, solver_params=None, limiter=None):
-        super().__init__(state, field, equation, solver_params=solver_params, limiter=limiter)
+    def __init__(self, state, field, equation=None, *, subcycles=None,
+                 solver_parameters=None, limiter=None):
+        super().__init__(state, field, equation,
+                         solver_parameters=solver_parameters, limiter=limiter)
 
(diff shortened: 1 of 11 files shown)


jshipton:
BTW, this deals with issue #72 - did I do it right @wence- ?

jshipton:
Setting up `logger` in the `__init__.py` file didn't work - it wasn't found by the other files. Instead I have moved `logger` to the `configuration.py` file and it can be imported using `from gusto.configuration import logger` with no need to pass state around. Is that OK @colinjcotter @wence- ?

wence-:
I agree that the `logger` should not be attached to `State`.  You could put it in a `logging.py` file I suppose.  But in `configuration.py` is fine, probably.

gusto/state.py
```diff
@@ -252,6 +253,27 @@ def __init__(self, mesh, vertical_degree=None, horizontal_degree=1,
...
+        elif output.log_level == "warning":
+            logger.setLevel(logging.WARNING)
+        else:
+            raise ValueError("log_level must be one of: %s, %s, %s, not %s" % ("info", "debug", "warning", output.log_level))
+        handler = logging.StreamHandler()
+        handler.setFormatter(logging.Formatter(fmt="%(name)s:%(levelname)s %(message)s"))
+        if self.mesh.comm.rank == 0:
+            logger.addHandler(handler)
+        else:
+            logger.addHandler(logging.NullHandler())
```
wence-: I would do this setup at top-level in `configuration.py`.  Rather than in `state`.  Oh, because you need access to the communicator.  Why not have a free function that you can call instead of shoving even more code in `__init__` here?

gusto/state.py
```diff
@@ -252,6 +253,27 @@ def __init__(self, mesh, vertical_degree=None, horizontal_degree=1,
...
 
+        # setup logger
+        if output.log_level == "debug":
+            logger.setLevel(logging.DEBUG)
+        elif output.log_level == "info":
+            logger.setLevel(logging.INFO)
+        elif output.log_level == "warning":
+            logger.setLevel(logging.WARNING)
+        else:
+            raise ValueError("log_level must be one of: %s, %s, %s, not %s" % ("info", "debug", "warning", output.log_level))
```
wence-: Why not just allow someone to using the enums from `logging` directly?

gusto/advection.py
```diff
@@ -57,10 +58,12 @@ def __init__(self, state, field, equation=None, *, solver_params=None, limiter=N
             self.dt = self.state.timestepping.dt
 
             # get default solver options if none passed in
-            if solver_params is None:
+            if solver_parameters is None:
                 self.solver_parameters = equation.solver_parameters
             else:
-                self.solver_parameters = solver_params
+                self.solver_parameters = solver_parameters
+                if state.output.log_level == "debug":
```
wence-: This is now wrong, no?

gusto/forcing.py
```diff
@@ -26,7 +27,7 @@ def __init__(self, state, euler_poincare=True, linear=False, extra_terms=None, m
         self.state = state
         if linear:
             self.euler_poincare = False
-            warning('Setting euler_poincare to False because you have set linear=True')
+            logger.info('Setting euler_poincare to False because you have set linear=True')
```
wence-: Use logger.warning as before?

gusto/forcing.py
```diff
@@ -105,7 +106,14 @@ def _build_forcing_solvers(self):
             a, L, self.uF, bcs=bcs
         )
 
-        self.u_forcing_solver = LinearVariationalSolver(u_forcing_problem)
+        solver_parameters = {}
+        if self.state.output.log_level == "debug":
```
wence-: Again debug needs aktered

wence-: Ie use the DEBUG constant

gusto/forcing.py
```diff
@@ -207,7 +215,14 @@ def _build_forcing_solvers(self):
 
             theta_problem = LinearVariationalProblem(a, L, self.thetaF)
 
-            self.theta_solver = LinearVariationalSolver(theta_problem)
+            solver_parameters = {}
+            if self.state.output.log_level == "debug":
```
wence-: And here

gusto/forcing.py
```diff
@@ -275,7 +290,14 @@ def _build_forcing_solvers(self):
         divergence_problem = LinearVariationalProblem(
             a, L, self.divu)
 
-        self.divergence_solver = LinearVariationalSolver(divergence_problem)
+        solver_parameters = {}
+        if self.state.output.log_level == "debug":
```
wence-: And here

gusto/forcing.py
```diff
@@ -322,7 +344,14 @@ def _build_forcing_solvers(self):
             a, L, self.bF
         )
 
-        self.b_forcing_solver = LinearVariationalSolver(b_forcing_problem)
+        solver_parameters = {}
+        if self.state.output.log_level == "debug":
+            solver_parameters["ksp_monitor_true_residual"] = True
```
wence-: Ditto

gusto/forcing.py
```diff
@@ -370,7 +399,14 @@ def _build_forcing_solvers(self):
             a, L, self.thetaF
         )
 
-        self.theta_forcing_solver = LinearVariationalSolver(theta_forcing_problem)
+        solver_parameters = {}
+        if self.state.output.log_level == "debug":
```
wence-: Ditto

gusto/linear_solvers.py
```diff
@@ -37,6 +37,9 @@ def __init__(self, state, solver_parameters=None,
                 solver_parameters = p
             self.solver_parameters = solver_parameters
 
+        if state.output.log_level == "debug":
```
wence-: Ditto

gusto/transport_equation.py
```diff
@@ -67,6 +67,8 @@ def __init__(self, state, V, *, ibp="once", solver_params=None):
             self.solver_parameters = {'ksp_type': 'cg',
                                       'pc_type': 'bjacobi',
                                       'sub_pc_type': 'ilu'}
+        if state.output.log_level == "debug":
+            self.solver_parameters["ksp_monitor_true_residual"] = True
```
wence-: Ditto

PR #148, opened 2017-11-01
Implicit midpoint2

jshipton:
This PR implements an implicit midpoint timestepping scheme, with the forcing terms evaluated as part of the rhs of the advection scheme (i.e. there is no Strang splitting).

As there are now 2 semi implicit timestepping classes and some of the timestepping options and functions setup in state don't apply to both, I have moved things around a little:

1) I have removed the `maxi` and `maxk` options from `TimesteppingParameters` - these were used to set the maximum number of inner and outer iterations in the `CrankNicolson` timestepping class, but the `ImplicitMidpoint` class only requires one iteration loop. Instead, these parameters can be passed in to the `run` method of the timestepping class. They are given sensible default values (which no one seems to change anyway).

2) Since the two semi implicit time steppers have different numbers of intermediate function evaluations, I have removed the `_allocate_state` method from the `State` class. Instead the functions are set up in their respective timestepping classes. For example, all time steppers require an `xnp1` function to hold the values at the next time level, but only `CrankNicolson` requires an `xstar` function to hold the values after the first application of the forcing term.

There is also a bit of general tidying - e.g. removing the class name from the brackets after `super` in line with new Python 3 coding.

### DIFF
--- a/examples/dense_bubble.py
+++ b/examples/dense_bubble.py
@@ -26,7 +26,7 @@ for delta, dt in res_dt.items():
     mesh = ExtrudedMesh(m, layers=nlayers, layer_height=H/nlayers)
 
     fieldlist = ['u', 'rho', 'theta']
-    timestepping = TimesteppingParameters(dt=dt, maxk=4, maxi=1)
+    timestepping = TimesteppingParameters(dt=dt)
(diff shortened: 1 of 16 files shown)


wence-:
I thought this looked fine too.

jshipton:
There are better ways of doing this.

PR #150, opened 2017-11-08
Fallout

tommbendall:
This introduces a physics routine that performs advection of a rain field. The rain is designed to fall out of the bottom of the domain.

### DIFF
--- a/gusto/diagnostics.py
+++ b/gusto/diagnostics.py
@@ -448,14 +448,14 @@ class Precipitation(DiagnosticField):
             rain = state.fields('rain')
             rho = state.fields('rho')
             v = state.fields('rainfall_velocity')
-            phi = TestFunction(space)
+            self.phi = TestFunction(space)
             flux = TrialFunction(space)
             n = FacetNormal(state.mesh)
             un = 0.5 * (dot(v, n) + abs(dot(v, n)))
             self.flux = Function(space)
 
-            a = phi * flux * dx
-            L = phi * rain * un * rho
+            a = self.phi * flux * dx
+            L = self.phi * rain * un * rho
             if space.extruded:
                 L = L * (ds_b + ds_t + ds_v)
             else:
@@ -467,7 +467,7 @@ class Precipitation(DiagnosticField):
 
     def compute(self, state):
         self.solver.solve()
-        self.field.assign(self.field + self.flux)
+        self.field.assign(self.field + assemble(self.flux * self.phi * dx))
         return self.field
 
 
--- a/gusto/diagnostics.py
+++ b/gusto/diagnostics.py
@@ -1,12 +1,12 @@
 from firedrake import op2, assemble, dot, dx, FunctionSpace, Function, sqrt, \
     TestFunction, TrialFunction, CellNormal, Constant, cross, grad, inner, \
-    LinearVariationalProblem, LinearVariationalSolver, exp
+    LinearVariationalProblem, LinearVariationalSolver, exp, FacetNormal, \
+    ds, ds_b, ds_v, ds_t
 from abc import ABCMeta, abstractmethod, abstractproperty
 from gusto.forcing import exner
 import numpy as np
 
-
-__all__ = ["Diagnostics", "CourantNumber", "VelocityX", "VelocityZ", "VelocityY", "Energy", "KineticEnergy", "CompressibleKineticEnergy", "ExnerPi", "Sum", "Difference", "SteadyStateError", "Perturbation", "PotentialVorticity", "Theta_e", "InternalEnergy", "RelativeVorticity", "AbsoluteVorticity", "ShallowWaterKineticEnergy", "ShallowWaterPotentialEnergy", "ShallowWaterPotentialEnstrophy"]
+__all__ = ["Diagnostics", "CourantNumber", "VelocityX", "VelocityZ", "VelocityY", "Energy", "KineticEnergy", "CompressibleKineticEnergy", "ExnerPi", "Sum", "Difference", "SteadyStateError", "Perturbation", "PotentialVorticity", "Theta_e", "InternalEnergy", "RelativeVorticity", "AbsoluteVorticity", "ShallowWaterKineticEnergy", "ShallowWaterPotentialEnergy", "ShallowWaterPotentialEnstrophy", "Precipitation"]
 
 
 class Diagnostics(object):
@@ -401,12 +401,47 @@ class Perturbation(Difference):
         return self.field1+"_perturbation"
 
 
+class Precipitation(DiagnosticField):
+    name = "Precipitation"
+
+    def setup(self, state):
+        if not self._initialised:
+            space = state.spaces("DG1", state.mesh, "DG", 1)
+            super(Precipitation, self).setup(state, space=space)
+
+            rain = state.fields('rain')
+            rho = state.fields('rho')
+            v = state.fields('rainfall_velocity')
+            phi = TestFunction(space)
+            flux = TrialFunction(space)
+            n = FacetNormal(state.mesh)
+            un = 0.5 * (dot(v, n) + abs(dot(v, n)))
+            self.flux = Function(space)
+            self.total_water = Function(space)
+
+            if space.extruded:
+                self.ds = ds_b + ds_t + ds_v
+            else:
+                self.ds = ds
+
+            a = phi * flux * dx
+            L = phi * rain * un * rho * self.ds
+
+            # setup solver
+            problem = LinearVariationalProblem(a, L, self.flux)
+            self.solver = LinearVariationalSolver(problem)
+
+    def compute(self, state):
+        self.solver.solve()
+        self.total_water.assign(self.total_water + self.flux)
+        return self.field.interpolate(self.total_water)
+
+
 class Vorticity(DiagnosticField):
     """Base diagnostic field class for vorticity."""
 
     def setup(self, state, vorticity_type=None):
         """Solver for vorticity.
-
         :arg state: The state containing model.
         :arg vorticity_type: must be "relative", "absolute" or "potential"
         """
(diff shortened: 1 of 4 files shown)


gusto/diagnostics.py
```diff
@@ -401,12 +401,47 @@ def name(self):
...
+        self.total_water.assign(self.total_water + self.flux)
+        return self.field.interpolate(self.total_water)
+
+
 class Vorticity(DiagnosticField):
     """Base diagnostic field class for vorticity."""
 
     def setup(self, state, vorticity_type=None):
         """Solver for vorticity.
-
```
dorugeber: ???

gusto/diagnostics.py
```diff
@@ -461,7 +496,6 @@ def setup(self, state):
         a weighted mass system to generate the
         potential vorticity from known velocity and
         depth fields.
-
```
dorugeber: ???

gusto/diagnostics.py
```diff
@@ -473,7 +507,6 @@ class AbsoluteVorticity(Vorticity):
 
     def setup(self, state):
         """Solver for absolute vorticity.
-
```
dorugeber: ???

gusto/diagnostics.py
```diff
@@ -485,7 +518,6 @@ class RelativeVorticity(Vorticity):
 
     def setup(self, state):
         """Solver for relative vorticity.
-
```
dorugeber: ???

gusto/physics.py
```diff
@@ -1,15 +1,16 @@
...
 
 
-__all__ = ["Condensation"]
+__all__ = ["Condensation", "Fallout"]
 
 
 class Physics(object, metaclass=ABCMeta):
     """
     Base class for physics processes for Gusto.
-
```
dorugeber: ???

gusto/physics.py
```diff
@@ -31,7 +32,6 @@ class Condensation(Physics):
     into liquid water and evaporation of liquid
     water into water vapour, with the associated
     latent heat changes.
-
```
dorugeber: ???

gusto/diagnostics.py
tommbendall: Ah sorry, it wasn't deliberate. I had copied and pasted this code over from a different branch that had other features and somehow copying into the emacs window has deleted these lines!

dorugeber:
FWIW, the reason I took issue to this is that it suggests you aren't running `git diff` before committing changes?  If so, you're not really "in control" of what you're actually committing.  @miklos1 would tell you to run `git status` and `git diff` like you have some OCD about it.

miklos1:
Yes, some OCDs should be called disciple, and those ones are a good thing.

Yes, if you don't check the diff before committing, then you basically don't know what you are doing.

gusto/transport_equation.py
```diff
@@ -178,6 +186,10 @@ def advection_term(self, q):
                             dot(self.ubar('+'), self.n('+'))*q('+'))
                       + inner(self.test('-'),
                               dot(self.ubar('-'), self.n('-'))*q('-')))*self.dS
+
+        if self.outflow and self.ibp is not None:
```
jshipton: what happens if `ibp` is not None and `self.inflow` is `True`? Should this warn the user that they have set incompatible options?

tommbendall: Ah have I written this a bit confusingly? This means that if `self.outflow is True` and if `ibp is not None` then we need to add the surface integrals on the edge of the domain.

And I think that if you set `ibp=None` then you get both inflow and outflow, so I think this is the only case we need to consider?

gusto/diagnostics.py
```diff
@@ -401,6 +401,42 @@ def name(self):
...
+            self.flux = Function(space)
+            self.total_water = Function(space)
+
+            if space.extruded:
+                self.ds = ds_b + ds_t + ds_v
+            else:
+                self.ds = ds
+
+            a = phi * flux * dx
+            L = phi * rain * un * rho * self.ds
```
wence-: So this is diagnosing the flux into the top and out of the bottom?

gusto/physics.py
```diff
@@ -104,6 +106,39 @@ def __init__(self, state):
...
+
+        # function spaces
+        Vt = self.rain.function_space()
+        Vu = state.fields('u').function_space()
+
+        # introduce sedimentation rate
+        # for now assume all rain falls at terminal velocity
+        terminal_velocity = 10  # in m/s
+        self.v = state.fields("rainfall_velocity", Vu)
+        self.v.project(as_vector([0, -terminal_velocity]))
```
wence-: No need to make a whole function for this, just do:

```python
self.v = as_vector([0, -terminal_velocity])
```

wence-: Oh, but then looking at the code, the `update_ubar` function requires fields, so OK.

gusto/transport_equation.py
wence-: The advection term as coded doesn't add any surface integrals if `ibp` is `None`.  So I'm confused like Jemma.

wence-: FWIW, I note that `self.inflow` is not a thing that exists.  So is `outflow` really the right name?

wence-: I think this would somehow be clearer if the "outflow" term were somehow added separately.  But that would necessitate much larger redesign.

The idea would follow something like what thetis does:

You have `Term` objects that implement `a` and `L`.

Then `Equation` objects don't implement the terms directly, but rather you call `Equation.add_term(TermInstance)`.

So you would have:

```python
class Term(object):
   ...

class OutflowTerm(Term):
    def residual(...):
       pass

class AdvectionEquation(Equation):
    ...

eq = AdvectionEquation(...)
if outflow:
   eq.add_term(OutflowTerm(...))
```

Or something.

But this is food for thought, rather than to be done here.

tommbendall: The short answer is that this only works if `self.outflow=True` and `ibp is not None`. I'll just make this clearer by adding a warning if someone sets `ibp=None`.

What the `outflow` integral is trying to do is to add the contribution of any outgoing flux, but not incoming. This should be because `self.un` is `u` on exterior facets when `u` is pointing outwards, whilst `self.un=0` when `u` is pointing inwards.

You're right, there isn't such a thing as `self.inflow`, although we could add it in the same way if we wanted to.

With the integration by parts, I had thought that when `ibp=None`, the whole surface integral is there inside the main integral, `test*inner(u, grad(q))*dx`. When you integrate by parts, you separate out the surface integral to
`-q*div(test*u)*dx + test*q*inner(u,n)*dF`,
where `dF` is over all facets. We then separate the facets into being exterior and interior, and then normally we use boundary conditions to neglect the integral on the exterior facets. So I had presumed that if someone was choosing to use `ibp=None` then they would be trying to take include flow through the surfaces, and so wouldn't need to use the `outflow` option (which restores only outflow and not inflow).

Probably there is a mistake in something above that I said above and that's why it's confusing you!

gusto/diagnostics.py
tommbendall: Yes, although I am expecting there only to be flux out and not in!

gusto/diagnostics.py
```diff
@@ -401,6 +401,42 @@ def name(self):
...
+            flux = TrialFunction(space)
+            n = FacetNormal(state.mesh)
+            un = 0.5 * (dot(v, n) + abs(dot(v, n)))
+            self.flux = Function(space)
+            self.total_water = Function(space)
+
+            if space.extruded:
+                self.ds = ds_b + ds_t + ds_v
+            else:
+                self.ds = ds
```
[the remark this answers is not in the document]
tommbendall: I think this was done because otherwise you are redefining `firedrake.ds` which gives a lint error, and assigning it to be a member seemed a convenient enough way to do that.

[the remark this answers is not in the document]
wence-: OK, yes, I see (I hate python's scoping rules).

Yeah, either use a different variable name or else maybe write:

```python
            a = phi * flux * dx
            L = phi * rain * un * rho
            if space.extruded:
                L = L*(ds_b + ds_t + ds_v)
            else:
                L = L*ds
```

Or similar?

gusto/diagnostics.py
```diff
@@ -438,6 +437,42 @@ def name(self):
...
+            a = phi * flux * dx
+            L = phi * rain * un * rho * self.ds
+
+            # setup solver
+            problem = LinearVariationalProblem(a, L, self.flux)
+            self.solver = LinearVariationalSolver(problem)
+
+    def compute(self, state):
+        self.solver.solve()
+        self.field.assign(self.total_water + self.flux)
```
wence-: I think this is wrong, `total_water` is always zero.

gusto/transport_equation.py
```diff
@@ -181,6 +183,13 @@ def advection_term(self, q):
                             dot(self.ubar('+'), self.n('+'))*q('+'))
                       + inner(self.test('-'),
                               dot(self.ubar('-'), self.n('-'))*q('-')))*self.dS
+
+        if self.outflow:
+            if self.ibp is not None:
+                L += self.test*self.un*q*self.ds
+            else:
+                raise ValueError("ibp=None is not compatible with outflow option")
```
wence-: I think I would prefer the error checking to happen in `__init__`, then you can just write:
```python
if self.outflow:
    L += ...
```
here.

gusto/transport_equation.py
```diff
@@ -149,16 +153,14 @@ class AdvectionEquation(TransportEquation):
...
         if equation_form == "advective" or equation_form == "continuity":
             self.continuity = (equation_form == "continuity")
         else:
             raise ValueError("equation_form must be either 'advective' or 'continuity'")
-        if vector_manifold:
-            self.vector_manifold = True
-        else:
-            self.vector_manifold = False
+        self.vector_manifold = vector_manifold
+        self.outflow = outflow
```
wence-: So check here with:
```python
if outflow and ibp is not None:
    raise ValueError(...)
```

PR #151, opened 2017-12-13
Prune dead solver options

thomasgibson:
Removes dead solver options in the linear shallow water solver. The hybridization preconditioner no longer builds a mass KSP for projections.


thomasgibson:
These options configured a mass matrix solve to perform projections of the residual into the hybridized spaces, and projections of the hybridized solutions into the original spaces. But these are no longer necessary. The preconditioner now uses facet-averaging to project discontinuous solutions and produce the right residual for the hybrid-problem, removing the need to build and solve a mass matrix system.

PR #153, opened 2018-01-09
Thermodynamic expressions

tommbendall:
This pull request adds a series of functions that return expressions for thermodynamic variables.
The hope is that this should simplify things when more variables are added!


gusto/expressions.py
```diff
@@ -0,0 +1,202 @@
+"""
+Some thermodynamic expressions to help declutter the code.
+"""
+from firedrake import exp
```
wence-: Why not call this file `thermodynamics`, and remove all the `_expr` from the function names.

Then in the gusto code, rather than importing all of these into the global namespace, you can do:

```python
from gusto import thermodynamics

...

Pi = thermodynamics.pi(...)
theta = thermodynamics.theta(...)
```
etc...

gusto/expressions.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+__all__ = ["theta_expr", "pi_expr", "pi_rho_expr", "pi_theta_expr", "p_expr", "T_expr", "rho_expr", "r_sat_expr", "Lv_expr", "theta_e_expr", "I_expr"]
+
+
+def theta_expr(parameters, T, p):
+    """
+    Returns an expression for dry potential temperature theta in K.
+
+    :arg parameters: an OutputParameters object.
```
wence-: Are you sure these are `OutputParameters`?  And not `PhysicalParameters` or something like that?

tommbendall: Ah, you're right, `CompressibleParameters` would be better!

gusto/expressions.py
tommbendall: Sure, thank you!

PR #154, opened 2018-01-22
Recovered space scheme

tommbendall:
This pull request implements an advection scheme for the zeroth-degree set of function spaces. It is treated as an option to the `EmbeddedDGAdvection` equation class, as the scheme works via 'recovering' the fields in the first-degree spaces, where the advection happens.

### DIFF
--- a/examples/moist_bf_bubble.py
+++ b/examples/moist_bf_bubble.py
@@ -2,7 +2,8 @@ from gusto import *
 from firedrake import PeriodicIntervalMesh, ExtrudedMesh, \
     SpatialCoordinate, conditional, cos, pi, sqrt, exp, NonlinearVariationalProblem, \
     NonlinearVariationalSolver, TestFunction, dx, TrialFunction, Constant, Function, \
-    LinearVariationalProblem, LinearVariationalSolver, DirichletBC
+    LinearVariationalProblem, LinearVariationalSolver, DirichletBC, \
+    FunctionSpace, BrokenElement, VectorFunctionSpace
 import sys
 
 dt = 1.0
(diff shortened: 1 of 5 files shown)


gusto/advection.py
```diff
@@ -73,17 +80,34 @@ def __init__(self, state, field, equation=None, *, solver_parameters=None,
         # equation class and we can get the correct function space from
         # the output function.
         if isinstance(equation, EmbeddedDGAdvection):
+            # check that the field and the equation are compatible
+            if not equation.V0.__eq__(field.function_space()):
```
wence-: Can you not write `if equation.V0 != field.function_space()`?  If not, we should fix firedrake so you can.

gusto/transport_equation.py
```diff
@@ -212,11 +212,37 @@ class EmbeddedDGAdvection(AdvectionEquation):
               constructed for you.
     :arg solver_params: (optional) dictionary of solver parameters to pass to the
                         linear solver.
+    :arg recovered_spaces: A list or tuple of function spaces to be used for the
+                           recovered space advection method. The must be three
+                           spaces, indexed in the following order:
+                           [0]: the embedded space in which the advection takes place.
+                           [1]: the continuous recovered space.
+                           [2]: a broken or discontinuous version of the original space.
```
wence-: Can you not make `2` automatically from `0`?  `Vbroken = FunctionSpace(V0.ufl_domain(), BrokenElement(V0.ufl_element())` I think?

gusto/advection.py
```diff
@@ -300,3 +324,82 @@ def apply(self, x_in, x_out):
         self.q1.assign(x_in)
         self.solver.solve()
         x_out.assign(self.dq)
+
+
+def recovered_apply(self, x_in):
+    """
+    Extra steps to the apply method for the recovered advection scheme.
```
wence-: I think you could explain what this recovered scheme is for.  Is it a means for obtain a 2nd order accurate advection scheme in the lowest order case?

gusto/transport_equation.py
tommbendall: Yeah I had thought about doing this so that it automatically chooses Vbroken. I wasn't for now though because I wanted to be able to choose different Vbroken so I was putting it in by hand!

gusto/advection.py
tommbendall: Sure, thanks!

gusto/advection.py
tommbendall: Ah yes I can, I'm not sure why I did it in this way!

wence-:
You failed to resolve the merge conflicts when you backmerged master.

examples/moist_bf_bubble.py
```diff
@@ -55,6 +58,37 @@
...
+w_sat1 = params.w_sat1
+w_sat2 = params.w_sat2
+w_sat3 = params.w_sat3
+w_sat4 = params.w_sat4
+T_0 = params.T_0
+g = params.g
+cp = params.cp
+
+=======
+>>>>>>> master
```
wence-: This wants fixed.

gusto/transport_equation.py
```diff
@@ -220,11 +220,41 @@ class EmbeddedDGAdvection(AdvectionEquation):
...
+                           will not be used.
     """
 
+<<<<<<< HEAD
+    def __init__(self, state, V, ibp="once", equation_form="advective", vector_manifold=False, Vdg=None, solver_params=None, recovered_spaces=None):
+=======
     def __init__(self, state, V, ibp="once", equation_form="advective", vector_manifold=False, Vdg=None, solver_params=None, outflow=False):
-
-        if Vdg is None:
+>>>>>>> master
```
wence-: You need to resolve the conflict in the argument list to `__init__`.  Presumably you need to union of the two argument lists now.

tommbendall: Oops, sorry. Hopefully lesson learned!

wence-:
OK!

PR #155, opened 2018-02-06
A hybrid-mixed compressible solver

thomasgibson:
This PR adds a new compressible solver using a hybridized-mixed method.

### DIFF
--- a/examples/compressible_eady.py
+++ b/examples/compressible_eady.py
@@ -15,6 +15,11 @@ else:
     tmax = 30*day
     tdump = 2*hour
 
+if '--hybrid' in sys.argv:
+    hybridization = True
+else:
+    hybridization = False
+
 ##############################################################################
 # set up mesh
 ##############################################################################
@@ -186,11 +191,41 @@ advected_fields.append(("theta", SSPRK3(state, theta0, thetaeqn)))
 # Set up linear solver for the timestepping scheme
 ##############################################################################
 # Set up linear solver
-linear_solver_params = {'ksp_monitor_true_residual': False,
-                        'fieldsplit_1_pc_gamg_sym_graph': True,
-                        'fieldsplit_1_mg_levels_ksp_max_it': 5}
+if hybridization:
+    linear_solver_params = {'ksp_type': 'gmres',
+                            'pc_type': 'gamg',
+                            'pc_gamg_sym_graph': True,
+                            'mg_levels': {'ksp_type': 'chebyshev',
+                                          'ksp_chebyshev_esteig': True,
+                                          'ksp_max_it': 5,
+                                          'pc_type': 'bjacobi',
+                                          'sub_pc_type': 'ilu'}}
+    linear_solver = HybridisedCompressibleSolver(state, solver_parameters=linear_solver_params,
+                                                 overwrite_solver_parameters=True)
 
-linear_solver = CompressibleSolver(state, solver_parameters=linear_solver_params)
+else:
+    linear_solver_params = {
+        'pc_type': 'fieldsplit',
+        'pc_fieldsplit_type': 'schur',
+        'ksp_type': 'gmres',
+        'ksp_max_it': 100,
+        'ksp_gmres_restart': 50,
+        'pc_fieldsplit_schur_fact_type': 'FULL',
+        'pc_fieldsplit_schur_precondition': 'selfp',
+        'fieldsplit_0': {'ksp_type': 'preonly',
+                         'pc_type': 'bjacobi',
+                         'sub_pc_type': 'ilu'},
+        'fieldsplit_1': {'ksp_type': 'preonly',
+                         'pc_type': 'gamg',
+                         'pc_gamg_sym_graph': True,
+                         'mg_levels': {'ksp_type': 'chebyshev',
+                                       'ksp_chebyshev_esteig': True,
+                                       'ksp_max_it': 5,
+                                       'pc_type': 'bjacobi',
+                                       'sub_pc_type': 'ilu'}}
+    }
+    linear_solver = CompressibleSolver(state, solver_parameters=linear_solver_params,
+                                       overwrite_solver_parameters=True)
 
 ##############################################################################
 # Set up forcing
--- a/examples/compressible_eady.py
+++ b/examples/compressible_eady.py
@@ -15,112 +15,8 @@ else:
     tmax = 30*day
     tdump = 2*hour
 
-if '--hybrid' in sys.argv:
-    hybridization = True
-else:
-    hybridization = False
 
-##############################################################################
-# set up mesh
-##############################################################################
-# Construct 1d periodic base mesh
-columns = 30  # number of columns
-L = 1000000.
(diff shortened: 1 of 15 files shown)


wence-:
So you add some more examples, but no tests.  Is that possible too?

examples/compressible_eady.py
```diff
@@ -14,6 +14,11 @@
     tmax = 30*day
     tdump = 2*hour
 
+if '--hybrid' in sys.argv:
+    hybridization = True
+else:
+    hybridization = False
```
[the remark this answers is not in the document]
thomasgibson: Yes, I love Argparse. I use it in my own git repositories, but I know I can't really do it here. I don't think we're ready to completely replace the old compressible solver (that's why they remain separate). Eventually, if we're convinced it's a much better alternative, this solver will replace the existing one.

I'm also not sure how to test this. Should I just duplicate a few of the examples using only the hybridized solver? Or should I just set the `--running-test` argument to run these demos twice (one with and without hybridization)?

gusto/linear_solvers.py
```diff
@@ -229,6 +233,290 @@ def solve(self):
...
+
+    (2d) Project the "broken" velocity field into the HDiv-conforming
+         space using local averaging.
+
+    (3) Reconstruct theta
+
+    :arg state: a :class:`.State` object containing everything else.
+    :arg quadrature degree: tuple (q_h, q_v) where q_h is the required
+    quadrature degree in the horizontal direction and q_v is that in
+    the vertical direction
```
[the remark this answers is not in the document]
thomasgibson: Okay, thanks! I will update.

gusto/linear_solvers.py
```diff
@@ -229,6 +233,290 @@ def solve(self):
...
+        theta = TrialFunction(Vtheta)
+        gamma = TestFunction(Vtheta)
+
+        self.theta = Function(Vtheta)
+        theta_eqn = gamma*(theta - theta_in +
+                           dot(k, self.u_hdiv)*dot(k, grad(thetabar))*beta)*dx
+
+        theta_problem = LinearVariationalProblem(lhs(theta_eqn), rhs(theta_eqn), self.theta)
+        self.theta_solver = LinearVariationalSolver(theta_problem,
+                                                    solver_parameters={'ksp_type': 'gmres',
```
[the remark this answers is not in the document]
thomasgibson: Yes, you're correct. I'll replace `gmres` with `cg`

examples/compressible_eady.py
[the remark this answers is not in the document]
wence-: So, for example, the `test_weightless_tracer` test runs a `CompressibleSolver`.  You could parameterise it to also run a `HybridisedCompressibleSolver`?

thomasgibson:
This PR is on temporary hiatus while I work on this a bit more. Turns out to not be nearly as straightforward as just "hybridizing the linearized system." (This formulation happens to work for most examples, but not really in general)

gusto/linear_solvers.py
```diff
@@ -229,6 +234,321 @@ def solve(self):
...
+            water_t = Function(Vtheta).assign(0.0)
+            for water in self.moisture:
+                water_t += self.state.fields(water)
+            theta = theta / (1 + water_t)
+            thetabar = thetabar / (1 + water_t)
+
+        # "broken" u and rho system
+        Aeqn = (inner(w, (state.h_project(u) - u_in))*dx
+                - beta*cp*div(theta*V(w))*pibar*dxp
+                # + beta*cp*dot(theta*V(w), n)*self.pibar_avg('+')*dS_vp
```
thomasgibson: General question for all: Do we need this term? I originally thought so since the space for `w` is discontinuous, but theta is still continuous in the vertical. We need to be careful about these surface terms when hybridizing.

gusto/linear_solvers.py
```diff
@@ -229,6 +234,327 @@ def solve(self):
...
+        # HDiv-conforming velocity
+        self.u_hdiv = Function(Vu)
+
+        # Reconstruction of theta
+        theta = TrialFunction(Vtheta)
+        gamma = TestFunction(Vtheta)
+
+        self.theta = Function(Vtheta)
+        theta_eqn = gamma*(theta - theta_in +
+                           dot(k, self.u_hdiv)*dot(k, grad(thetabar))*beta)*dx
```
tommbendall: Hmm, I've realised that this isn't right as `thetabar` has been scaled by `1/(1+water_t)` while `theta_in` has not (definitely my bad, so it'll also be wrong in the `CompressibleSolver`)

thomasgibson: So basically we just need something like `theta_in = 1/(1+water_t) * theta_in`?

thomasgibson: I can take the opportunity to correct this in this PR since it seems to be a small update

thomasgibson: Or if you want, you can check out this branch and push your commit to it

thomasgibson:
OK. I think all the quirks are addressed. The gusto test suite will now run both a hybridized and nonhybridized test run for the compressible tests.

gusto/linear_solvers.py
```diff
@@ -229,6 +237,332 @@ def solve(self):
...
+    :arg overwrite_solver_parameters: boolean, if True use only the solver_parameters
+                                      that have been passed in, if False then update.
+                                      the default solver parameters with the
+                                      solver_parameters passed in.
+    :arg moisture (optional): list of names of moisture fields.
+    """
+
+    # Solver parameters for the Lagrange multiplier system
+    # NOTE: The reduced operator is not symmetric
+    solver_parameters = {'ksp_type': 'gmres',
```
thomasgibson: Just a general remark: Can we do better here in terms of an iterative solver? I thought bicg-stab would be a good choice. It works for nonsymmetric systems, seemed stable enough in serial runs. However, running in parallel seems to dramatically increase the number of iterations to reduce the residual by the specified tolerance. I'm talking going from 6 iterations in serial to around 20 for 4 processes (rtol 1e-8).

thomasgibson:
I would like some input on testing the examples: @wence-, @colinjcotter, @jshipton, @tommbendall. I feel that running over a flag for `hybridization` in all the compressible examples is just ugly. Here are a couple proposals:

- Why don't we just select a few interesting test cases (such as the compressible eady, sk nonlinear, dense bubble, and hydrostatic mountain) and create separate hybridized examples. I understand there will be some code-duplication, but I think it's a bit better than what's currently in this PR.

- Alternatively, we could use an argument (`python3 example.py --hybridization`). Then I can just modify `test_examples_run.py` and pass the hybridization flag there. However, I need to find a way to separate out the compressible from the non-compressible examples to avoid duplicated runs.

Let me know what you think.

jshipton:
I like the flag idea. I think the examples could do with a bit of reorganising and having some additional input arguments could really help with avoiding code duplication.

thomasgibson:
Should be good to go now, as everything is passing. I modified `tests/examples/test_examples_run.py` to run the compressible tests with hybridization switched on and everything is passing.

PR #156, opened 2018-02-26
Recovered limiters

tommbendall:
With this pull request, I am adding the facility to apply a slope limiter to the 'recovered' advection scheme used for lowest-degree spaces. The new additions are:

- an updated project step at the end of the 'recovered' advection scheme. This is essentially a bounded projection, preventing new overshoots and undershoots by breaking the projection into interpolation and recovery steps. This type of projection will only be used when a limiter is provided for the advection step.
- a `NoLimiter`option as a limiter, i.e. a limiter which does nothing, which allows us to access the new projection without limiting the advection step.
- two tests for limiters in the 0th-degree and 1st-degree sets of spaces. There are two tests so that limited vertical and horizontal motions are checked. This also replaces the old limiter test.
- a new example, the Bryan and Fritsch dry bubble (which I should have added long ago!)


gusto/limiters.py
```diff
@@ -130,3 +130,21 @@ def apply(self, field):
         self.copy_vertex_values_back(field)
         self.check_midpoint_values(field)
         self.remap_to_embedded_space(field)
+
+
+class NoLimiter(object):
+    """
```
thomasgibson: Wait. Nevermind about the above comment. This is basically like `NoAdvection`.

PR #160, opened 2018-07-04
New example: non-orographic gravity waves on a small planet

thomasgibson:
Gusto should have some 3D examples, so I've added the DCMIP test case 3-1 (see title). This uses the recently added hybridized compressible solver.

### DIFF
--- a/examples/dcmip_3_1_meanflow_quads.py
+++ b/examples/dcmip_3_1_meanflow_quads.py
@@ -1,14 +1,14 @@
 from gusto import *
 from firedrake import CubedSphereMesh, ExtrudedMesh, Expression, \
     FunctionSpace, Function, SpatialCoordinate, as_vector
-from firedrake import exp, acos, cos, sin
+from firedrake import exp, acos, cos, sin, pi
 import sys
 
 
-nlayers = 20           # Number of vertical layers
-refinements = 4        # Number of horiz. cells = 20*(4^refinements)
+nlayers = 10           # Number of vertical layers
+refinements = 3        # Number of horiz. refinements
 
-dt = 8.0               # Time-step size (s)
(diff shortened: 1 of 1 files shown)


jshipton:
Fantastic! Thanks for setting this up! Interested to see how much time it adds to the testing...

thomasgibson:
I tried to set the problem up so that it's not *too* demanding for a laptop. It's very coarse. I also configured the solver options (for the balancing calculation during initialization) and it's a bit faster than the default parameters.

We still need to work on a better solver for the balanced density computation. I think hybridizing vertically in the columns may be a potential solution. This will require some work though, as I don't think we have the right functionality to create a "vertical" trace space.

examples/dcmip_3_1_meanflow_quads.py
```diff
@@ -0,0 +1,190 @@
...
+mesh = ExtrudedMesh(m, layers=nlayers,
+                    layer_height=z_top/nlayers,
+                    extrusion_type="radial")
+
+# Create polar coordinates:
+# Since we use a CG1 field, this is constant on layers
+W_Q1 = FunctionSpace(mesh, "CG", 1)
+z_expr = Expression("sqrt(x[0]*x[0] + x[1]*x[1] + x[2]*x[2]) - a", a=a)
+z = Function(W_Q1).interpolate(z_expr)
+lat_expr = Expression("asin(x[2]/sqrt(x[0]*x[0] + x[1]*x[1] + x[2]*x[2]))")
```
tommbendall: Does this need to be a c expression or can you write it with Firedrake functions?

thomasgibson: No, you are right...thanks for holding me accountable :P I will update.

Issue #158, opened 2018-04-12
Problem with AdvectionDiffusion


tommbendall:
I think that the `AdvectionDiffusion` timestepper doesn't work as intended, as "active" fields (for instance `u`, `rho` or `theta`) don't _appear_ to be advected.

An example of this is in the `theta_limiter` test, visualising the output shows that the field at each time step is identical to the one before.

I think the problem comes in lines 100-108 of `timeloop.py`:
```
for name, advection in self.passive_advection:
    field = getattr(state.fields, name)
    # first computes ubar from state.xn and state.xnp1
    advection.update_ubar(state.xn, state.xnp1, state.timestepping.alpha)
    # advects a field from xn and puts result in xnp1
    advection.apply(field, field)

state.xb.assign(state.xn)
state.xn.assign(state.xnp1)
```

For the `AdvectionDiffusion` timestepper, all the fields are "passive" and are advected in this loop. However any "active" field is also part of `state.xn` and `state.xnp1`. In `AdvectionDiffusion`, `state.xnp1` is never updated, but is then used as the next `state.xn`, so any "active" field is always replaced by what it was initially.

tommbendall:
Will be fixed by the FML branch so closing this issue

PR #161, opened 2018-07-07
Advection diffusion bug

tommbendall:
This pull request is intended to address the problem I raised in issue 158
[](https://github.com/firedrakeproject/gusto/issues/158)

The problem is that the "active fields" `u`, `rho` and `theta` don't get advected by the `AdvectionDiffusion` timestepper.

I'm aware that the timesteppers might change fairly soon but I thought it would be better to get the bug fixed in the meanwhile. My solution isn't particularly elegant though so I'd be very happy to hear better ways to fix this.

I've also added a brief test to highlight the issue.

### DIFF
--- a/gusto/timeloop.py
+++ b/gusto/timeloop.py
@@ -222,16 +222,16 @@ class AdvectionDiffusion(BaseTimestepper):
 
         super(AdvectionDiffusion, self).__init__(state, advected_fields, diffused_fields, physics_list)
 
-        # build a lists to contain the fields that should be advected from state.fieldlist
-        # i.e. these are the active fields to be advected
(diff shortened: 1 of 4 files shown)


jshipton:
Would you mind if we left this as a branch for now? I have been working on changing the time steppers and I think it will fix this issue - I hope to have something to share later this week and I will put in a test like this one to ensure it is working.

tommbendall:
Yes, that's fine! I knew that would be coming along, I just thought that if was going to be a while then it might be nice to have this fixed in the meantime!

jshipton:
I'm being a bit lazy about this making my PR more work!

PR #162, opened 2018-07-24
Richardson number

jshipton:
This PR implements some diagnostic fields: a generic Gradient and a Richardson number.

The Gradient diagnostic can be applied to any field and is computed by projecting the gradient of the field into a CG1 space - the only surface terms are then from the boundaries so this code makes the assumption that if the mesh is extruded then surface terms are required on the top and bottom of the domain. This is not ideal but is in line with other parts of the code.

The Richardson number is the ratio between the buoyancy and the vertical shear: Ri = N^2 / ((u_z)^2 + (v_z)^2) [where _z indicates the derivative w.r.t. z, u and v are the horizontal components of the velocity, and N is the buoyancy frequency]. N can be defined in several equivalent ways depending on which field you use to define the buoyancy: N = sqrt((g/rho_0) * rho_z) = sqrt((g/theta_0) * theta_z) = sqrt(b_z), where _0 denotes a reference value, rho is the density, theta is the potential temperature and b is the buoyancy. To use the Richardson number diagnostic, you must specify which field you are using to define N and you can also specify the 'factor' g/whatever_0, if required. The Richardson number calculation depends on the gradient of velocity and the gradient of density so these diagnostics must also be specified in the setup file - I have added checks for these in the Richardson number calculation. I don't want to get into the thorny issue of automatically figuring out diagnostic field dependencies!!

This code is "tested" in that I have added these diagnostic fields to the sk_nonlinear example - this is not the ideal example for the Richardson number is there is practically no shear so the Richardson number is huge and not really relevant.

Questions:
1.) Should I add an epsilon to the denominator of the Richardson number to ensure we don't pick up Nans? What should that epsilon be?
2.) Should we make a better test? @P-Burns I wonder if you could try this diagnostic out on the layer problem and see if it gives reasonable numbers? Maybe we could find a reduced version of the layer problem to include in the examples folder?

### DIFF
--- a/gusto/diagnostics.py
+++ b/gusto/diagnostics.py
@@ -64,27 +64,18 @@ void maxify(double *a, double *b) {
 
 class DiagnosticField(object, metaclass=ABCMeta):
 
-    def __init__(self):
+    def __init__(self, required_fields=()):
         self._initialised = False
+        self.required_fields = required_fields
 
     @abstractproperty
     def name(self):
         """The name of this diagnostic field"""
         pass
 
-    def check_fields_exist(self, state, fieldlist):
-        field_names = [f.name() for f in state.fields]
-        for field_name in fieldlist:
-            if field_name not in field_names:
-                raise RuntimeError("Field called %s required to compute diagnostic %s" % (field_name, self.name))
-
-    def setup(self, state, space=None, base_field=None, fieldlist=None):
+    def setup(self, state, space=None):
         if not self._initialised:
-            if fieldlist is not None:
-                self.check_fields_exist(state, fieldlist)
-            if base_field:
-                space = state.fields(base_field).function_space()
-            elif space is None:
+            if space is None:
                 space = state.spaces("DG0", state.mesh, "DG", 0)
             self.field = state.fields(self.name, space, pickup=False)
             self._initialised = True
@@ -199,13 +190,13 @@ class RichardsonNumber(DiagnosticField):
     name = "RichardsonNumber"
 
     def __init__(self, density_field, factor=1.):
-        super().__init__()
+        super().__init__(required_fields=(density_field, "u_gradient"))
         self.density_field = density_field
         self.factor = Constant(factor)
 
     def setup(self, state):
         rho_grad = self.density_field+"_gradient"
-        super().setup(state, fieldlist=[rho_grad, "u_gradient"])
+        super().setup(state)
         self.grad_density = state.fields(rho_grad)
         self.gradu = state.fields("u_gradient")
 
@@ -412,7 +403,7 @@ class HydrostaticImbalance(DiagnosticField):
 class Sum(DiagnosticField):
 
     def __init__(self, field1, field2):
-        super(Sum, self).__init__()
+        super(Sum, self).__init__(required_fields=(field1, field2))
         self.field1 = field1
         self.field2 = field2
 
@@ -422,8 +413,8 @@ class Sum(DiagnosticField):
 
     def setup(self, state):
         if not self._initialised:
-            super(Sum, self).setup(state, base_field=self.field1,
-                                   fieldlist=[self.field1, self.field2])
+            space = state.fields(self.field1).function_space()
+            super(Sum, self).setup(state, space=space)
 
     def compute(self, state):
         field1 = state.fields(self.field1)
@@ -434,7 +425,7 @@ class Sum(DiagnosticField):
 class Difference(DiagnosticField):
 
     def __init__(self, field1, field2):
-        super(Difference, self).__init__()
+        super(Difference, self).__init__(required_fields=(field1, field2))
         self.field1 = field1
         self.field2 = field2
 
@@ -444,8 +435,8 @@ class Difference(DiagnosticField):
 
     def setup(self, state):
         if not self._initialised:
-            super(Difference, self).setup(state, base_field=self.field1,
-                                          fieldlist=[self.field1, self.field2])
+            space = state.fields(self.field1).function_space()
+            super(Difference, self).setup(state, space=space)
 
     def compute(self, state):
         field1 = state.fields(self.field1)
@@ -471,9 +462,9 @@ class SteadyStateError(Difference):
 class Perturbation(Difference):
 
     def __init__(self, name):
-        DiagnosticField.__init__(self)
         self.field1 = name
         self.field2 = name+'bar'
+        DiagnosticField.__init__(self, required_fields=(self.field1, self.field2))
 
     @property
     def name(self):
--- a/gusto/diagnostics.py
+++ b/gusto/diagnostics.py
@@ -174,7 +174,9 @@ class Gradient(DiagnosticField):
         trial = TrialFunction(space)
         n = FacetNormal(state.mesh)
         a = inner(test, trial)*dx
-        L = -inner(div(test), f)*dx + dot(dot(test, n), f)*(ds_t + ds_b)
+        L = -inner(div(test), f)*dx
+        if space.extruded:
+            L += dot(dot(test, n), f)*(ds_t + ds_b)
         prob = LinearVariationalProblem(a, L, self.field)
         self.solver = LinearVariationalSolver(prob)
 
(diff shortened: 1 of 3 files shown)


jshipton:
Yikes. OK, I *do* have to worry about figuring out diagnostic field dependencies. What do people think of this solution? [See state.py - I've commented it fairly comprehensively...] I think it will only be robust if we are consistent in using the new `check_fields_exist` method when setting up diagnostic fields that depend on others. The only thing that concerns me is that it might make it harder to debug errors with diagnostic field setup (because you'll just end up getting the `RuntimeError` from line 319 of state.py). I'd appreciate comments from @wence- (but I understand you're pretty busy at this time!)

gusto/diagnostics.py
```diff
@@ -72,12 +72,22 @@ def name(self):
         """The name of this diagnostic field"""
         pass
 
-    def setup(self, state, space=None):
+    def check_fields_exist(self, state, fieldlist):
+        field_names = [f.name() for f in state.fields]
+        for field_name in fieldlist:
+            if field_name not in field_names:
+                raise RuntimeError("Field called %s required to compute diagnostic %s" % (field_name, self.name))
```
wence-: AKA:
```python
fields = set(f.name() for f in state.fields)
missing = set(fieldlist).difference(fields)
if missing:
    raise RuntimeError("Missing fields %s" % missing)
```

gusto/state.py
```diff
@@ -278,15 +278,47 @@ def setup_diagnostics(self):
...
+            elif len(postponed_fields) < len(remaining_diagnostic_fields):
+                # if we've managed to reduce the number of fields we
+                # postpone then continue with another sweep
+                remaining_diagnostic_fields = postponed_fields.copy()
+            else:
+                # if we're not managing to reduce the number of fields
+                # we postpone then something bad is happening
+                raise RuntimeError("There is a problem with your diagnostic fields")
+        # update the list of diagnostic fields to the reordered version
+        self.diagnostic_fields = reordered_diagnostic_fields
```
wence-: You want a topological sort of the field dependency graph, which should be a DAG:

Here, I've implemented [Kahn's algorithm](https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm).
```python
def topo_sort(field_deps):
    # map node: (input_deps, output_deps)
    graph = dict((f, (list(deps), [])) for f, deps in field_deps)
    roots = []
    for f, input_deps in field_deps:
        if len(input_deps) == 0:
            # No dependencies, candidate for evaluation
            roots.append(f)
        for d in input_deps:
            # add f as output dependency
            graph[d][1].append(f)

    schedule = []
    while roots:
        n = roots.pop()
        schedule.append(n)
        output_deps = list(graph[n][1])
        for m in output_deps:
            # Remove edge
            graph[m][0].remove(n)
            graph[n][1].remove(m)
            # If m now as no input deps, candidate for evaluation
            if len(graph[m][0]) == 0:
                roots.append(m)
    if any(len(i) for i, _ in graph.values()):
        cycle = "\n".join("%s -> %s" % (f, i) for f, (i, _) in graph.items()
                          if f not in schedule)
        raise RuntimeError("Field dependencies have a cycle:\n\n%s" % cycle)
    return schedule
```

Run as:

```python
field_deps = [("a", ("b", "c"),),
              ("b", ("c", "d"),),
              ("c", ()),
              ("d", ("c", ))]

schedule = topo_sort(field_deps)
```

If there is a cycle, you get a runtime error:

```python
field_deps = [("a", ("b", "c"),),
              ("b", ("c", "d"),),
              ("c", ()),
              ("d", ("c", "b"))]
topo_sort(field_deps)

=> RuntimeError: Field dependencies have a cycle:

a -> ['b']
b -> ['d']
d -> ['b']
```

gusto/state.py
```diff
@@ -284,6 +284,10 @@ def setup_diagnostics(self):
             f = SteadyStateError(self, name)
             self.diagnostic_fields.append(f)
 
+        fields = set([f.name() for f in self.fields])
+        field_deps = [(d, set(d.required_fields).difference(fields),) for d in self.diagnostic_fields]
```
wence-: We need to care about ordering here.  Because in parallel we need the schedule to come out the same, and the order of the field dependencies determines the order of the schedule.

So perhaps write:

```python
field_deps = [(d, sorted(set(d.required_fields).difference(fields))) for d in self.diagnostic_fields]
```
?

gusto/state.py
```diff
@@ -530,3 +534,34 @@ def get_latlon_mesh(mesh):
...
+        n = roots.pop()
+        schedule.append(n)
+        output_deps = list(graph[n.name][1])
+        for m in output_deps:
+            # Remove edge
+            graph[m.name][0].remove(n.name)
+            graph[n.name][1].remove(m)
+            # If m now as no input deps, candidate for evaluation
+            if len(graph[m.name][0]) == 0:
+                roots.append(m)
```
wence-: Simpler if the scheduler only works with the names (and then you map them back into fields at the end?)

jshipton: possibly... although it is not the field, it is the diagnostic field class. I can get the field name easily from the class (as here) but I can't get the class from the name without defining a dictionary (I don't think) and I couldn't decide what to call the dictionary...

wence-: I thought something like this:

```python
def topo_sort(field_deps):
    name2field = dict((f.name, f) for f, _ in field_deps)
    # map node: (input_deps, output_deps)
    graph = dict((f.name, (list(deps), [])) for f, deps in field_deps)
    roots = []
    for f, input_deps in field_deps:
        if len(input_deps) == 0:
            # No dependencies, candidate for evaluation
            roots.append(f.name)
        for d in input_deps:
            # add f as output dependency
            graph[d][1].append(f.name)

    schedule = []
    while roots:
        n = roots.pop()
        schedule.append(n)
        output_deps = list(graph[n][1])
        for m in output_deps:
            # Remove edge
            graph[m][0].remove(n)
            graph[n][1].remove(m)
            # If m now has no input deps, candidate for evaluation
            if len(graph[m][0]) == 0:
                roots.append(m)
    if any(len(i) for i, _ in graph.values()):
        cycle = "\n".join("%s -> %s" % (f, i) for f, (i, _) in graph.items()
                          if f not in schedule)
        raise RuntimeError("Field dependencies have a cycle:\n\n%s" % cycle)
    return list(map(name2field.__getitem__, schedule))
```
No?

wence-: It's marginal, but you don't have to worry about using the `field` when deleting from output deps, but the `name` when indexing the graph, or deleting from input deps.

jshipton: `name2field` is the kind of name I was looking for! I got hung up on the fact that it isn't the field, but the class used to calculate the field - but we don't care about that when we name, for example, the `diagnostic_fields` list.

jshipton:
@P-Burns any idea whether this gives reasonable Richardson numbers for the layers problem? Do you have a setup that can compare with a Dedaelus run? Thanks!

P-Burns:
So here are a few figures which compare the Richardson number diagnostic in Gusto and Dedalus.  I have plotted the results at t=2 and t=60 s.  The colourmaps are not the same, however, it is clear that the results are quite similar with similar ranges of values and field structures at each time point.  

I don't expect the results to be exactly the same due to differences between the models.  There are small differences in the ICs - something we are in the process of trying to overcome.  There are also differences in the BCs; the basis we use in Dedalus, which was chosen with our mathematical analysis in mind, means that the x-component of the flow field is the same on the top/bottom boundary, which is not the case for our Gusto set-up.  However both models use a slip boundary condition on the top/bottom boundary and both models use periodic BCs in the horizontal direction.  Then we should at least consider differences caused by the different numerical methods (e.g. different numerical diffusion).  But I think the results look good overall.

The plots are showing that initially the flow becomes turbulent in places (with regions of Ri<1/4), which is due to the ICs.  Subsequently the lack of any external forcing in these runs together with diffusion (both numerical and physical) leads to a decay of the turbulence (and the layers that initially develop) and the flow speeds drop.  The drop in flow speeds results in Ri increasing.  I have made runs with an external forcing term in the past and the results are quite different, with turbulence evident (diagnosed by Ri) throughout the run.

ri_2_dedalus
ri_2_gusto
ri_60_dedalus
ri_60_gusto

jshipton:
Thanks for making the comparison @P-Burns 

@colinjcotter @wence- The only outstanding issue now is dealing with potential division by zero which will occur when the vertical gradient of the velocity is zero. @colinjcotter's suggested adding a small parameter, defined by the user, to the denominator but as @P-Burns has managed to plot the Richardson number in Paraview by filtering out the inf values, I wonder if this is necessary - inf is the correct value after all!

PR #164, opened 2018-07-31
Wet hydrostatic balances

tommbendall:
This adds one major feature: a new hydrostatic balance for unsaturated atmospheres (which will be important for rain tests!). Working out how to do this properly also led to rewriting the hydrostatic balance for saturated atmospheres.

The new concept underpinning all of that we want to do our thermodynamics using the density in the temperature space. We decided that the best way to do this was to recover the density into the temperature space using the averaging recovery operator. This idea then needed to follow through to solving the equations to set up hydrostatic balance in wet atmospheres.

Setting up the wet hydrostatic balance involves solving both the hydrostatic equation for `rho` given `theta_v` and `r_v`, but also solving thermodynamic equations for `theta_v` and `r_v` given `rho` and either `theta_e` and `r_t` (in the saturated case) or `theta_d` and `H` (in the unsaturated case). We iteratively solve between these equations, making sure that `rho` is recovered into temperature space for solving the thermodynamic equations. 

Other minor additions:

- some new thermodynamic variables

- new diagnostics for moist variables

- tests for each of the hydrostatic balances

- amendments to the condensation scheme to use `rho` recovered into temperature space

- an example of a rising bubble in an unsaturated wet atmosphere

Sorry for adding lots at once -- I felt they all had to come together!

### DIFF
--- a/gusto/initialisation_tools.py
+++ b/gusto/initialisation_tools.py
@@ -278,7 +278,9 @@ def calculate_Pi0(state, theta0, rho0):
 
 def saturated_hydrostatic_balance(state, theta_e, water_t, pi0=None,
                                   top=False, pi_boundary=Constant(1.0),
-                                  max_iterations=40):
+                                  max_outer_solve_count=40,
+                                  max_theta_solve_count=5,
+                                  max_inner_solve_count=3):
     """
     Given a wet equivalent potential temperature, theta_e, and the total moisture
     content, water_t, compute a hydrostatically balance virtual potential temperature,
@@ -297,7 +299,10 @@ def saturated_hydrostatic_balance(state, theta_e, water_t, pi0=None,
     :arg top: If True, set a boundary condition at the top, otherwise
               it will be at the bottom.
     :arg pi_boundary: The value of pi on the specified boundary.
-    :arg max_iterations: Max number of iterations for balance solver.
+    :arg max_outer_solve_count: Max number of outer iterations for balance solver.
+    :arg max_theta_solve_count: Max number of iterations for theta solver (middle part of solve).
+    :arg max_inner_solve_count: Max number of iterations on the inner most
+                                loop for the water vapour solver.
     """
 
     theta0 = state.fields('theta')
@@ -331,8 +336,7 @@ def saturated_hydrostatic_balance(state, theta_e, water_t, pi0=None,
     r_v_expr = thermodynamics.r_sat(state.parameters, T, p)
     theta_e_expr = thermodynamics.theta_e(state.parameters, T, p, water_v0, water_t)
 
-    iterations = 0
-    for i in range(max_iterations):
+    for i in range(max_outer_solve_count):
         # solve for rho with theta_vd and w_v guesses
         compressible_hydrostatic_balance(state, theta0, rho_h, top=top,
                                          pi_boundary=pi_boundary, water_t=water_t,
@@ -350,14 +354,14 @@ def saturated_hydrostatic_balance(state, theta_e, water_t, pi0=None,
         rho_recoverer.project()
 
         # now solve for r_v
-        for j in range(5):
+        for j in range(max_theta_solve_count):
             theta_h.interpolate(theta_e / theta_e_expr * theta0)
             theta0.assign(theta0 * (1 - delta) + delta * theta_h)
 
             # break when close enough
             if errornorm(theta_e_test, theta_e) < 1e-6:
                 break
-            for k in range(3):
+            for k in range(max_inner_solve_count):
                 w_h.interpolate(r_v_expr)
                 water_v0.assign(water_v0 * (1 - delta) + delta * w_h)
 
@@ -365,9 +369,9 @@ def saturated_hydrostatic_balance(state, theta_e, water_t, pi0=None,
                 theta_e_test.assign(theta_e_expr)
                 if errornorm(theta_e_test, theta_e) < 1e-6:
                     break
-        iterations += 1
-        if iterations == max_iterations:
-            raise RuntimeError('Hydrostatic balance solve has not converged within %i' % iterations, 'iterations')
+
+        if i == max_outer_solve_count:
+            raise RuntimeError('Hydrostatic balance solve has not converged within %i' % i, 'iterations')
 
     if pi0 is not None:
         pie = thermodynamics.pi(state.parameters, rho0, theta0)
@@ -381,7 +385,8 @@ def saturated_hydrostatic_balance(state, theta_e, water_t, pi0=None,
 
 def unsaturated_hydrostatic_balance(state, theta_d, H, pi0=None,
                                     top=False, pi_boundary=Constant(1.0),
-                                    max_iterations=40):
+                                    max_outer_solve_count=40,
+                                    max_inner_solve_count=20):
     """
     Given vertical profiles for dry potential temperature
     and relative humidity compute hydrostatically balanced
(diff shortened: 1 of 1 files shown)


gusto/initialisation_tools.py
```diff
@@ -293,117 +305,170 @@ def moist_hydrostatic_balance(state, theta_e, water_t, pi_boundary=Constant(1.0)
...
+    delta = 0.8
+
+    # expressions for finding theta0 and water_v0 from theta_e and water_t
+    pie = thermodynamics.pi(state.parameters, rho_averaged, theta0)
+    p = thermodynamics.p(state.parameters, pie)
+    T = thermodynamics.T(state.parameters, theta0, pie, water_v0)
+    r_v_expr = thermodynamics.r_sat(state.parameters, T, p)
+    theta_e_expr = thermodynamics.theta_e(state.parameters, T, p, water_v0, water_t)
+
+    for i in range(40):
```
thomasgibson: Magic numbers like this make me nervous. Where does this 40 come from and what does it mean? Why 40? Why not 25 or 10? Scanning through this bit of code, it looks like you have arbitrarily picked a number to apply a sequence of solves hoping to get some field which brings down some error (line 342).

This can potentially be dangerous. Suppose you have an application where this initialization just doesn't work. The code here will be executed over all 40 stages, never breaking early because the condition in line 342 doesn't get satisfied. Junk then potentially gets propagated through and the simulation goes wrong. The user may never know what went wrong.

Here is an alternative. Use a `while` loop and keep track of how many times you're doing this. Something like

```python
counter = 0
while errornorm(theta_e_test, theta_e) > 1.0e-8:

    if counter >= counter_max:
        raise RuntimeError( ... )
    [...do computations...]
    counter += 1
```

That way, when this goes wrong, an error message gets raised and the user will know where to look.

gusto/initialisation_tools.py
```diff
@@ -293,117 +305,170 @@ def moist_hydrostatic_balance(state, theta_e, water_t, pi_boundary=Constant(1.0)
...
+        theta_e_test.assign(theta_e_expr)
+        if errornorm(theta_e_test, theta_e) < 1e-8:
+            break
+
+        # calculate averaged rho
+        rho_broken.interpolate(rho0)
+        rho_recoverer.project()
+
+        # now solve for r_v
+        for j in range(5):
```
thomasgibson: Same problem as mentioned above... I don't know where this 5 is coming from (or the 3 in line 357). I'm not quite sure if I like the idea of having three nested while loops. I need to think about this some more. Maybe others will have some comments.

gusto/initialisation_tools.py
```diff
@@ -293,117 +305,170 @@ def moist_hydrostatic_balance(state, theta_e, water_t, pi_boundary=Constant(1.0)
...
+    r_v_expr = thermodynamics.r_v(state.parameters, H, T, p)
+
+    # make expressions to evaluate residual
+    pi_ev = thermodynamics.pi(state.parameters, rho_averaged, theta0)
+    p_ev = thermodynamics.p(state.parameters, pi_ev)
+    T_ev = thermodynamics.T(state.parameters, theta0, pi_ev, water_v0)
+    RH_ev = thermodynamics.RH(state.parameters, water_v0, T_ev, p_ev)
+    RH = Function(Vt)
+
+    for i in range(40):
```
thomasgibson: My comments and worries about the first instance of this still applies. I recommend making sure all this fails gracefully when it does.

gusto/initialisation_tools.py
tommbendall: Yes you're right, it should raise an error in this case, and a `while` loop is the right way to do it. I'll change that, and thanks for the suggestion!

However, your suggestion doesn't avoid arbitrarily picking the cut off though. So you'd still face the problem you raised about the magic number! It needs to be generous enough so that you get convergence of a difficult problem, but obviously not so high that a divergent problem runs for ever. I think 40 then basically came from trial and error...

thomasgibson: You're correct it doesn't. I can't comment on the science of this, as this is way out of my area, but what is important is gracefully failing.

gusto/initialisation_tools.py
tommbendall: Yes will do the same, thanks! (At least both methods are very similar...)

gusto/initialisation_tools.py
tommbendall: I'm now not sure that a `while` loop is quite right? Sometimes within the solve we expect the error norm to lower than a threshold, (so would cause a `while` loop to end) but that doesn't necessarily mean we have hydrostatic balance. I've pushed some changes that hopefully embody the spirit of your comment though

gusto/initialisation_tools.py
thomasgibson: I still think you should probably also parametrize the inner loop with some argument (set to a default value in the function args). Doing multiple linear solves is expensive. It would be ideal to be able to tweak this from the top level.

gusto/initialisation_tools.py
```diff
@@ -275,15 +276,28 @@ def calculate_Pi0(state, theta0, rho0):
...
+
     :arg state: The :class:`State` object.
     :arg theta_e: The initial wet equivalent potential temperature profile.
     :arg water_t: The total water pseudo-mixing ratio profile.
-    :arg pi_boundary: the value of pi on the lower boundary of the domain.
+    :arg pi0: Optional function to put exner pressure into.
+    :arg top: If True, set a boundary condition at the top, otherwise
+              it will be at the bottom.
+    :arg pi_boundary: The value of pi on the specified boundary.
+    :arg max_iterations: Max number of iterations for balance solver.
```
thomasgibson: I don't think this is an accurate docstring for the `max_iterations` arg. Some may confuse this with iterations of the Krylov method to invert the linear system. Something like:

```python
"""
...
:arg max_outer_solve_count: Maximum number of outer linear solves for the entire balance solver.
:arg max_inner_solve_count: Maximum number of inner linear solves for the entire balance solver.
"""
```
```

File: gusto/initialisation_tools.py (hunk 2 above)
Comment (jshipton) [reply gusto/initialisation_tools.py]: Yes, I agree with @thomasgibson  that the user should be able to set the number of iterations in the inner loop too - fine to have a default value though.

File: gusto/initialisation_tools.py [hunk 5]
```diff
@@ -293,117 +307,181 @@ def moist_hydrostatic_balance(state, theta_e, water_t, pi_boundary=Constant(1.0)
...
+    delta = 0.8
+
+    # expressions for finding theta0 and water_v0 from theta_e and water_t
+    pie = thermodynamics.pi(state.parameters, rho_averaged, theta0)
+    p = thermodynamics.p(state.parameters, pie)
+    T = thermodynamics.T(state.parameters, theta0, pie, water_v0)
+    r_v_expr = thermodynamics.r_sat(state.parameters, T, p)
+    theta_e_expr = thermodynamics.theta_e(state.parameters, T, p, water_v0, water_t)
+
+    iterations = 0
```
jshipton: Do you need the `iterations` variable? Can't you just check the value of `i`after the for loop?

jshipton:
3 more tests?! 😉 Are the moist balance tests more sensitive than the dry? Is that why you're running them for a shorter time and they have lower tolerances? Have you run all of the tests for a longer time?

gusto/initialisation_tools.py
tommbendall: Done, thanks!

gusto/initialisation_tools.py
tommbendall: Hi both, thanks, I've done this now!

tommbendall:
Yeah the wet tests are more sensitive, particularly the saturated one which has the lowest tolerance -- but even this only fails for velocities over 10 micrometers per second which seemed good enough!

I have run the tests for longer, although I was conscious that particularly the wet tests take quite a lot of wall time so I wanted them to be as short as possible -- and from trying a few things out, keeping them short with very harsh tolerances was good enough an indicator of whether the balance procedures were working. The dry balance test is run for longer as the dry solver is used by both the wet solvers -- so more important to be confident in?

examples/moist_bf_bubble.py
```diff
@@ -70,7 +70,7 @@
 Vt = theta0.function_space()
 Vr = rho0.function_space()
 x = SpatialCoordinate(mesh)
-quadrature_degree = (5, 5)
+quadrature_degree = (4, 4)
```
thomasgibson: Out of curiosity, any particular reason for lowering the quadrature degree? Is it a cost issue?

PR #166, opened 2018-09-19
Use ILU for CompressibleHybridizedSolver default

colinjcotter:
Given the lack of understanding we have of AMG with hybridized compressible right now,
and given that ILU works quite well for thin domains, I'm suggesting this as the default for now.


examples/dcmip_3_1_meanflow_quads.py
```diff
@@ -164,19 +164,31 @@
...
+        'ksp_monitor': True,
+        'ksp_rtol': 1.0e-6,
+        'ksp_max_it': 100,
+        'pc_type': 'gamg',
+        'pc_gamg_sym_graph': True,
+        'mg_levels': {'ksp_type': 'gmres',
+                      'ksp_max_its': 8,
+                      'pc_type': 'bjacobi',
+                      'sub_pc_type': 'ilu'}
+    }
```
thomasgibson: Out of curiosity: is the issue that the solver fails to converge for this example, or does it just take too many iterations?

PR #167, opened 2018-10-04
Boundary recovery

tommbendall:
This pull request is designed to improve the recovery process so that it has second-order accuracy on the boundaries.

The main features are:
- the `Recoverer` operator which used to sit in `advection.py` has been taken out and put in a new file, `recovery.py` and renamed to be the `Averager` operator (this is just the `Projector` using the `average` method written by @thomasgibson?)
- there is a new `Boundary_Recoverer` operator which contains the kernels to find the correct values at the boundaries. At the moment it only works for recovering a CG1xCG1 field from a DG0xDG0 field in a vertical slice.
- there is a new `Recoverer` object that composes these operations. It's default action is that of the `Averager`.

Note: while the `Recoverer` can be used in the same way that the `Averager` was, the boundary bits are only currently working for the advection scheme. I also do 'recovery' to convert `rho` into the temperature space in the physics schemes, but this involves recovering CG1xDG0 from DG0xDG0 or CG2xDG1 from DG1xDG1. An implementation of this should be coming soon!

### DIFF
--- a/gusto/recovery.py
+++ b/gusto/recovery.py
@@ -289,7 +289,7 @@ class Recoverer(object):
                 self.project_to_scalar_DG = Projector(self.v_in[0], self.v_in_scalar)
                 self.project_to_scalar_CG = Projector(self.v_out[0], self.v_out_scalar)
                 # need an extra averager that works on the scalar fields rather than the vector one
-                self.extra_averager(self.v_scalar, self.v_out_scalar)
+                self.extra_averager = Averager(self.v_scalar, self.v_out_scalar)
 
     def extract_scalar(self):
         self.project_to_scalar_DG.project()
--- a/gusto/recovery.py
+++ b/gusto/recovery.py
@@ -116,14 +116,15 @@ class Boundary_Recoverer(object):
         # check function spaces of functions -- this only works for a particular set
         if v0.function_space() != VDG0:
             raise NotImplementedError("We can currently only do boundary recovery when v0 is in DG0.")
-        #if v1.function_space() != FunctionSpace(v1.function_space().mesh(), "CG", 1):
-        #    raise NotImplementedError("We can currently only do boundary recovery when v1 is in CG1.")
-        #if v_out.function_space() != FunctionSpace(v_out.function_space().mesh(), "DG", 1):
-        #    raise NotImplementedError("We can currently only do boundary recovery when v_out is in DG1.")
+        if v1.function_space() != FunctionSpace(v1.function_space().mesh(), "CG", 1):
+            raise NotImplementedError("We can currently only do boundary recovery when v1 is in CG1.")
+        if v_out.function_space() != FunctionSpace(v_out.function_space().mesh(), "DG", 1):
+            raise NotImplementedError("We can currently only do boundary recovery when v_out is in DG1.")
 
         VuDG1 = VectorFunctionSpace(VDG0.mesh(), "DG", 1)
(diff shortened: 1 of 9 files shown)


gusto/timeloop.py
```diff
@@ -148,11 +149,12 @@ class CrankNicolson(BaseTimestepper):
...
         :class:`~.Diffusion` to use.
     :arg physics_list: optional list of classes that implement `physics` schemes
-    :arg prescribed_fields: an order list of tuples, pairing a field name with a function that returns the field as a function of time.
+    :arg prescribed_fields: an order list of tuples, pairing a field name with a
+         function that returns the field as a function of time.
     """
 
     def __init__(self, state, advected_fields, linear_solver, forcing,
-                 diffused_fields=None, physics_list=None):
+                 diffused_fields=None, physics_list=None, prescribed_fields=None):
```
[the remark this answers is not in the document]
tommbendall: Accidental omission -- thanks for spotting!

gusto/recovery.py
```diff
@@ -0,0 +1,318 @@
...
+            # now specify things that we'll need if we are doing boundary recovery
+            if boundary_method == 'density':
+                # check dimensions
+                if self.V.value_size != 1:
+                    raise ValueError('This method only works for scalar functions.')
+                self.boundary_recoverer = Boundary_Recoverer(self.v_in, self.v_out, self.v)
+            if boundary_method == 'velocity':
+                # check dimensions
+                if self.V.value_size != 2:
+                    raise ValueError('This method only works for 2D vector functions.')
```
thomasgibson: Are there plans on extending this to 3D? If so, this should also be a `NotImplementedError`. I know this sounds nitpicky, but `ValueError` implies using this `Recoverer` on a 3D problem is not mathematically sensible when in reality I'm sure this approach extends?

gusto/recovery.py
```diff
@@ -0,0 +1,318 @@
...
+        self.averager.project()
+        if self.boundary_method is not None:
+            if self.boundary_method == 'velocity':
+                self.extract_scalar()
+                self.boundary_recoverer.apply()
+                self.extra_averager.project()
+                self.restore_vector()
+            elif self.boundary_method == 'density':
+                self.boundary_recoverer.apply()
+                self.averager.project()
```
thomasgibson: What happens if I set `self.boundary_method = "cinnamon_buns"`? Perhaps there should be an `else` clause which rules out methods that aren't acceptable. In which case, that would be a `ValueError`.

gusto/recovery.py
tommbendall: Yes!

gusto/recovery.py
```diff
@@ -0,0 +1,318 @@
...
+        if v_out.function_space() != FunctionSpace(v_out.function_space().mesh(), "DG", 1):
+            raise NotImplementedError("We can currently only do boundary recovery when v_out is in DG1.")
+
+        VuDG1 = VectorFunctionSpace(VDG0.mesh(), "DG", 1)
+        x, z = SpatialCoordinate(VDG0.mesh())
+        self.coords = Function(VuDG1).project(as_vector([x, z]))
+        self.interpolator = Interpolator(self.v1, self.v_out)
+
+        # check that we're using quads on extruded mesh -- otherwise it will fail!
+        if not VDG0.extruded:
```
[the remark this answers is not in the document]
thomasgibson: I think you will need `from firedrake.ufl import quadrilateral`

PR #170, opened 2018-10-11
define equation and term classes and make it all work for a basic sha…

jshipton:
…llow water equation test

This PR is for discussion...

The eventual aim is to make it easier to do different timestepping schemes by tagging terms (e.g. "explicit"/"implicit" for IMEX schemes or "linear"/"nonlinear" for exponential integrators). This code makes some progress towards that by defining an `Equation` class that takes a function space arg and defines a test function and a mass term, and has a method for adding further terms. The terms are defined in terms.py or transport_terms.py (previously known as transport_equation.py). Equations can then be bundled together using the imaginatively named `Equations` class. The relevant subclass (e.g. `ShallowWaterEquations`) can be passed to the `Forcing` class to generate the forcing solver. At the moment this just reproduces the current split scheme so I haven't labelled the terms, just separated them based on whether they are subclasses of the `TransportTerm` (meaning they're treated as "advection") or not (meaning they are "forcing").

Note:
- The `Forcing` class is now generic.
- The allocation of various bits of state has been moved to the places that need them - so the linear solver class creates the functions it needs, the timestepping class creates the functions it needs (e.g. the xstar and xp functions for the intermediate steps).
- The function spaces are built by the `Equations` class but still stored in state. I think this makes sense and helped avoid circular dependencies.
- Each term has an `off_centering` property that is currently hardcoded to 0.5 but can be exposed to the user to set.
- I have removed some functionality - this will all go back once I figure out how!

To-do list (looks short but isn't):
- Fix all the various advection options (SUPG, EmbeddedDG etc.)
- Implement the terms and equations classes for the other sets of equations that we solve.
- Fix tests and examples.

Before we continue, I'd love some comments on how this could be improved... thanks!


thomasgibson:
I think you need to commit `terms.py`, as I cannot see it anywhere in the branch.

gusto/equations.py
```diff
@@ -0,0 +1,168 @@
...
+        self.test = TestFunction(function_space)
+        self.trial = TrialFunction(function_space)
+        self.bcs = None
+
+    def mass_term(self, q):
+        return inner(self.test, q)*dx
+
+    def add_term(self, term):
+        key = term.__class__.__name__
+        self.terms[key] = term
```
dham: This is a classic case where you should overload intrinsic operators to minimise notational clutter. Instead of calling this method `add_term`, you should call it `__iadd__`. This will cause the notation `equation += term` to work.

gusto/terms.py
```diff
@@ -0,0 +1,43 @@
...
+    def __call__(self, q, fields):
+        return self.evaluate(q, fields)
+
+
+class ShallowWaterPressureGradientTerm(Term):
+
+    def evaluate(self, q, fields):
+        g = self.parameters.g
+        D = fields("D")
+        return g*div(self.test)*D*dx
```
dham: There's something funny in the design here. Somehow the `Term` knows which coefficients to extract from the state but doesn't know which test function. This causes a kind of broken concept of who owns what information.

jshipton: Yes, I think the parameters should be part of the equation class, rather than state.

gusto/equations.py
```diff
@@ -0,0 +1,168 @@
...
+    def __init__(self, state, family, degree):
+
+        self.state = state
+        self._build_function_spaces(state.spaces, state.mesh, family, degree)
+        state.fields = FieldCreator(self)
+
+        if hasattr(state, "diagnostics"):
+            state.diagnostics.register(*self.fieldlist)
+        else:
+            state.diagnostics = Diagnostics(*self.fieldlist)
```
wence-: This ugliness could be avoided if `state` always made an (initially empty) `Diagnostics` object.

gusto/equations.py
wence-: Don't do that latter thing, since it swiftly becomes a nightmare.  Is it `(term, "label")` or `("label", term)`. What if you want to add some further thing? Why not instead just label the terms themselves, e.g., `foo = SomeTerm(); foo = foo.label("implicit")`

Where `foo.label` takes a shallow copy and adds a label (so that you don't have worries about sharing data).

gusto/state.py
```diff
@@ -29,14 +27,16 @@ def __call__(self, name, mesh=None, family=None, degree=None):
...
         self.fields = []
-        if fieldlist is not None:
-            for name, func in zip(fieldlist, xn.split()):
-                setattr(self, name, func)
-                func.dump = name in dumplist
-                func.pickup = pickup
-                func.rename(name)
+        if equations is not None:
+            xn = Function(equations.mixed_function_space)
+            setattr(self, 'xfields', xn)
```
wence-: Here you can write `self.xfields = xn`

gusto/terms.py
```diff
@@ -0,0 +1,43 @@
+from abc import ABCMeta, abstractmethod
+from firedrake import FacetNormal, div, dx, inner
+
+
+class Term(object, metaclass=ABCMeta):
+
+    off_centering = 0.5
```
wence-: This seems odd, it's somehow prescribing an implicitness.

gusto/terms.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def __init__(self, state, test):
+        self.state = state
+        self.test = test
+        self.parameters = state.parameters
+        self.n = FacetNormal(state.mesh)
```
wence-: Why does this guy need to carry around a symbolic facet normal? Why can't you make it when you need it?

gusto/equations.py
dham: No, you can't add terms to magic methods. However I also think you shouldn't. The Right Thing is to build mathematical objects and apply mathematical operators to them. Terms are objects. If you want a label on the term then there should be an operator which takes a term and returns a term with a label. The labelled term should then be added to the equation. This way we keep composability and echo the maths more closely. 

I think Gusto is in a little danger of digging itself into a hole here. Can we get everyone relevant around a whiteboard and work out what the maths of this (and therefore the code) should look like?

gusto/terms.py
dham: Sorry, I misspoke. I mean which function space it's taking pressure gradient against. Same issue.

gusto/transport_terms.py
```diff
@@ -0,0 +1,205 @@
...
+    The equation is assumed to be in the form:
+
+    q_t + L(q) = 0
+
+    where q is the (scalar or vector) field to be solved for.
+
+    :arg state: :class:`.State` object.
+    :arg test: :class:`.TestFunction` object.
+    :arg ibp: string, stands for 'integrate by parts' and can take the value
+              None, "once" or "twice". Defaults to "once".
```
wence-: Prefer an `enum`:

```python
from enum import Enum

class TransportTerm(Term):
   class IntegrateByParts(Enum):
      NO
      ONCE
      TWICE
   def __init__(self, ..., ibp):
      ...

...

transport = TransportTerm(..., TransportTerm.IntegrateByParts.ONCE)
```

wence-: This way, you get error checking of values and extensibility somehow "for free".

gusto/transport_terms.py
```diff
@@ -0,0 +1,205 @@
...
+        super().__init__(state=state, test=test, ibp=ibp)
+        if equation_form == "advective" or equation_form == "continuity":
+            self.continuity = (equation_form == "continuity")
+        else:
+            raise ValueError("equation_form must be either 'advective' or 'continuity', not %s" % equation_form)
+
+        self.qbar = qbar
+
+        # currently only used with the following option combinations:
+        if self.continuity and ibp is not "once":
```
wence-: Careful, you are lucky that python interns short strings. There is no guaranteed that `str is str.copy()`, use rather `str == str.copy()` or `str != str.copy()` as appropriate.

dham:
I've done a bunch of thinking about this problem, the outcome of which is: https://github.com/firedrakeproject/gusto/wiki/Programmable-equation-transformations . Thoughts welcome.

gusto/transport_terms.py
wence-: BTW, everyone points out to me that the zero-equivalent of ONCE is NEVER.

wence-:
> I've done a bunch of thinking about this problem, the outcome of which is: https://github.com/firedrakeproject/gusto/wiki/Programmable-equation-transformations . Thoughts welcome.

We discussed this proposal in the Firedrake meeting on Tuesday, and came to the conclusion that it is broadly right. The only slightly difficult bit of coding is in doing the lazily evaluated label filter. @jshipton/@dham, how should we proceed with this aspect? Do we merge this current refactoring (or something similar), and then update with the newer proposal. Or do we work out the proposed "term-algebra" stuff and update this to use that?

wence-:
> One thing I don't understand is how to implement the magic methods for logical expressions (like in @dham's example) - I can do `__add__` but how do I make something like `if label_a and not label_b` work?

This is the slightly tricky bit. You get to learn how to implement your own mini symbolic language.

jshipton:
where does one learn that?!

dham:
We can do it tomorrow, Jemma. It's not going to take long.

wence-:
> > One thing I don't understand is how to implement the magic methods for logical expressions (like in @dham's example) - I can do `__add__` but how do I make something like `if label_a and not label_b` work?
> 
> This is the slightly tricky bit. You get to learn how to implement your own mini symbolic language.

I just edited the wiki page. You can't actually override the logical operators, only the bitwise ones. So do we want:

```python
(label_a & ~label_b) | label_c
```

or
```python
Or(And(label_a, Not(label_b)), label_c)
```

One annoyance is that I think the bitwise operators have different precedence rules to the logical ones.
In either case, the internal representation will look like the latter, but the question is how we construct it.

dham:
Hmm. That sucks. 
OK what about the following alternative. We dump the filter and just have label map take a function which takes a `Term` and returns a `Term` or `Equation`. 
E.g. 
```
eq.label_map(lambda term: new_label(term) if term.has(old_label) else term) 
```
That way we hand the delayed evaluation problem back to Python.

wence-:
Totally untested:
```
from functools import singledispatch
from gem.node import Memoizer


@singledispatch
def _evaluate(expr, self):
    raise AssertionError("Unhandle")


@_evaluate.register(bool)
def _evaluate_bool(expr, self):
    return expr


@_evaluate.register(Label)
def _evaluate_term(label, self):
    return label(self.term)


@_evaluate.register(Not)
def _evaluate_not(expr, self):
    child, = (self(c) for c in expr.children)
    return not child


@_evaluate.register(And)
def _evaluate_and(expr, self):
    return all(self(c) for c in expr.children)


@_evaluate.register(Or)
def _evaluate_or(expr, self):
    return any(self(c) for c in expr.children)


def evaluate(expr, term):
    mapper = Memoizer(_evaluate)
    mapper.term = term
    return mapper(expr)


class Expr(object):
    def __init__(self, *operands):
        self.children = tuple(operands)

    def __call__(self, term):
        return evaluate(self, term)


class And(object):
    def __new__(cls, a, b):
        if type(a) is bool:
            if not a:
                return False
            return b
        if type(b) is bool:
            if not b:
                return False
            return a
        super().__new__(cls)


class Or(object):
    def __new__(cls, a, b):
        if type(a) is bool:
            if a:
                return True
            return b
        if type(b) is bool:
            if b:
                return True
            return a
        super().__new__(cls)


class Eq(object):
    def __new__(cls, a, b):
        if type(a) != type(b):
            return False
        super().__new__(cls)


class Not(object):
    def __new__(cls, expr):
        if type(expr) is bool:
            return not expr
        return super().__new__(cls)

    def __init__(self, expr):
        self.children = (expr, )
```

dham:
Yeah, I get how you implement it. The question is whether it's more ugly to require functions or to have a non-pythonic interface.

wence-:
> Hmm. That sucks.
> OK what about the following alternative. We dump the filter and just have label map take a function which takes a `Term` and returns a `Term` or `Equation`.
> E.g.
> 
> ```
> eq.label_map(lambda term: new_label(term) if term.has(old_label) else term) 
> ```
> That way we hand the delayed evaluation problem back to Python.

I prefer, I think, using a lambda, but only as a filter. So:

```python
eq.label_map(lambda term: term.has(some_label), new_label)
```

With the internal logic handling the creation of passing through of terms as appropriate:

```python
def label_map(self, filter, fn):
   return type(self)(*(fn(t) if filter(t) else t for t in self.terms))
```

wence-:
> Yeah, I get how you implement it. The question is whether it's more ugly to require functions or to have a non-pythonic interface.

That was for @jshipton's edification.

PR #171, opened 2018-10-12
Make it rain!

tommbendall:
A storm is brewing in Gusto...

This pull requests adds some parametrizations to Gusto relating to rain: specifically involving the formation of rain from cloud water, the evaporation of rain water and a more sophisticated model of sedimentation.

And if your will to live hasn't already d-_rained_ away, it also updates `unsaturated_bubble.py` to be a warm rain test.


gusto/physics.py
```diff
@@ -84,62 +97,285 @@ def __init__(self, state):
...
+            self.water_v.assign(self.water_v_new.interpolate())
+            self.water_c.assign(self.water_c_new.interpolate())
 
 
 class Fallout(Physics):
     """
     The fallout process of hydrometeors.
 
     :arg state :class: `.State.` object.
+    :arg moments: an integer denoting which rainfall scheme to use.
```
jshipton: I was trying to think if there are more intuitive values for this argument but it's so complex that I'm not sure there is an obvious answer

wence-: Use an Enum. 

AdvectedMoments.MEAN

And similar

thomasgibson: https://docs.python.org/3/library/enum.html

tommbendall: I don't think I understand enum well enough to see how to use it here!

wence-: It just removes the need for magic numbers in the code.

```python
class AdvectedMoments(Enum):
   M0 = 0
   M3 = 1
   # any other moments here
```

```python
Fallout(..., moments=AdvectedMoments.M0, ...)
```

Now the value describes what is going on.

tommbendall: I've tried to make this change now, possibly I still haven't exactly understood how to do it properly

jshipton:
Looks great!

PR #172, opened 2018-10-14
A vertically-oriented hybridization preconditioner for Gusto initializations

thomasgibson:
This add a new PC type specifically for Gusto. This is based on the `HybridizationPC` in Firedrake, but modified for solving the hydrostatic pressure equation in Gusto's initialization routines. The hope is that this will be much faster in the long run than the current setup.

I have done some tests (running various Gusto examples) and it seems to work. I would appreciate it if some of you could also stress-test this by trying it out for whatever compressible/incompressible problems you have.


colinjcotter:
I think this looks good. Please can we have a standalone test that just does a hydrostatic compressible solve?

gusto/preconditioners.py
```diff
@@ -0,0 +1,388 @@
...
+        Vp = V[self.pidx]
+
+        # Create the space of approximate traces in the vertical.
+        # NOTE: Technically a hack since the resulting space is technically
+        # defined in cell interiors, however the degrees of freedom will only
+        # be geometrically defined on edges. Arguments will only be used in
+        # surface integrals
+        deg, _ = Vv.ufl_element().degree()
+
+        # Assumes a tensor product cell (quads, cubes)
```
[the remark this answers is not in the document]
thomasgibson: You are correct. I will update this

thomasgibson:
Updated and ready for round-two of reviews. Here is an updated summary:

- @colinjcotter and I need to work out what needs to be done for the incompressible case. Currently, this preconditioner will not produce equivalent results. This will be updated in a future PR.
- For the compressible case, this works fine and is tested in `gusto/tests/test_vertical_hybridization.py`.
- It only works for tensor produce cells. The whole "vertical trace" thing is an exploitation of the tensor product structure. This is fine since everything Gusto works with has this tensor product formulation in the function spaces.

gusto/__init__.py
```diff
@@ -1,14 +1,15 @@
...
-from gusto.timeloop import *        # noqa
-from gusto.transport_equation import *     # noqa
+from gusto.limiters import *              # noqa
+from gusto.linear_solvers import *        # noqa
+from gusto.physics import *               # noqa
+from gusto.preconditioners import *       # noqa
+from gusto.recovery import *              # noqa
+from gusto.state import *                 # noqa
+from gusto.timeloop import *              # noqa
+from gusto.transport_equation import *    # noqa
```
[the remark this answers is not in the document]
jshipton: love this attention to detail!

gusto/preconditioners.py
```diff
@@ -0,0 +1,392 @@
...
+__all__ = ["VerticalHybridizationPC"]
+
+
+class VerticalHybridizationPC(PCBase):
+    """ A Slate-based python preconditioner for solving
+    the hydrostatic pressure equation (after rewriting as
+    a mixed vertical HDiv x L2 system). This preconditioner
+    transforms the original mixed system by hybridizing
+    the equations in the vertical direction. Lagrange multipliers
+    are introduced on the top/bottom faces of each element are
```
jshipton: this sentence doesn't quite make sense

gusto/preconditioners.py
```diff
@@ -0,0 +1,392 @@
...
+    condensed out via static condensation.
+
+    This PC assembled a condensed problem for the multipliers
+    and inverted using the provided solver options. The orginal
```
jshipton: "assembles", "inverts" and "original"

gusto/preconditioners.py
```diff
@@ -0,0 +1,392 @@
...
+
+        # Set up relevant functions
+        self.broken_solution = Function(V_d)
+        self.broken_residual = Function(V_d)
+        self.trace_solution = Function(Vv_tr)
+        self.unbroken_solution = Function(V)
+        self.unbroken_residual = Function(V)
+
+        # Set up transfer kernels to and from the broken velocity space
+        # NOTE: Since this snippet of code is used in a couple places in
```
jshipton: yes! could you create an issue so that we don't forget this?

PR #174, opened 2018-10-30
Fix lint

tommbendall:
Quick pull request that hopefully fixes the lint following the flake8 update, so that branches will pass on jenkins again.

This also neatens up some of the headers of scripts when importing functions, etc.


jshipton:
This looks great - thanks for doing that! I'll approve once Jenkins has got its act together...

PR #178, opened 2018-11-06
Tidy embedded dg decorator

jshipton:
This PR tidies the embedded_dg decorator used for the explicit advection schemes by defining a `pre_apply` and `post_apply` method that do what is needed depending on whether we're doing standard embedded DG or using a recovered advection scheme.

The options for embedded DG and and the recovered schemes are now defined in their own `Configuration` classes. @tommbendall I couldn't see where the `boundary_method` option gets pulled from the equation so maybe I haven't set that option correctly...

This will make the new equation labelling a bit easier as I envisage that these option classes can be passed directly to the advection scheme in future.


jshipton:
@tommbendall I also combined your 2 limiters tests as somehow that felt like less work than changing the options in one of them... I hope they both do what you want them to...

gusto/transport_equation.py
```diff
@@ -234,36 +234,19 @@ class EmbeddedDGAdvection(AdvectionEquation):
...
-                    raise ValueError('recovered_spaces must be a list or tuple containing three spaces')
-                self.space = recovered_spaces[0]  # the space in which advection happens
-                self.V_rec = recovered_spaces[1]  # the recovered continuous space
-                self.V_brok = recovered_spaces[2]  # broken version of V0
-                self.recovered = True
-        elif boundary_method is not None:
-            raise ValueError('A boundary method can only be used with the recovered space method')
-        elif Vdg is None:
-            # Create broken space, functions and projector
+    def __init__(self, state, V, ibp="once", equation_form="advective",
```
tommbendall: I think we can now remove some of the documentation from this section

gusto/advection.py
```diff
@@ -74,46 +68,91 @@ def __init__(self, state, field, equation=None, *, solver_parameters=None,
...
+            self.x_rec_projector.project()
+            self.x_brok_projector.project()
+            self.xdg_interpolator.interpolate()
+
+    def post_apply(self, x_out, discretisation_option):
+        """
+        The projection steps for the recovered advection scheme,
+        used for the lowest-degree sets of spaces. This returns the
+        field to its original space, from the space the embedded DG
+        advection happens in. This step acts as a limiter.
```
tommbendall: I realise that I wrote this, but technically this step is only limited if `self.limiter is not None`

gusto/advection.py
```diff
@@ -74,46 +68,91 @@ def __init__(self, state, field, equation=None, *, solver_parameters=None,
...
+                self.discretisation_option = None
+                self.fs = field.function_space()
+
+            # setup required functions
+            self.dq = Function(self.fs)
+            self.q1 = Function(self.fs)
+
+    def _setup(self, state, field, options):
+
+        if options.name in ["embedded_dg", "recovered"]:
```
tommbendall: Is it worth raising an error at some point if the options aren't one of these?

gusto/configuration.py
```diff
@@ -133,3 +134,25 @@ class CompressibleEadyParameters(CompressibleParameters, EadyParameters):
...
     dthetady = theta_surf/g*EadyParameters.dbdy
     Pi0 = 0.0
+
+
+class AdvectionOptions(Configuration, metaclass=ABCMeta):
+
+    @abstractproperty
+    def name(self):
+        pass
+
```
tommbendall: This is definitely a smarter and neater way of doing these options! Thanks so much for tidying up the mess I left behind...

gusto/advection.py
```diff
@@ -91,37 +91,31 @@ def _setup(self, state, field, options):
                           'sub_pc_type': 'ilu'}
             self.Projector = Projector(self.xdg_out, self.x_projected,
                                        solver_parameters=parameters)
+        elif options.name is not None:
```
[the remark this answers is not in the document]
tommbendall: Oh I see -- seems like we shouldn't really need a check in that case then!

PR #182, opened 2018-11-28
Fix netCDF crash

dacreman:
Fix for crash when writing diagnostics.nc with >=16 MPI processes. The fix ensures that only the rank 0 process in the mesh communicator access the file, whilst all processes in the communicator participate in calculating the diagnostics.

### DIFF
--- a/gusto/state.py
+++ b/gusto/state.py
@@ -113,7 +113,7 @@ class PointDataOutput(object):
 
 
 class DiagnosticsOutput(object):
-    def __init__(self, filename, diagnostics, description, create=True):
+    def __init__(self, filename, diagnostics, description, rank, create=True):
         """Create a dump file that stores diagnostics.
 
         :arg filename: The filename.
@@ -123,19 +123,21 @@ class DiagnosticsOutput(object):
         """
         self.filename = filename
         self.diagnostics = diagnostics
+        self.rank = rank
         if not create:
             return
-        with Dataset(filename, "w") as dataset:
-            dataset.description = "Diagnostics data for simulation {desc}".format(desc=description)
-            dataset.history = "Created {t}".format(t=time.ctime())
-            dataset.source = "Output from Gusto model"
-            dataset.createDimension("time", None)
-            var = dataset.createVariable("time", np.float64, ("time", ))
-            var.units = "seconds"
-            for name in diagnostics.fields:
-                group = dataset.createGroup(name)
-                for diagnostic in diagnostics.available_diagnostics:
-                    group.createVariable(diagnostic, np.float64, ("time", ))
+        if self.rank == 0:
+            with Dataset(filename, "w") as dataset:
+                dataset.description = "Diagnostics data for simulation {desc}".format(desc=description)
+                dataset.history = "Created {t}".format(t=time.ctime())
+                dataset.source = "Output from Gusto model"
+                dataset.createDimension("time", None)
+                var = dataset.createVariable("time", np.float64, ("time", ))
+                var.units = "seconds"
+                for name in diagnostics.fields:
+                    group = dataset.createGroup(name)
+                    for diagnostic in diagnostics.available_diagnostics:
+                        group.createVariable(diagnostic, np.float64, ("time", ))
 
     def dump(self, state, t):
         """Dump diagnostics.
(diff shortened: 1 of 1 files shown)


wence-:
My concern here is that this is somehow hiding a deeper issue. I had thought that netCDF was a parallel file format. But it looks like one needs to do some extra things. Reading from the netcdf4 documentation (http://unidata.github.io/netcdf4-python/#section13).

Do things work if you follow their instructions and say:

```python
with DataSet(..., parallel=True, comm=mesh.comm) as f:
  ...
```

You probably then need to indicate that things are collective output too (also in the docs).

gusto/state.py
```diff
@@ -123,36 +123,47 @@ def __init__(self, filename, diagnostics, description, create=True):
...
-            dataset.history = "Created {t}".format(t=time.ctime())
-            dataset.source = "Output from Gusto model"
-            dataset.createDimension("time", None)
-            var = dataset.createVariable("time", np.float64, ("time", ))
-            var.units = "seconds"
-            for name in diagnostics.fields:
-                group = dataset.createGroup(name)
-                for diagnostic in diagnostics.available_diagnostics:
-                    group.createVariable(diagnostic, np.float64, ("time", ))
+        if self.rank == 0:
```
wence-: Prefer passing in a communicator, rather than just the rank.

gusto/state.py
```diff
@@ -123,36 +123,47 @@ def __init__(self, filename, diagnostics, description, create=True):
...
+                        var[idx:idx + 1] = diagnostic(field)
+        else:
             for name in self.diagnostics.fields:
                 field = state.fields(name)
-                group = dataset.groups[name]
                 for dname in self.diagnostics.available_diagnostics:
                     diagnostic = getattr(self.diagnostics, dname)
-                    var = group.variables[dname]
-                    var[idx:idx + 1] = diagnostic(field)
+                    diagnostic(field)
```
wence-: For clarity, I would prefer having this calculation of diagnostics happen up front, and then dump on rank 0. So like this perhaps:

```python
diagnostics = []
    field = state.fields(name)
    for dname in self.diagnostics.available_diagnostics:
        diagnostic = getattr(self.diagnostics, dname)
        diagnostics.append((name, dname, diagnostic(field))

if self.comm.rank == 0:
    with DataSet(...) as dataset:
        idx = dataset.dimensions["time"].size
        dataset.variables["time"][idx:idx+1] = t
        for field_name, name, diagnostic in diagnostics:
            group = dataset.groups[field_name]
            var = group.variables[name]
            var[idx:idx+1] = diagnostic
```

(untested)

dacreman:
If I understand correctly the diagnostics calculated in diagnostics.py are global values valid over the whole domain and Firedrake carries out MPI communication transparently when fields are accessed. I've checked that the values of the diagnostics are identical across all the MPI processes so they do appear to be global values. If this is so then we just need a single MPI process to write the global value to the netCDF file. This is a different case to writing out a domain decomposed field, where you could have different MPI processes writing their own sub-domain to a different part of the file instead of funnelling I/O through a single process. 

It appears that the netCDF library available to Gusto is not parallel enabled (opening the file with parallel access fails). However the fields are being written out in VTK format (one file per processor) so I've not seen any issues writing fields.

Even if we did have a parallel netCDF library I don't think we want all the MPI processes writing the diagnostics as they would all be accessing the same location in the file (unless we wanted diagnostics per subdomain in which case we would write a different value for each MPI process).

jshipton:
@dacreman that makes sense to me but then I'm not sure how a parallel enabled NetCDF library would work - if that would be better then maybe we should investigate using that instead but from your description (which is what I understand is happening) it looks like what you've done here is what we need anyway. Would you agree @wence- ?

dacreman:
Seeing the same problem with point_data.nc. Should be a similar fix. 

> Traceback (most recent call last):
  File "boussinesq_2d_lab_scaling.py", line 330, in <module>
    stepper.run(t=0, tmax=tmax)
  File "/gpfs/ts0/projects/Research_Project-189873/dma204/myBranch/firedrake/src/gusto/gusto/timeloop.py", line 96, in run
    t = self.setup_timeloop(state, t, tmax, pickup)
  File "/gpfs/ts0/projects/Research_Project-189873/dma204/myBranch/firedrake/src/gusto/gusto/timeloop.py", line 77, in setup_timeloop
    state.setup_dump(t, tmax, pickup)
  File "/gpfs/ts0/projects/Research_Project-189873/dma204/myBranch/firedrake/src/gusto/gusto/state.py", line 373, in setup_dump
    create=not pickup)
  File "/gpfs/ts0/projects/Research_Project-189873/dma204/myBranch/firedrake/src/gusto/gusto/state.py", line 74, in __init__
    with Dataset(filename, "w") as dataset:
  File "netCDF4/_netCDF4.pyx", line 2135, in netCDF4._netCDF4.Dataset.__init__
  File "netCDF4/_netCDF4.pyx", line 1752, in netCDF4._netCDF4._ensure_nc_success
PermissionError: [Errno 13] Permission denied: b'results/tmp/point_data.nc'

dacreman:
There is a failure in the Jenkins tests running h_mountain.py but I see the same failure on the parent flexible_dumping branch so I don't think my changes are responsible (and the merge test passes). Consequently I think it is OK to merge my branch back into flexible_dumping.

PR #183, opened 2018-12-18
Neaten diagnostics

tommbendall:
A quick pull requests that neatens up the diagnostics of the thermodynamic variables (and introduces a couple of new ones, in particular energy).

The way it does this is by introducing a parent class `ThermodynamicDiagnostic`, which has the prognostic variables as properties. The particular diagnostics then inherit these properties.

Hopefully this is a sensible way of doing this!

### DIFF
--- a/gusto/diagnostics.py
+++ b/gusto/diagnostics.py
@@ -11,9 +11,10 @@ import numpy as np
 
 __all__ = ["Diagnostics", "CourantNumber", "VelocityX", "VelocityZ", "VelocityY", "Gradient",
            "RichardsonNumber", "Energy", "KineticEnergy", "ShallowWaterKineticEnergy",
-           "ShallowWaterPotentialEnergy", "ShallowWaterPotentialEnstrophy", "ExnerPi", "Sum",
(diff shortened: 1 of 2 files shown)


gusto/diagnostics.py
```diff
@@ -326,145 +322,209 @@ def compute(self, state):
...
+    @property
+    def name(self):
+        return self.field1+"_plus_"+self.field2
 
     def setup(self, state):
         if not self._initialised:
-            space = state.spaces("CG1", state.mesh, "CG", 1)
-            super(Theta_e, self).setup(state, space=space)
+            space = state.fields(self.field1).function_space()
+            super(Sum, self).setup(state, space=space)
```
jshipton: you can just use `super().setup...`

jshipton:
Looks good to me. Is there a test or example file that you could add them to just so we can see that they work?

gusto/diagnostics.py
tommbendall: Oh yes you're right. I didn't actually change this bit of the code so I'm not sure why git is showing it as a change!

Issue #190, opened 2019-01-22
mkdir having problems in parallel


tommbendall:
We think the new mkdir is causing problems in parallel, @thomasgibson has done a brief fix with this commit: d65cb822c1e1778ec09e73539f314f6bcb846ede on `solver-refactor` branch.

I'm just making a note of this so I remember - I'll file a pull request for this tomorrow morning!

jshipton:
Yikes! We should have a parallel test...

tommbendall:
In `Jenkinsfile` the command to execute the tests is `python -m pytest -n 4 -v tests`. I'd have thought this was ensuring the tests in parallel but apparently not...?

thomasgibson:
Is it missing `mpiexec -n 4...`?

thomasgibson:
This has already been fixed I believe (in #191)

Issue #169, opened 2018-10-11
Hybrid solver failing on unsaturated bubble test


tommbendall:
I've noticed that the hybridized compressible solver fails on my new unsaturated bubble test (the one featuring rain), whilst the old compressible solver seems to work.

With the hybrid solver it looks like gravity waves end up propagating and amplifying themselves, until at some point the CFL limit is breached and the model can no longer converge.

thomasgibson:
Can you remind me: does the hybrid solver work using the "old style" advection schemes?

tommbendall:
I am satisfied that this has been solved by PRs 191 and 194

PR #191, opened 2019-01-23
Use Firedrake's static condensation preconditioner in the hybridized solver

thomasgibson:
This PR updates the linear solvers in Gusto. This includes a small QOL change that makes all time-stepping coefficients `ufl.Constant`s. This avoids repeated code-generation when all you want to do is change the time-step parameter.

More importantly, the hybridized compressible solver has a been refactored and organized. Rather than hand-coding the Slate expressions in the solver, we can do the static condensation directly via Firedrake's tested preconditioner `firedrake.SCPC`. This avoids a lot of potential mistakes that could be made when writing out the condensation operators by hand.

In fact, it turns out using `SCPC` has resolved some issues previously encountered by @tommbendall. Most notably, it appears that #169 has been fixed when using this branch.

Also, the code is much simpler to read I think.

### DIFF
--- a/gusto/linear_solvers.py
+++ b/gusto/linear_solvers.py
@@ -298,7 +298,6 @@ class HybridizedCompressibleSolver(TimesteppingSolver):
                          'pc_python_type': 'firedrake.SCPC',
                          'pc_sc_eliminate_fields': '0, 1',
                          'condensed_field': {'ksp_type': 'fgmres',
-                                             'ksp_monitor_true_residual': True,
                                              'ksp_rtol': 1.0e-8,
                                              'ksp_atol': 1.0e-8,
                                              'ksp_max_it': 100,
@@ -323,6 +322,10 @@ class HybridizedCompressibleSolver(TimesteppingSolver):
             if any(deg > 2 for deg in dgspace.ufl_element().degree()):
                 state.logger.warning("default quadrature degree most likely not sufficient for this degree element")
             self.quadrature_degree = (5, 5)
+
+        # Turn monitor on for the trace system when running in debug mode
+        if state.output.log_level == DEBUG:
+            self.solver_parameters["condensed_field"]["ksp_monitor_true_residual"] = True
 
         super().__init__(state, solver_parameters, overwrite_solver_parameters)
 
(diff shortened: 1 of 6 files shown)


gusto/state.py
```diff
@@ -326,14 +326,13 @@ def setup_dump(self, t, tmax, pickup=False):
...
-            else:
-                if not running_tests:
-                    makedirs(self.dumpdir)
+            if self.mesh.comm.rank == 0:
+                if not running_tests and path.exists(self.dumpdir) and not pickup:
+                    raise IOError("results directory '%s' already exists"
+                                  % self.dumpdir)
+                else:
+                    if not running_tests:
+                        makedirs(self.dumpdir)
```
thomasgibson: This bit here fixes #190. I just left it in this PR since it was a very small change to the existing logic.

jshipton: Great - thanks! It's be good to have this in separately if the test fixes prove tricky...

gusto/linear_solvers.py
```diff
@@ -306,37 +323,42 @@ def __init__(self, state, quadrature_degree=None, solver_parameters=None,
                 state.logger.warning("default quadrature degree most likely not sufficient for this degree element")
             self.quadrature_degree = (5, 5)
 
+        # Turn monitor on for the trace system when running in debug mode
+        if state.output.log_level == DEBUG:
```
jshipton: I just merged a PR that has replaced all the instances of `if state.output.log_level == DEBUG` with `if logger.inEnabledFor(DEBUG)`. I think that in the future we will not have to pass state around so much...

thomasgibson: Yep I noticed! I'm fixing it locally now as I'm addressing some Jenkins shenanigans

thomasgibson:
Okay something is very strange. I'm passing the tests which "fail" on Jenkins using my laptop (macbook). In fact, @tommbendall reports problems on Ubuntu with this branch. Example `dry_bf_bubble.py` is one test where it fails on Jenkins and Ubuntu but works just fine on mac. I have no idea how to proceed from here.

dham:
Try logging in to pex.ma.ic.ac.uk and running tests there. That's an ubuntu machine so you ought to at least be able to reproduce the issue.

thomasgibson:
OK, green lights on Jenkins now!

PR #196, opened 2019-02-06
fixes for stricter lint rules

jshipton:
A couple of small fixes as lint seems to have stricter rules for comparing things.


jshipton:
I can't pass lint on several branches so I thought I'd fix it separately... just waiting for tests to pass and I'll merge. @tommbendall I suggested you as a reviewer as the changes are mostly in `limiters.py` and I don't quite trust myself not to miss a `not`!

tommbendall:
Definitely looks as if it was fine from my point of view!

Issue #86, opened 2017-02-24
detecting if element is DG/CG


jshipton:
Currently broken and maybe not the best way of doing it anyway - add a user option to interface instead?

thomasgibson:
Couldn't this be determined by examining `V.ufl_element().sobolev_space()`, where `V` is your function space?

thomasgibson:
Can be done by checking `V.ufl_element().sobolev_space().name`

jshipton:
@thomasgibson that looks like it should work - thanks! Is there a way of figuring out in which direction a space is discontinuous? i.e. for the theta space?

thomasgibson:
Yes! So if `V` is your theta space, then `sobolev_space = V.ufl_element().sobolev_space()` gives you a `DirectionalSobolevSpace`. In the case of `V`, it will be `DirectionalH`. You can then index the space by the spatial indices (0, 1, or 2 (if 3d)). And you will get that `sobolev_space[i].name` is `"H1"` or `"L2"`

jshipton:
Ah, that's what the H means! Cool - thanks!

jshipton:
Hm, what if I have a `VectorFunctionSpace`? I'm getting: AttributeError: 'VectorElement' object has no attribute 'sobolev_space'

thomasgibson:
Oh I see. Might need fixing in UFL later. But for the moment, you can ask a `VectorElement` for its `sub_elements`: `E._sub_elements` and that gives a list of the `FiniteElement` objects in the spatial directions. Each of those should have a sobolev space method

jshipton:
How about `BrokenElement`?

jshipton:
oh, of course, that's always DG but it'll need a separate check...

jshipton:
This has been fixed.

PR #194, opened 2019-02-04
Hybridized compressible solver parameters and updates

thomasgibson:
This fixes the issue (from the Gusto Slack channel) with the hydrostatic mountain test case. The problem was that the default order of elimination in `firedrake.SCPC` attempted to invert a singular matrix due. For the `h_mountain.py` example, parameters have been set to reverse the order. This fixes the issue.

I also made some other minor changes/corrections. Including adding the `unsaturated_bubble.py` test to the hybrid solver tests. It seems that #169 is resolved, as I was able to run the example to completion without error.

PR #197, opened 2019-02-06
Tidy transport eqn otps

jshipton:
This pull request sorts out the transport equation options in 2 ways.

1) the `ibp` (integrate by parts) option is now an Enum class.

2) uses the names of the ufl element Sobolev spaces to figure out continuity properties, as discussed in issue #86 

[Note: the rest of transport_equation.py will be significantly tidier when FML comes in, but I realised that the FML branch is based on this branch so I wanted to separate the PRs]


gusto/transport_equation.py
```diff
@@ -1,13 +1,101 @@
...
+    """
+    Function to check is a given space, V, is CG. Broken elements are
+    always discontinuous; for vector elements we check the names of
+    the sobolev spaces of the subelements and for all other elements
+    we just check the sobolev space name.
+    """
+    ele = V.ufl_element()
+    if isinstance(ele, BrokenElement):
+        return False
+    elif type(ele) == VectorElement:
```
thomasgibson: Small comment: Is there a reason this is used over `isinstance(ele, VectorElement)`?

gusto/transport_equation.py
```diff
@@ -1,13 +1,101 @@
...
+    """
+    Function to check is a given space, V, is DG. Broken elements are
+    always discontinuous; for vector elements we check the names of
+    the sobolev spaces of the subelements and for all other elements
+    we just check the sobolev space name.
+    """
+    ele = V.ufl_element()
+    if isinstance(ele, BrokenElement):
+        return True
+    elif type(ele) == VectorElement:
```
thomasgibson: Same question as above

gusto/transport_equation.py
```diff
@@ -1,13 +1,101 @@
...
+    if is_cg(V):
+        return None
+    else:
+        # for extruded meshes we have to unpick things
+        if V.extruded:
+            dim = V.mesh().topological_dimension()
+            ele = V.ufl_element()
+
+            # Broken elements are discontinuous so need dS_v and dS_h
+            if type(ele) == BrokenElement:
```
thomasgibson: Could become `if is_dg(ele):` yeah?

gusto/transport_equation.py
```diff
@@ -1,13 +1,101 @@
...
+
+            # return surface measures
+            if is_discontinuous_h and is_discontinuous_v:
+                return (dS_v + dS_h)
+            elif is_discontinuous_h:
+                return dS_v
+            elif is_discontinuous_v:
+                return dS_h
+            else:
+                raise ValueError("Sorry, we have failed to figure out the surface measures for this space")
```
thomasgibson: So polite :)

gusto/transport_equation.py
```diff
@@ -1,13 +1,101 @@
...
+
+            # Figure out which spaces we have
+            elif type(ele) == VectorElement:
+                spaces = [ele._sub_elements[i].sobolev_space().name
+                          for i in range(dim)]
+            else:
+                spaces = [ele.sobolev_space()[i].name for i in range(dim)]
+
+            # Based on the space names, these ones are discontinuous
+            discont = [i for i, name in enumerate(spaces) if name in ["L2", "HDiv"]]
```
thomasgibson: Question: should HDiv be treated as discontinuous? It's not as discontinuous as L2, but not as continuous as H1. I guess for the purpose of this function, you still have surface terms anyways?

gusto/transport_equation.py
```diff
@@ -38,30 +126,7 @@ def __init__(self, state, V, *, ibp="once", solver_params=None):
...
-        else:
-            if V.extruded:
-                self.dS = (dS_h + dS_v)
-                self.ds = (ds_b + ds_t + ds_v)
-            else:
-                self.dS = dS
-                self.ds = ds
-            self.n = FacetNormal(state.mesh)
-            self.un = 0.5*(dot(self.ubar, self.n) + abs(dot(self.ubar, self.n)))
+        self.dS = surface_measures(V)
```
thomasgibson: What happened to the exterior surface measures (`ds`, `ds_tb`, etc)? Do we not need them?

gusto/transport_equation.py
```diff
@@ -162,40 +228,45 @@ def __init__(self, state, V, *, ibp="once", equation_form="advective",
...
+                            dot(self.ubar('+'), n('+'))*q('+'))
                       + inner(self.test('-'),
-                              dot(self.ubar('-'), self.n('-'))*q('-')))*self.dS
+                              dot(self.ubar('-'), n('-'))*q('-')))*self.dS
 
         if self.outflow:
-            L += self.test*self.un*q*self.ds
+            n = FacetNormal(self.state.mesh)
+            un = 0.5*(dot(self.ubar, n) + abs(dot(self.ubar, n)))
+            L += self.test*un*q*(ds_v + ds_t + ds_b)
```
thomasgibson: Oh I see. They are here now!

gusto/transport_equation.py
```diff
@@ -162,40 +228,45 @@ def __init__(self, state, V, *, ibp="once", equation_form="advective",
...
 
         if self.vector_manifold:
-            un = self.un
+            n = FacetNormal(self.state.mesh)
             w = self.test
-            u = q
-            n = self.n
             dS = self.dS
+            u = q
```
thomasgibson: Do we need this redefinition, or can we just use `inner(q('+'), n('+'))` etc below?

gusto/transport_equation.py
```diff
@@ -290,55 +362,43 @@ def __init__(self, state, V, ibp="twice", equation_form="advective", supg_params
...
-            # need to include surface integrals on the horizontal faces
-            self.dS = dS_h
+        if supg_params is None:
+            supg_params = SUPGOptions()
+
+        dim = state.mesh.topological_dimension()
+        if supg_params.tau is not None:
+            # if tau is provided, check that is has the right size
+            tau = supg_params.tau
+            assert tau.ufl_shape == (dim, dim)
```
thomasgibson: Might be helpful to also return an error message. This will just raise an `AssertionError`. Something like:
```python
assert tau.ufl_shape == (dim, dim), "Provided tau has incorrect shape!"
```
Also, what if someone provides a `tau` which is not a UFL expression? It won't have a `ufl_shape` and trigger an `AttributeError`. One option would be to raise an error if it isn't, or simply cast via `as_ufl(supg_params.tau)`.

gusto/transport_equation.py
jshipton: heh, as I was writing this I thought I should probably check whether `type` or `isinstance` is considered better and was planning on asking @wence- so thanks for flagging it to remind me!

`isinstance` takes into account inheritance whereas `type` doesn't but I don't know enough about the structure of `ufl` for that information to be useful to me.

gusto/transport_equation.py
jshipton: Yes, for the purpose of this function we need the surface terms. Maybe I should change the comment to "these ones are discontinuous enough to require surface terms"?

gusto/transport_equation.py
jshipton: We only use them for outflow boundary conditions which are only used in @tommbendall's rain parameterisations. I thought it best to remove them from a more general function otherwise it looks like we might do something with them more general. Boundary conditions are something I plan to sort out, or at least improve in some way, after FML...

gusto/transport_equation.py
jshipton: Yes, good point - thanks!

gusto/transport_equation.py
jshipton: :)

gusto/transport_equation.py
jshipton: I'm happy with `q` but @colinjcotter wrote this bit (although it looks like I've edited it here - not sure how/why!)

jshipton: I think it would only be used for velocity so `u` makes sense...

jshipton:
Test failure is the hybridised version of the hydrostatic mountain - i.e. not to do with this PR

thomasgibson:
#194 is an updated version and is stable so far. Nearly finished running every compressible example to completion. Everything is looking good.

thomasgibson:
Overall, I'm happy with this! Just merge master into this branch and let the test suite run.

PR #198, opened 2019-02-12
Diagnostics for spherical components

tommbendall:
A small pull request which quickly allows for visualisation of meridional, zonal and radial components of vector fields by adding these as diagnostics. They are tested through `test_sw_triangle.py`.

### DIFF
--- a/gusto/diagnostics.py
+++ b/gusto/diagnostics.py
@@ -206,15 +206,15 @@ class SphericalComponent(DiagnosticField):
         if not self._initialised:
             # check geometric dimension is 3D
             if state.mesh.geometric_dimension() != 3:
-                raise ValueError('Spherical components only work in 3D!')
+                raise ValueError('Spherical components only work when the geometric dimension is 3!')
             space = FunctionSpace(state.mesh, "CG", 1)
             super().setup(state, space=space)
 
         V = VectorFunctionSpace(state.mesh, "CG", 1)
         self.x, self.y, self.z = SpatialCoordinate(state.mesh)
-        self.x_hat = Function(V).interpolate(as_vector([Constant(1.0), 0.0, 0.0]))
-        self.y_hat = Function(V).interpolate(as_vector([0.0, Constant(1.0), 0.0]))
-        self.z_hat = Function(V).interpolate(as_vector([0.0, 0.0, Constant(1.0)]))
+        self.x_hat = Function(V).interpolate(Constant(as_vector([1.0, 0.0, 0.0])))
+        self.y_hat = Function(V).interpolate(Constant(as_vector([0.0, 1.0, 0.0])))
+        self.z_hat = Function(V).interpolate(Constant(as_vector([0.0, 0.0, 1.0])))
         self.R = sqrt(self.x**2 + self.y**2)  # distance from z axis
         self.r = sqrt(self.x**2 + self.y**2 + self.z**2)  # distance from origin
         self.f = state.fields(self.fname)
@@ -226,10 +226,12 @@ class MeridionalComponent(SphericalComponent):
 
     @property
     def name(self):
-        return self.fname+"_meridional_component"
+        return self.fname+"_meridional"
 
     def compute(self, state):
-        lambda_hat = (self.x * self.y_hat - self.y * self.x_hat) / self.R
+        lambda_hat = (-self.x * self.z * self.x_hat / self.R
+                      - self.y * self.z * self.y_hat / self.R
+                      + self.R * self.z_hat) / self.r
         return self.field.project(dot(self.f, lambda_hat))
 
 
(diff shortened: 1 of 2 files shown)


gusto/diagnostics.py
```diff
@@ -195,6 +196,67 @@ def compute(self, state):
...
+
+    def __init__(self, name):
+        super().__init__()
+        self.fname = name
+
+    def setup(self, state):
+        if not self._initialised:
+            # check geometric dimension is 3D
+            if state.mesh.geometric_dimension() != 3:
+                raise ValueError('Spherical components only work in 3D!')
```
jshipton: This is a bit confusing as shallow water is 2D, although I see what you mean that the mesh is 3D.

gusto/diagnostics.py
```diff
@@ -195,6 +196,67 @@ def compute(self, state):
...
+            space = FunctionSpace(state.mesh, "CG", 1)
+            super().setup(state, space=space)
+
+        V = VectorFunctionSpace(state.mesh, "CG", 1)
+        self.x, self.y, self.z = SpatialCoordinate(state.mesh)
+        self.x_hat = Function(V).interpolate(as_vector([Constant(1.0), 0.0, 0.0]))
```
jshipton: Can `Constant` come outside `as_vector`? The zeros are constant too.

tests/test_sw_triangle.py
```diff
@@ -38,7 +38,10 @@ def setup_sw(dirname, euler_poincare):
                          Difference('SWPotentialEnstrophy_from_PotentialVorticity',
                                     'SWPotentialEnstrophy_from_RelativeVorticity'),
                          Difference('SWPotentialEnstrophy_from_PotentialVorticity',
-                                    'SWPotentialEnstrophy_from_AbsoluteVorticity')]
+                                    'SWPotentialEnstrophy_from_AbsoluteVorticity'),
+                         MeridionalComponent('u'),
```
jshipton: Can we check we get the right thing here?

tommbendall: What kind of check do you have in mind?

gusto/diagnostics.py
tommbendall: Hopefully that's better!

gusto/diagnostics.py
tommbendall: Done!

gusto/diagnostics.py
```diff
@@ -195,6 +196,67 @@ def compute(self, state):
...
+        if not self._initialised:
+            # check geometric dimension is 3D
+            if state.mesh.geometric_dimension() != 3:
+                raise ValueError('Spherical components only work when the geometric dimension is 3!')
+            space = FunctionSpace(state.mesh, "CG", 1)
+            super().setup(state, space=space)
+
+        V = VectorFunctionSpace(state.mesh, "CG", 1)
+        self.x, self.y, self.z = SpatialCoordinate(state.mesh)
+        self.x_hat = Function(V).interpolate(Constant(as_vector([1.0, 0.0, 0.0])))
```
thomasgibson: I think `Constant((1.0, 0.0, 0.0))` works as well

tests/test_sw_triangle.py
jshipton: This is Williamson test 2 isn't it? So we couldn't we check that the meridional and vertical components are zero and the zonal component has the expected maximum?

thomasgibson: It will be near-zero if you're using super bendy elements (mesh `degree=42`). But I think a check that it's "small" relative to the degree of the mesh would be good enough.

jshipton: Yes, sorry, by "zero" I meant "appropriately small for the mesh size and degree" :)

tommbendall: I agree completely and think these are very sensible suggestions. I wanted to do something like this to do the check: 

```u_max = data.groups["u"]["max"][0]
tolerance = 0.05

u_meridional = data.groups["u_meridional"]
assert u_meridional["max"][0] < tolerance * u_max

u_radial = data.groups["u_radial"]
assert u_radial["max"][0] < tolerance * u_max

u_zonal = data.groups["u_zonal"]
assert u_max * (1 - tolerance) < u_zonal["max"][0] < u_max * (1 + tolerance)
```

BUT for some reason `u_max` is huge, like 4e9 m/s (faster than the speed of light!) when it should be about 38 m/s. Not really sure why this is and it doesn't show up when you plot the field.

Not sure what to do in this case!

thomasgibson: Something like that would _definitely_ show up in the plotted field. How is `max` obtained?

tommbendall: Through this `maxify` function in `diagnostics.py`:
```
@staticmethod
    def max(f):
        fmax = op2.Global(1, np.finfo(float).min, dtype=float)
        op2.par_loop(op2.Kernel("""
void maxify(double *a, double *b) {
    a[0] = a[0] < fabs(b[0]) ? fabs(b[0]) : a[0];
}
""", "maxify"), f.dof_dset.set, fmax(op2.MAX), f.dat(op2.READ))
return fmax.data[0]
```

In the debugger though I did:
```
u_array  = np.array(u0.dat.data[:])
u_array.max()
```
and got out a similarly large value, so it does look like there's a DOF with a very high value?

jshipton: your `u_max` is in the HDiv space - its numerical values are not useful!

jshipton: to check the diagnostic you can just use the value that we know `u_max` takes - it is defined in the setup part of the file

tommbendall: Of course, thanks Jemma! It seems a bit messy to have to define the value of `u_max` in the `test_sw_setup` part as well as the `setup_sw` function though?

The question whether there is a way of doing this that isn't messy! Another way could be to open the netcdf file at the end of the setup file and add a one-off `u_max`, something like
```
data = Dataset("results/"+dirname+"/sw/diagnostics.nc", 'a')
data.createGroup('umax')
data['umax'].createDimension('value', 1)
data['umax'].createVariable('umax', float ('value',))
data['umax']['umax'][0] = u_max
```
but this also seems messy!

jshipton: I don't think we should mess with the netCDF file. Can't you just move the lines defining `u_max` outside of the function definitions so that all functions can see them?

tommbendall: Yes! Brain not working today...

PR #199, opened 2019-02-14
Programmable eqn transformations

jshipton:
PR for discussion: note that all I have done is reproduce our current timestepping schemes for the shallow water equations and the advection schemes. Once the overall structure has been agrees, it will be straightforward to define classes for the other equation sets that we solve.

Requires generic_replace branch of ufl.

The following tests should work:
test_dg_advection, test_embedded_dg_advection, test_supg_advection, test_advection_diffusion and test_sw_triangle, and test_labelling

Briefly, the setup is as follows:

- The equation form is defined in a child class of `PrognosticEquation` or `PrognosticMixedEquation` class in equations.py. When instantiating the class you provide `state`, the function space and the prognostic field name. The base class creates the field and registers it so that diagnostics are calculated. The base class also defines and labels the mass term.

- There are 2 timestepper classes:

`Timestepper` takes in a tuple of (equation, schemes) where `schemes` is either an Advection class or a tuple of (Advection class, *active_labels) where the labels specify which parts of the form to select. This means, for example, we can apply different schemes to advection and diffusion (see test_advection_diffusion).

`PrescribedAdvectionTimestepper` uses the velocity in state.fields("u") as the advecting velocity.

`CrankNicolson` also takes in a tuple of (equation, schemes) but this is now for any additional equations (such as tracer advection) - the main `equation_set` (e.g. ShallowWaterEquations) is passed in separately, along with a list of `advected_fields` and `diffused_fields`. This keeps the interface the same (although I'm happy to change it if that's better). The user no longer has to specify the `LinearSolver` or `Forcing` classes as these are now generic as they get forms from the `PrognosticMixedEquation` class.

- "Advection" schemes (these should probably be renamed to "Timestepping" schemes as they no longer just deal with the advection terms). On instantiation the user just specifies optional arguments, such as a limiter or spatial discretisation options. The schemes are set up from the timestepper class, mainly because the timestepper class defines the advecting velocity.

- `LinearTimesteppingSolver` (this could use a better name) The aim is that this should replace all the other classes in the linear_solvers.py file. It was straightforward to replace the shallow water specific class but the others will need some work to sort out the manipulations.

Some remaining issues:

- How to deal with linear equations (we use these to test schemes and for debugging).
- How should the timestepping schemes (in advection.py) label their terms? At the moment `BackwardEuler` just labels everything as `implicit` and `ExplicitAdvection` labels everything `explicit`. For IMEX schemes I think we're going to need to name the terms in the equation and have some default (name, time_level) pairing which the user can change if they want. We may also want to do this so that we can chose which terms to include in the `LinearTimesteppingSolver` - e.g. sometimes we put the Coriolis term in.


gusto/form_manipulation_labelling.py
```diff
@@ -0,0 +1,328 @@
...
+
+    Returns a function that takes in t, a :class:`Term`, and returns
+    a new :class:`Term` with form containing the new_test and
+    labels=t.labels
+    """
+
+    def rep(t):
+        test = t.form.arguments()[0]
+        new_form = ufl.replace(t.form, {test: new_test})
+        return Term(new_form, t.labels)
```
thomasgibson: Does this work if `new_test` and `test` are `TestFunctions` from a mixed finite element space? What about more general replacements like replacing test functions on a mixed space of two spaces `W = U * V` with ones in `Wtilde = U * V * T`?

gusto/form_manipulation_labelling.py
```diff
@@ -0,0 +1,328 @@
...
+            # apply to part of the form so we cannot replace it.
+            if type(labelled) is bool:
+                raise ValueError("This label does not label a part of this term's form")
+
+            # flag if replacer is a ufl expression - if it is then we
+            # trust it is of the right shape below
+            repl_expr = (
+                isinstance(replacer, ufl.core.expr.Expr)
+                and not isinstance(replacer, Function)
+            )
```
thomasgibson: If we know the shape that it _should_ be, would it be possible to add a check here? That way, we know when we are doing some illegal stuff. I think we should be able to ask for the `ufl_shape` and use that to determine if it's sensible

gusto/form_manipulation_labelling.py
```diff
@@ -0,0 +1,328 @@
...
+        elif isinstance(other, Term):
+            return LabelledForm(self, other)
+        elif isinstance(other, LabelledForm):
+            return LabelledForm(self, *other.terms)
+        else:
+            return NotImplemented
+
+    __radd__ = __add__
+
+    def __rmul__(self, other):
```
thomasgibson: Wouldn't it be better if here we define the `__mul__` method and just set `__mul__ = __rmul__`? I guess in the end, Python will first try to call `__mul__` and notice it doesn't have one, then the right argument's `__rmul__` will be called.

gusto/form_manipulation_labelling.py
```diff
@@ -0,0 +1,328 @@
...
+        if type(other) is Term:
+            return LabelledForm(*self, Constant(-1.)*other)
+        elif type(other) is LabelledForm:
+            return LabelledForm(*self, *[Constant(-1.)*t for t in other])
+        elif other is None:
+            return self
+        else:
+            return NotImplemented
+
+    def __rmul__(self, other):
```
thomasgibson: Same question as above

gusto/timeloop.py
```diff
@@ -138,108 +172,245 @@ def run(self, t, tmax, pickup=False):
...
 
-    def __init__(self, state, advected_fields, linear_solver, forcing,
-                 diffused_fields=None, physics_list=None, prescribed_fields=None):
+    def __init__(self, state, equation_set, advected_fields,
+                 diffused_fields=None, equations_schemes=None,
+                 physics_list=None, prescribed_fields=None, **kwargs):
+
+        self.maxk = kwargs.pop("maxk", 4)
+        self.maxi = kwargs.pop("maxi", 1)
+        self.alpha = kwargs.pop("alpha", 0.5)
```
thomasgibson: I think we should start trying to `Constant`ize coefficients that we know won't change throughout a run. That way, if we want to change the parameters and rerun a simulation, we won't have to regenerate and recompile code.

gusto/form_manipulation_labelling.py
jshipton: I haven't tried... in fact I make sure that we assert that the function space isn't mixed before we use this (see line 240 of advection.py). At the moment this function is only used in the `CrankNicolson` timestepper where we advect the fields separately. At the moment I am only trying to replicate the functionality that we already have. I imagine when I want to apply an IMEX scheme to the mixed problem, I will have to revisit this....

gusto/form_manipulation_labelling.py
jshipton: yes, that's probably wise

gusto/form_manipulation_labelling.py
jshipton: I'm happy to do that if it's the right way - I don't have much experience with these magic methods.

gusto/timeloop.py
jshipton: I think I was pretty careful to wrap `Constant` around things when the forms are manipulated. In fact, if you look in the `__rmul__` method of `Term` you'll see that I wrap floats with `Constant` so even if someone forgets the `Constant` it'll be done anyway.

gusto/form_manipulation_labelling.py
dham: Yes. That's the right thing to do.

gusto/advection.py
```diff
@@ -37,74 +48,254 @@ class Advection(object, metaclass=ABCMeta):
...
+    :arg solver_parameters: (optional) solver_parameters
+    :arg limiter: (optional) :class:`.Limiter` object.
+    :arg options: (optional) :class:`.AdvectionOptions` object
     """
 
-    def __init__(self, state, field, equation=None, *, solver_parameters=None,
-                 limiter=None):
+    def __init__(self, *,
+                 solver_parameters=None,
+                 limiter=None, options=None):
```
jshipton: You will notice that this `__init__` method is fairly empty. It would be possible to pass in the equation here and do a lot of the manipulations that are currently in the `setup` and `_setup_from_options` method. The only thing that really has to wait is replacing the advecting velocity.

gusto/equations.py
```diff
@@ -0,0 +1,388 @@
...
+                    # integrators.
+                    u_form += -g*div(w)*field*dx
+
+        # define the depth continuity term and its linearisation
+        Dadv = continuity_form(state, W, 1)
+        Dadv_linear = linear_continuity_form(state, W, 1, qbar=H).label_map(
+            all_terms, replace_labelled(trials, subject, advecting_velocity))
+        D_form = linearisation(Dadv, Dadv_linear)
+
+        return index(u_form, 0) + index(D_form, 1)
```
jshipton: This `index` label is possibly not necessary but maybe I don't know enough ufl. I want to be able to extract the part of the form that corresponds to the `u` equation or the `D` equation. I thought I would be able to do this by examining the function space of the test function but I couldn't figure out a way - any thoughts?

gusto/timeloop.py
```diff
@@ -138,108 +172,245 @@ def run(self, t, tmax, pickup=False):
...
+             iterations and alpha is the offcentering parameter
     """
 
-    def __init__(self, state, advected_fields, linear_solver, forcing,
-                 diffused_fields=None, physics_list=None, prescribed_fields=None):
+    def __init__(self, state, equation_set, advected_fields,
+                 diffused_fields=None, equations_schemes=None,
+                 physics_list=None, prescribed_fields=None, **kwargs):
+
+        self.maxk = kwargs.pop("maxk", 4)
```
jshipton: I have got rid of the `TimesteppingParameters` class because different time steppers have different options and we can just pass them in, as here. However, this does mean that we pass `dt` around a bit - would it be better to make `dt` global, and how does one go about doing that?

jshipton:
Closing as superseded by FML merge #261

PR #204, opened 2019-04-11
Stop grandmas falling from the sky

tommbendall:
If negative some negative moisture concentrations formed (possibly through advection) this could cause nans in the `Coalescence` and `Evaporation` schemes (it turns out you shouldn't raise negative numbers to non-integer powers!)

This very short pull request fixes this, adding some `conditionals` to the limiting step.

PS: sorry for the terrible name...

### DIFF
--- a/gusto/physics.py
+++ b/gusto/physics.py
@@ -294,17 +294,17 @@ class Coalescence(Physics):
         coalesce_rate = Function(Vt)
 
         # adjust coalesce rate using min_value so negative cloud concentration doesn't occur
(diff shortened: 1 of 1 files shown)


jshipton:
@tommbendall looks like this needs merging from master please...

tommbendall:
I've rerun this on Jenkins and again it's failing because of `h_mountain.py`. Again, I can't reproduce... I am going to raise an issue, but given that the same error occurs on other pull requests it is not related to this pull request. Are people happy for me to merge this anyway?

PR #206, opened 2019-04-12
Fix/loopy changes

wence-:
Ports the custom kernels to the new kernel interface. Although we can still use string kernels, one level of indirection has been removed. I also correct some of the access descriptors.

Needs firedrakeproject/firedrake#1408 to land simultaneously.

### DIFF
--- a/gusto/recovery.py
+++ b/gusto/recovery.py
@@ -7,7 +7,7 @@ from firedrake import (expression, function, Function, FunctionSpace, Projector,
                        dx, Interpolator, quadrilateral, BrokenElement, interval,
                        TensorProductElement, FiniteElement)
 from firedrake.utils import cached_property
-from firedrake.parloops import par_loop, READ, INC, RW
+from firedrake.parloops import par_loop, READ, INC, RW, WRITE
 from pyop2 import ON_TOP, ON_BOTTOM
 import ufl
 import numpy as np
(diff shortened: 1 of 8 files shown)


wence-:
The test suite passes for me locally with these changes and the relevant firedrake etc... branches.

gusto/recovery.py
```diff
@@ -163,91 +167,91 @@ def __init__(self, v0, v1, v_out, method='density'):
...
-            float y = COORDS[1][1] - COORDS[0][1];
-            float a = CG1[3][0];
-            float b = CG1[1][0];
-            float c = DG0[0][0];
-            DG1[1][0] = a;
-            DG1[3][0] = b;
-            DG1[2][0] = (1.0 / (pow(x, 2.0) + 4.0 * pow(y, 2.0))) * (-3.0 * a * pow(y, 2.0) - b * pow(x, 2.0) - b * pow(y, 2.0) + 2.0 * c * pow(x, 2.0) + 8.0 * c * pow(y, 2.0));
-            DG1[0][0] = 4.0 * c - b - a - DG1[2][0];
+            float x = COORDS[2*2] - COORDS[0];
+            float y = COORDS[2*1 + 1] - COORDS[1];
```
tommbendall: To make sure I've got this right, for vector function `f` in spatial dimension `d` we have `f[a][b] -> f[a*d + b]` ?

wence-: Yes.

wence-:
Testing on Jenkins, I hope it will be happy and will then hit go on Monday.

PR #207, opened 2019-05-05
Update limiters for loopy

tommbendall:
This updates the code for the `ThetaLimiter` so that it uses loopy kernels (and so will not raise the c parloop warning).

I have also generalised the `ThetaLimiter` so that it will work on the temperature space in 3D.

### DIFF
--- a/gusto/limiters.py
+++ b/gusto/limiters.py
@@ -1,5 +1,4 @@
 from __future__ import absolute_import, print_function, division
-from gusto.configuration import logger
 from firedrake import dx, BrokenElement, Function, FunctionSpace
 from firedrake.parloops import par_loop, READ, WRITE, INC
 from firedrake.slope_limiter.vertex_based_limiter import VertexBasedLimiter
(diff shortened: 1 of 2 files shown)


tommbendall:
> Looks fine to me. The Jenkin's failure is related to the infamous `h_mountain.py` problem. We will worry about it separately. I think this should be okay. Can you confirm with a short, 3D example that this indeed works?

I can confirm that this limiter does indeed work, I haven't added the example as it would be too long to run in 3D.

tommbendall:
I've just updated the checks for the space being used so that it raises an error if the space used isn't the right one. After this has run the tests then this PR should be ready.

gusto/limiters.py
```diff
@@ -58,63 +23,122 @@ def __init__(self, space):
...
 
         self.Vt = FunctionSpace(space.mesh(), BrokenElement(space.ufl_element()))
-        # check this is the right space, only currently working for 2D extruded mesh
+        # check this is the right space, currently working for 2D and 3D extruded meshes
+        # check that horizontal degree is 1 and vertical degree is 2
+        if self.Vt.ufl_element().degree()[0] != 1 or \
+           self.Vt.ufl_element().degree()[1] != 2:
+            raise ValueError('This is not the right limiter for this space.')
+
         if self.Vt.extruded and self.Vt.mesh().topological_dimension() == 2:
```
jshipton: maybe just have a separate assert that the mesh is extruded

jshipton:
Merging - the only failure is the infamous hydrostatic mountain.

PR #210, opened 2019-05-13
Non periodic domains

tommbendall:
The intention of this pull request is to enable Gusto to run in non-periodic domains, such as 3D channels that are only periodic in one direction.

This involves two aspects:
1) Adding extra measures for vertical side-walls to the linear solver (it was already done for outflow in the advection step)
2) Changing the way that boundary conditions are called. They are now a property of `state`, which has an argument `u_bc_ids` in which you can list the ids of additional boundaries that need zero-velocity boundary conditions. I have changed the `dry_bf_bubble` to demonstrate how this is done (the test in their paper isn't on a periodic domain). The intention is that when the code is rewritten using form labelling, boundary conditions will be taken away from `state`.

### DIFF
--- a/examples/dry_bf_bubble.py
+++ b/examples/dry_bf_bubble.py
@@ -69,7 +69,7 @@ state = State(mesh, vertical_degree=degree, horizontal_degree=degree,
               diagnostics=diagnostics,
               fieldlist=fieldlist,
               diagnostic_fields=diagnostic_fields,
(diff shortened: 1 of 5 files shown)


tommbendall:
This has been rerun and looks to have only failed on `h_mountain.py` (which we are hoping to prevent on an imminent pull request!)

Are we happy to merge this?

PR #211, opened 2019-05-30
Retire old compressible solver

thomasgibson:
This PR removes the old compressible solver, leaving just the hybridized solver as the "compressible solver." We discussed this a little while ago, and people seemed fine with the idea. Let's discuss here if there are things we need to consider!


thomasgibson:
Note that the jenkins/branch build is failing due to an install error. I'm not sure how to fix this... it's failing due to the error:
```
ERROR: Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/home/firedrake/firedrake/lib/python3.6/site-packages/cftime'
```

wence-:
You probably need to do something similar to the thetis builder to workaround username issues (jenkins always boots the docker container as user jenkins):

https://github.com/thetisproject/thetis/blob/master/Jenkinsfile#L29

tests/examples/test_examples_run.py
```diff
@@ -9,7 +9,8 @@
 examples_dir = join(cwd, "..", "..", "examples")
 example_files = glob.glob("%s/*.py" % examples_dir)
 
-@pytest.fixture(params=[pytest.param(f, marks=pytest.mark.xfail(reason="unreliable")) if f.endswith("h_mountain.py") else f for f in example_files],
+
+@pytest.fixture(params=glob.glob("%s/*.py" % examples_dir),
```
tommbendall: Is the xfail marking being removed because `mountain_hydrostatic.py` is now expected to pass?

gusto/linear_solvers.py
```diff
@@ -322,13 +134,13 @@ def __init__(self, state, quadrature_degree=None, solver_parameters=None,
...
-            self.solver_parameters["ksp_type"] = "gmres"
+            self.solver_parameters["ksp_type"] = "fgmres"
             self.solver_parameters["mat_type"] = "aij"
             self.solver_parameters["pmat_type"] = "matfree"
-            self.solver_parameters["ksp_monitor_true_residual"] = True
+            self.solver_parameters["ksp_monitor_true_residual"] = None
 
             # Turn monitor on for the trace system
-            self.solver_parameters["condensed_field"]["ksp_monitor_true_residual"] = True
+            self.solver_parameters["condensed_field"]["ksp_monitor_true_residual"] = None
```
tommbendall: I don't fully understand these changes, but happy to trust you on them. Only thing I was wondering is whether the comment on line 142 still matches what you are doing in 143?

examples/mountain_hydrostatic.py
```diff
@@ -157,34 +152,25 @@ def minimum(f):
...
+                              'ksp_atol': 1.0e-8,
+                              'ksp_max_it': 100,
+                              'pc_type': 'gamg',
+                              'pc_gamg_sym_graph': True,
+                              'mg_levels': {'ksp_type': 'gmres',
+                                            'ksp_max_it': 5,
+                                            'pc_type': 'bjacobi',
+                                            'sub_pc_type': 'ilu'}}}
+linear_solver = CompressibleSolver(state, solver_parameters=params,
+                                   overwrite_solver_parameters=True)
```
tommbendall: If I understand how `overwrite_solver_parameters` is meant to work, we could actually reduce this list to just:
```
params = {'pc_sc_eliminate_fields': '1, 0'}
```
I think this is the only difference with the default solver parameters?

thomasgibson: Oh good point. I'll chec that out.

gusto/linear_solvers.py
thomasgibson: Ah, this is because of a recent options change made in Firedrake. I don't really like it either, but basically booleans like `True` and `False` are no longer accepted when you want to turn on certain PETSc options. Options that are normally called via the command line act like switches, such as `python foo.py --options_prefix_ksp_monitor` would turn on the monitor. You don't pass `True`, you just pass the command. So now we just pass `None` so something is there. Then it gets easier to translate this when we pass options to PETSc.

tests/examples/test_examples_run.py
thomasgibson: Yes, because I think it should pass. I need to verify this though.

thomasgibson: Right now, I'm fighting with Jenkins...

thomasgibson:
Gusto is still failing to install on Jenkins. I'm moving this to another branch until I can find a working solution...

tests/examples/test_examples_run.py
tommbendall: OK, I think I still expect it to fail unless it's been updated?

tommbendall: Well I expect it to fail sometimes on Jenkins but to pass for everyone else...

gusto/linear_solvers.py
tommbendall: Ah cool, thanks for the explanation!

tests/examples/test_examples_run.py
thomasgibson: You're right. It still fails. This is a bad test case. Still don't know why and it's so extremely irritating.

tommbendall: I suggest you just put this xfail line back in and then at least it will pass!

thomasgibson: I'm pretty confident this will pass now.

examples/mountain_hydrostatic.py
tommbendall: We are going to attempt to solve this in a separate pull request

PR #213, opened 2019-07-15
ensure boundary recovery does happen in physics for 3D domains

tommbendall:
I realised I hadn't quite got the physics properly updated for the 3D recovery methods. I noticed I was getting condensation happening when it shouldn't in the bottom layer of the domain in baroclinic wave tests -- this fixes that!

### DIFF
--- a/gusto/physics.py
+++ b/gusto/physics.py
@@ -68,7 +68,7 @@ class Condensation(Physics):
 
         # make rho variables
         # we recover rho into theta space
-        if state.vertical_degree == 0 and state.mesh.geometric_dimension() == 2:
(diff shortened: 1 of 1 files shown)


jshipton:
Is it possible to test this? I expect not if you only saw it in the baroclinic wave but thought I'd check...

tommbendall:
> Is it possible to test this? I expect not if you only saw it in the baroclinic wave but thought I'd check...

Sadly not easily as it would need a 3D test which I'm reluctant to put in! Probably we should think about having some 3D tests in the future that don't get run as part of the suite...

PR #214, opened 2019-07-23
Ensure Vorticity in Cartesian Geometries

jtessier10:
Needed Vorticity for geometries other than 2D spherical. I believe this is correct and general enough to work for other geometries as well. Placing instead of NotImplementedError for now such that the 2D spherical geometry computation is unaffected.

- Jonathan Tessier

### DIFF
--- a/gusto/diagnostics.py
+++ b/gusto/diagnostics.py
@@ -705,7 +705,7 @@ class Vorticity(DiagnosticField):
                 gradperp = lambda psi: cross(cell_normals, grad(psi))
                 L = (- inner(gradperp(gamma), u))*dx
             else:
-                raise NotImplementedError("The vorticity diagnostics have only been implemented for 2D spherical geometries.")
+                L = (- inner(state.perp(grad(gamma)), u))*dx
 
             if vorticity_type != "relative":
                 f = state.fields("coriolis")
--- a/gusto/diagnostics.py
+++ b/gusto/diagnostics.py
@@ -700,12 +700,7 @@ class Vorticity(DiagnosticField):
             else:
                 a = q*gamma*dx
 
-            if state.on_sphere:
-                cell_normals = CellNormal(state.mesh)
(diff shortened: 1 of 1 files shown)


gusto/diagnostics.py
```diff
@@ -705,7 +705,7 @@ def setup(self, state, vorticity_type=None):
                 gradperp = lambda psi: cross(cell_normals, grad(psi))
                 L = (- inner(gradperp(gamma), u))*dx
             else:
-                raise NotImplementedError("The vorticity diagnostics have only been implemented for 2D spherical geometries.")
+                L = (- inner(state.perp(grad(gamma)), u))*dx
```
jshipton: You should be able to remove the code in the `if` block because you're getting the `perp` function from `state` so it is already correct on the sphere.

jshipton: P.S. thanks for fixing this - I should've done it this way in the first place!

jtessier10: No problem at all! `if` statement removed.

jshipton: Thanks! Unfortunately the code cannot currently be built because lint is failing. Lint checks that the code syntax is correct and in this case it is complaining because some functions are imported from Firedrake that are no longer used because the code in the `if` statement was the only place they were used in this file. You can check this locally by running `make lint` - let me know if you have trouble figuring it out...

jshipton: Oh, but the good news is that lint is run after the tests - so the tests all still pass after this change. `test_sw_triangle.py` compares the vorticity to an analytical expression for the vorticity so the computation still works on the sphere - the hard part is done!

jtessier10: Removing the unnecessary imports and a bit of white-space locally made lint pass. Would that be an acceptable solution?

jshipton: sounds perfect... expecting the tests to pass very soon then I will merge your pull request - many thanks!

PR #215, opened 2019-07-26
Sw bc fix

jshipton:
We recently added the facility to set no normal flow boundary conditions by passing the boundary ids in to state. This was done with the 3D baroclinic channel test case in mind, but we can also apply these to the shallow water equations as required by Francis and Jonathan.


jshipton:
> Looks fine to me -- presuming we don't want to add an example of shallow water flow in a channel?

Yes, at some point... :)

PR #216, opened 2019-09-11
Recovered diffusion

tommbendall:
This extends the diffusion process we have to work with the lowest order spaces. Basically it repeats the same trick as for advection: we recover into the higher order spaces and use the existing interior penalty method.

However the implementation is quite different: I invoke a new diffusion class called `RecoveredDiffusion` which takes a diffusion method, e.g. `InteriorPenalty` as an argument. I'm not really sure if this is a good way to do this, so very happy for feedback on changing this.

I've also updated the Straka dense bubble so that it can use the lowest order spaces (I've lalso updated the initial condition which I think was previously wrong!). I have also provided a test.


gusto/diffusion.py
```diff
@@ -71,3 +74,64 @@ def apply(self, x_in, x_out):
...
         x_out.assign(self.phi1)
+
+
+class RecoveredDiffusion(Diffusion):
+    """
+    A diffusion strategy for the lowest order spaces, allowing us to recover to a
+    higher order space, where diffusion will be performed.
+
+    :arg state: the state object.
+    :arg diffusion_scheme: an existing Diffusion class. For a vector field this should
```
jshipton: This seems a bit weird to me. I wonder if we should do recovered diffusion like we do recovered advection - i.e. with a decorator. Would be good to know if that is good coding or not...

examples/dense_bubble.py
```diff
@@ -75,8 +80,13 @@
     zc = 3000.
     zr = 2000.
     r = sqrt(((x[0]-xc)/xr)**2 + ((x[1]-zc)/zr)**2)
-    theta_pert = conditional(r > 1., 0., -7.5*(1.+cos(pi*r)))
-    theta0.interpolate(theta_b + theta_pert)
+
+    physics_boundary = Boundary_Method.physics if recovered else None
+    rho_b_Vt = Function(Vt)
+    rho_b_recoverer = Recoverer(rho_b, rho_b_Vt, VDG=Vt_brok, boundary_method=physics_boundary).project()
```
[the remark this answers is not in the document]
tommbendall: I actually called `project` in the same line as declaring the `Recoverer` which was a bit lazy. I'll separate them out to make it a bit clearer.

It's necessary for both sets of spaces as it is used to take `rho` into `Vt`, so that we algebraically evaluate exner pi in `Vt`

gusto/diffusion.py
tommbendall: I think the decorator idea is definitely better than this -- I'll give that a go.

PR #217, opened 2019-09-16
Update limiter project

tommbendall:
This is a correction to the `ThetaLimiter` which wasn't working as it should. This limiter is used with the Embedded DG advection scheme, which finishes with a projection step.

This pull request makes a simple change to the projection step, so that it uses the recovery operator rather than a Galerkin projection (this was what we intended to do originally).

I also took the opportunity to update the limiter test so that it was no longer separated into vertical and horizontal parts -- it's now a rotational test with both horizontal and vertical motion.

### DIFF
--- a/gusto/advection.py
+++ b/gusto/advection.py
@@ -111,7 +111,7 @@ class Advection(object, metaclass=ABCMeta):
                 self.x_brok_interpolator = Interpolator(self.xdg_out, x_brok)
                 self.x_out_projector = Recoverer(x_brok, self.x_projected)
             else:
-                self.x_out_projector = Projector(x_brok, self.x_projected)
+                self.x_out_projector = Projector(self.xdg_out, self.x_projected)
 
     def pre_apply(self, x_in, discretisation_option):
         """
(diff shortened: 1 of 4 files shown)


gusto/advection.py
```diff
@@ -92,6 +92,9 @@ def _setup(self, state, field, options):
             self.Projector = Projector(self.xdg_out, self.x_projected,
                                        solver_parameters=parameters)
 
+        if options.name == "embedded_dg" and self.limiter is not None:
```
jshipton: does this mean that self.Projector is defined but not used when `options.name=="embedded_dg"` and there is no limiter?

tommbendall: Yes you're right -- I could put this under an `else`

gusto/limiters.py
```diff
@@ -154,7 +150,6 @@ def apply(self, field):
         self.copy_vertex_values(field)
         self.vertex_limiter.apply(self.theta_hat)
         self.copy_vertex_values_back(field)
-        self.remap_to_embedded_space(field)
```
tommbendall: Since I am not using this function I should also delete it

gusto/advection.py
jshipton: maybe also find a name that works for both `self.Projector` and `self.x_recoverer` then you can simplify the code in lines 144-147

PR #218, opened 2019-09-16
remove S.force_evaluation now firedrake is updated

tommbendall:
Fix so hydrostatic balance should work again


jshipton:
Thanks for fixing this @tommbendall  - are you happy for it to be merged without waiting for Jenkins?

tommbendall:
> Thanks for fixing this @tommbendall - are you happy for it to be merged without waiting for Jenkins?

Yes I think it should be fine now (I tested it locally), I just didn't want to take full responsibility and force push!

PR #220, opened 2020-01-30
Fix failing kernels

tommbendall:
Update Gusto now that Firedrake uses finite elements with Gauss-Legendre points. Gusto should be extended in future so that these types of elements are supported.


jshipton:
This looks like it is not picking up the latest Firedrake...

tommbendall:
> 
> 
> This looks like it is not picking up the latest Firedrake...

I've restarted the runs, so hopefully that will do the trick!

PR #221, opened 2020-02-12
Dumping on a Lat-Lon mesh in 3D

tommbendall:
This extends the lat-lon mesh so that it can be used in 3D and with cubed sphere meshes.

There are also some updates to the Petsc solver options (as the old ones were becoming obsolete).

### DIFF
--- a/gusto/state.py
+++ b/gusto/state.py
@@ -571,39 +571,19 @@ class State(object):
 
 def get_latlon_mesh(mesh):
     coords_orig = mesh.coordinates
-    
-    if coords_orig.function_space().extruded:
-        cell = mesh._base_mesh.ufl_cell().cellname()
-        DG1_hori_elt = FiniteElement("DG", cell, 1, variant="equispaced")
-        DG1_vert_elt = FiniteElement("DG", interval, 1, variant="equispaced")
-        DG1_elt = TensorProductElement(DG1_hori_elt, DG1_vert_elt)
-    else:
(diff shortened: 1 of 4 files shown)


gusto/state.py
```diff
@@ -571,48 +571,56 @@ def _allocate_state(self):
...
+    if coords_fs.extruded:
+        cell = mesh._base_mesh.ufl_cell().cellname()
+        DG1_hori_elt = FiniteElement("DG", cell, 1, variant="equispaced")
+        DG1_vert_elt = FiniteElement("DG", interval, 1, variant="equispaced")
+        DG1_elt = TensorProductElement(DG1_hori_elt, DG1_vert_elt)
+    else:
+        cell = mesh.ufl_cell().cellname()
+        DG1_elt = FiniteElement("DG", cell, 1, variant="equispaced")
+    vec_DG1 = VectorFunctionSpace(mesh, DG1_elt)
+    coords_dg = Function(vec_DG1).interpolate(coords_orig)
```
jshipton: it'd be good to know why we didn't just do this interpolation in the first place - @dorugeber can you remember? (I think this is code you gave me...)

tommbendall:
> I can confirm that I've seen plot showing this works for cubed sphere meshes in 2D - Tom, does it actually solve our problem with 3D visualisation in Paraview?

Here is the demonstration!
cubed_sphere_3D

PR #222, opened 2020-02-12
Purge EulerPoincare from examples

tommbendall:
This pull request replaces the `EulerPoincare` option with the `VectorInvariant` option for the tests and examples to avoid future confusion.


tommbendall:
> It seems like it would make sense for `euler_poincare=False` to be the default in the Forcing class... (sorry)

I did think about doing this (and I am still happy to do it), but it means that if anyone has their own example which doesn't set the `euler_poincare` option then they will suddenly see different behaviour. If we don't mind that then I'm happy to go ahead and make this change!

jshipton:
I think it has to happen. We can highlight it in the Slack channel and send an email.

jshipton:
Has Jenkins hung? Not sure why the tests haven't gone green...

jshipton (approved):

PR #224, opened 2020-03-26
Test kernels

tommbendall:
The idea of this pull request is to separate out kernels performed by `par_loops` into their own functions. That way each kernel can have its own unit-test.

**Things this PR does**
- pulls out some kernels from `recovery.py` and puts them in `kernels.py`
- introduces new unit-tests for these kernels in a new `kernel_tests` folder
- reorganises the existing recovery tests by:
a) using more py test features
b) pulling them into their own folder
c) reducing the number of lines of code
d) sorting them in terms of geometry
e) using clearer error messages                        

NB: there are still some other kernels that need pulling out, but some of these will be removed in subsequent PRs

**Code design questions**
- Would it be better for each kernel to have its own file and for there to be a `kernels` folder?
- The way I've written it, each kernel is a class, whose `__init__` method defines the string for the kernel. Each kernel has an `apply` method which performs the `par_loop`. In the previous commit, each kernel was a function that returned the string, and the `par_loop` had to be called separately.  Which of these ways is better?


gusto/kernels.py
```diff
@@ -0,0 +1,420 @@
+"""
+This file provides kernels for par loops.
+This are contained in this file as functions so that they can be tested separately.
```
jshipton: typo - this should be these?

gusto/kernels.py
```diff
@@ -0,0 +1,420 @@
...
+from pyop2 import ON_TOP, ON_BOTTOM
+
+
+class GaussianElimination(object):
+    """
+    A kernel for performing Gaussian elimination locally in each element
+    for the BoundaryRecoverer procedure.
+
+    This takes a scalar field in DG1, the coordinates of the DoFs and the
+    "effective" coordinates of the DoFs. These "effective" coordinates
```
jshipton: If I understand correctly, this describes more input args than the `__init__` method takes.

gusto/kernels.py
tommbendall: Yes good spot!

gusto/kernels.py
tommbendall: I've updated this comment to make it clearer

PR #229, opened 2020-07-31
Reorganise recovery

tommbendall:
The goal of this pull request is to reorganise the boundary recovery procedure.

The main thing that this does is replace the calculation to find the `effective_coordinates` used by the `BoundaryRecoverer` to improve the accuracy of recovered fields at the boundary of the domain.

Previously this was done by a very specific and complicated Loopy kernel. Now it is just done by the `Averager` object! As well as removing some hideous code, this also works for calculating the `effective_coordinates` for domains with triangular cells!

How it works:
- find the coordinates at DoFs in the original space (by projecting in the coordinate field)
- as with the normal `Recoverer`, this will be interpolated up to a higher order discontinuous space
- average this field using the `Averager` to the target (partially-)continuous space
- interpolate back out to the discontinuous space used for the `BoundaryRecoverer`
- do some adjustments to ensure that the coordinates within a given cell are close together (on periodic domains averaging coordinates to a continuous space will give the wrong values at the break in the coordinates)
---
Recap of how the boundary recovery works:
- at the boundaries, we are only taking the average of the field on one side of the boundary, meaning that it does not accurately reconstruct the gradient of the field
- it turns out that these values would be accurate if the new DoFs were at the `effective_coordinates`
- to fix this, we introduced the `BoundaryRecoverer`, which uses the originally recovered values and the `effective_coordinates` to expand the field using a Taylor expansion
- this creates a linear problem to solve for the coefficients in the Taylor expansion, which we do using a Loopy kernel which performs Gaussian elimination locally in each cell
- using the `actual_coordinates` of the DoFs and the Taylor expansion, extrapolate out the field to find accurate values at the boundary
---
I've attached a pdf explaining lots of aspects of recovery
[Recovery.pdf](https://github.com/firedrakeproject/gusto/files/5018031/Recovery.pdf)

### DIFF
--- a/examples/compressible_eady.py
+++ b/examples/compressible_eady.py
@@ -142,31 +142,9 @@ theta_pert = Function(Vt).interpolate(theta_exp)
 theta0.interpolate(theta_b + theta_pert)
 
 # calculate hydrostatic Pi
-piparams = {'pc_type': 'fieldsplit',
-            'pc_fieldsplit_type': 'schur',
-            'ksp_type': 'gmres',
-            'ksp_rtol': 1.e-8,
-            'ksp_atol': 1.e-8,
-            'ksp_max_it': 100,
-            'ksp_gmres_restart': 50,
-            'pc_fieldsplit_schur_fact_type': 'FULL',
-            'pc_fieldsplit_schur_precondition': 'selfp',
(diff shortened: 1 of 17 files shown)


jshipton (approved):
This is great - thanks Tom!

PR #231, opened 2021-06-29
remove geometric bc

tommbendall:
A tiny change to update the marking of the boundaries for the recovery methods, removing the use of `"geometric"` boundary conditions which are no longer supported.


dham:
You probably want to rebase this on top of #230 in order to get it tested properly.

Issue #233, opened 2021-07-01
Failing recovery tests


tommbendall:
Some of the tests that use the recovery methods are currently failing -- for now this seems to indicate that the recovery methods are not suitable for use.

We'll do some further investigation, but in the meantime we should probably mark the tests as `xfail`.

tommbendall:
I'm closing this issue as this has now been resolved by #235 !

PR #234, opened 2021-07-01
mark the failing recovery tests with an xfail

tommbendall:
As mentioned in Issue #233, this PR is just to mark the currently failing recovery tests with `xfail` as these tests are no longer up-to-date with Firedrake and its dependencies.

We'll separately work to understand why this is but in the meantime show that we know that the tests are failing!


jshipton (approved):
Great - thanks @tommbendall !

PR #235, opened 2021-07-06
Improve kernel tests and fix broken recovery tests

tommbendall:
This PR does three things:
- rewrites the kernel tests, so that they do not assume specific a specific numbering of the DoFs in a function space. I do this by introducing a new function, `set_val_at_point`, which takes the coordinate and returns the number of the DoF at that coordinate.
- fixes the Gaussian elimination kernel, by ensuring that the array of Taylor expansion coefficients is initialised
- since all the tests are now passing, removes the `xfail` markings


jshipton (approved):
Great - thanks Tom! Good to have more robust tests and they even pass now too!

PR #236, opened 2021-09-02
Remove reference to firedrake.Expression

wence-:
We've long deprecated Expression objects, firedrakeproject/firedrake#2202 finally removes them. Cull the only reference in gusto to them.

### DIFF
--- a/gusto/recovery.py
+++ b/gusto/recovery.py
@@ -1,14 +1,16 @@
 """
 The recovery operators used for lowest-order advection schemes.
 """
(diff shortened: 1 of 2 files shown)


wence-:
This failed because since the last successful gusto run Firedrake switched to duals provided by finat and the mechanism for determining that interpolation is not possible changed. I think I have fixed that now.

wence-:
Ah, we need to wait for a firedrake docker image to be rebuilt

jshipton (approved):
Thanks Lawrence!

PR #240, opened 2022-02-14
Small fix to keep up-to-date with Firedrake

tommbendall:
This is a series of changes to keep Gusto up-to-date with Firedrake. At first I just hoped it was a simple one line change but it's ended up requiring quite a bit!

These changes are:
* updating `functionspaceimpl.WithGeometry(...)` to `functionspaceimpl.WithGeometry.create(...)` following the changes in Firedrake at https://github.com/firedrakeproject/firedrake/commit/f8f30a9051f360f96b2fdb13f78aea5113b72f11
* changing `iterate` to `iteration_region` for the parloop kernels in `kernels.py` that only run in the top / bottom layers of an extruded mesh
* changing the spaces used by `DirichletBC` so that it is a pure function space, and not a subspace of a `MixedFunctionSpace`
* changing the way function spaces are extracted from `state` for our examples. This was only necessary for those that use diffusion (because this requires BCs to be set up), but to keep the examples consistent I have changed all of those that get function spaces from `state`
* in `linear_solvers.py`, the boundary conditions are required for a mixed problem, so I have added some code to convert `state.bcs` to the appropriate boundary conditions there
* avoiding the warnings for `create_assembly_callable` being deprecated, by removing these in `preconditioners.py`

### DIFF
--- a/examples/compressible_eady.py
+++ b/examples/compressible_eady.py
@@ -101,9 +101,9 @@ rho0 = state.fields("rho")
 theta0 = state.fields("theta")
 
 # spaces
-Vu = u0.function_space()
-Vt = theta0.function_space()
-Vr = rho0.function_space()
+Vu = state.spaces("HDiv")
+Vt = state.spaces("HDiv_v")
+Vr = state.spaces("DG")
 
 # first setup the background buoyancy profile
 # z.grad(bref) = N**2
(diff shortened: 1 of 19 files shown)


jshipton (approved):
Thank you so much Tom - this all looks great!

PR #242, opened 2022-05-11
Active Tracers

tommbendall:
This PR introduces the `ActiveTracer` object to the FML branch.

The idea is to introduce a way to flexibly add active tracers (chiefly moisture) to equation sets, so that we don't require separate `Moist` versions of each equation set. The `ActiveTracer` class stores metadata describing an "active tracer" -- which interacts with the other prognostic variables and so needs including in the `MixedFunctionSpace` for the whole equation set.

Most of the work is done in `active_tracers.py` which introduces the object, and various enumerators to encode properties of the tracer being described. Some common tracer objects for moisture species are also provided. These are then added to the equation sets through routines in `equations.py`.

Along the way, I also:
- moved the `equation.field_names` to be set up in the `__init__` of the equation sets
- encounted an issue with the vector recovered transport. I've introduce a standalone test
- fixed some lint
- renamed some of the moisture variables (this isn't complete and I suggest following it up in a separate PR)
- fixed a handful of the examples that didn't work with the `recovered` settings 

----

**Issues still to solve before merging**:
- I have set the tracer transport forms to use `IntegrateByParts.TWICE` as is used in for transporting `theta`. However I'm not sure if we want to enforce that `theta` is always transported with this option. This should be addressed here or postponed to another PR.
- In the `repl` routine in `labels.py`, functions in a `MixedFunctionSpace` and a `VectorFunctionSpace` need handling separately (this caused failures when using the recovered transport scheme for the wind field). I've put in a temporary fix but there will be a better way to do this. 

----

**This should be extended in the future by**:
- adding the `active_tracers` to the other equation sets (it is currently not implemented for `Incompressible` or `ShallowWater` equations)
- this may be better facilitated introducing a new parent class, e.g. `PrognosticEquationSet`, and moving the routines to add `active_tracer` things to belong to that?
- adding support for tracer variable types that aren't mixing ratios. This might fit with a general consideration of how we create flexibility in which prognostic variables are used (and maybe even a Gibbs function approach?!)

### DIFF
--- a/examples/dry_bf_bubble.py
+++ b/examples/dry_bf_bubble.py
@@ -62,7 +62,10 @@ if diffusion:
 else:
     diffusion_options = None
 
+u_advection_option = "vector_advection_form" if recovered else "vector_invariant_form"
+
 eqns = CompressibleEulerEquations(state, "CG", degree,
+                                  u_advection_option=u_advection_option,
                                   diffusion_options=diffusion_options,
                                   no_normal_flow_bc_ids=[1, 2])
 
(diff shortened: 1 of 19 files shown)


jshipton (approved):
This looks great - thanks @tommbendall!

One thing to bear in mind for the future is that we shouldn't need to create new equations, even in the physics code (i.e. the `AdvectionEquation` for the rain mixing ratio). But we will sort this out in a future rewrite of the physics code, as discussed by email.

PR #245, opened 2022-07-04
FML unit tests

tommbendall:
This PR implements thorough unit-tests for the various aspects of FML.

It also:
- implements subtraction and division methods for `Term`s and `LabelledForm`s
- alters the `Term.has_label` routine so that it can be forced to return a tuple if a list of labels of length 1 is passed as an argument, which sometimes we might want
- Adds a `NullTerm`, allowing label_map to return an empty `LabelledForm` (see below for discussion)
- fixes remaining lint issues

**Review details**
- this is currently a draft, based on various other FML branches
- the important changes to focus on are in [gusto/fml/form_manipulation_labelling.py](https://github.com/firedrakeproject/gusto/compare/FML_transport...FML_test#diff-e5325f0db0b5dcd51bd0015356dd400f62048232731a350a2e3588bb4d678645)

**Questions**:
- do the `__rmul__` routines do what we expect them to do?
- should we allow more multiplication of `Term`s by things (e.g. functions / Forms)?
- it is not clear in my mind what we should do about the initialiser for the `label_map`. Here I have proposed a `NullTerm` (contains `None` as its `Form`) but we should think about what the best thing to do here is

### DIFF
--- a/examples/compressible_eady.py
+++ b/examples/compressible_eady.py
@@ -131,8 +131,8 @@ state.set_reference_profiles([('rho', rho_b),
 
 # Set up transport schemes
 transported_fields = [SSPRK3(state, "u"),
-                   SSPRK3(state, "rho"),
-                   SSPRK3(state, "theta")]
+                      SSPRK3(state, "rho"),
+                      SSPRK3(state, "theta")]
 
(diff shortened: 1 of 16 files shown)


tommbendall:
Merging as will be included in FML_examples

PR #246, opened 2022-07-11
fix transport velocity for equation sets

jshipton:
This fixes a bug with the transport velocity. The `Timestepper` class now assumes that the velocity is a prognostic variable and sets `uadv="prognostic"` so that the transport velocity can be correctly replaced with the first component of the `subject` in the time discretisation.


tommbendall:
Thanks for noticing it. The solution works so I'm happy to merge it on, but have a discussion next week about how whether there's anything else we want to do.

PR #247, opened 2022-07-12
FML Linearisations

tommbendall:
This organises the linearisation code, allowing linear terms to be generated from the terms in the main equation set. This:
- introduces an intermediate `PrognosticEquationSet` class, which has shared properties and methods for the main equation sets.
- adds a routine to generate linear terms from the main equations (by replacing the subject with the trial function)
- adds `Linear` equation sets, which replace all the terms with their linearisations

This means that all tests should now be passing, except for the `mountain_hydrostatic.py` example which is also now failing on master branch.

There are several messy aspects that should be followed up on:
1. How the terms to be linearised are specified. Currently this is done through a `terms_to_linearise` dictionary that is probably not easy to use in practice. Maybe we could make a function to use as a term filter instead?
2. Currently we don't specify the base state to linearise about. It would be good to make this configurable, and possibly to connect it to the reference profiles.
3. There have been some special cases to deal with transport terms, as the transport terms used in the linear solver are not simply linear versions of the terms used by transport schemes. I was thinking that we could attach the terms used by the transport scheme as a label to the main terms in the equation set, in just the same way that we attach the linear terms to the terms in the main equation set as a label.
4. Linear terms should really be generated by the `ufl.derivative` function.
5. The linearisation of the `state.perp` function will need special treatment in 2D.


jshipton (approved):
This looks wonderful Tom - thank you!

jshipton:
P.S. maybe appease lint a little before merging?

tommbendall:
Merging as will be included in FML_examples branch. Various things will be followed up on in new issues

PR #249, opened 2022-07-18
fix sponge layer by renaming pi to exner_pi when it is the Exner pres…

jshipton:
…sure


tommbendall (approved):
Looks great, thanks! This fixes the mountain_nonhydrostatic case so that the sponge layer is computed correctly

PR #250, opened 2022-07-18
Fix moist compressible Euler tests

tommbendall:
This fixes two bugs in the moist compressible Euler tests


jshipton:
Great - thanks Tom!

jshipton (approved):

PR #251, opened 2022-07-19
FML incomprehensible incompressible

tommbendall:
This fixes the equations for the incompressible Boussinesq equations


jshipton (approved):
great - thanks!

PR #253, opened 2022-07-19
Fml nell

nhartney:
This adds the RK4 time-stepping scheme and Heun's method. It:

- introduces `RK4` and `Heun` classes to `time_discretisation.py`
- removes the time-stepping options from `test_dg_transport.py` to separate the testing of the transport options and the time-stepping options
- creates`test_time_discretisation.py` (which includes tests for RK4 and Heun's method) to test time-stepping options.


tommbendall (approved):
Looks fantastic, thanks!

PR #257, opened 2022-07-19
Make heat capacities compatible

tommbendall:
This corrects `cv` in the compressible parameters to make it compatible with `Rd` and `cp`.


jshipton (approved):
it's good to be compatible!

PR #259, opened 2022-07-19
make parameters UFL constants

tommbendall:
This makes the configuation parameters be UFL Constants to ensure that forms aren't recompiled


tommbendall:
Thanks for these changes, they make sense. Presumably there might be some other places not covered by the test-suite where things might also need casting to floats, so I guess we should just keep an eye out for that happening

jshipton (approved):

PR #260, opened 2022-07-22
Fml hydrostatic

jshipton:
This implements the hydrostatic compressible Euler equations.
- This defines a `HydrostaticCompressibleEuler` equation class.
- The `h_project` function has been moved from `state` to `linear_solvers` - this will be redundant when we get the residual directly from the equation class.
- The forcing class checks for the `hydrostatic` label and adds the required forms to the explicit and implicit solvers.

To Do:
- This currently only works in a vertical slice and should be generalised to the sphere.


tommbendall (approved):
Looks great, thanks Jemma!

PR #261, opened 2022-07-22
Fml

jshipton:
This is the epic Gusto rewrite, implementing the form manipulation labelling for all equations sets. The tests and examples have all been reorganised (thanks @tommbendall !).


tommbendall (approved):
This has been through a really length review process through many different tickets so there is not very much to add here. Lots of issues and discussions have been opened to address new enhancements or problems to tackle. It's really exciting that we'll finally be using this, and well done for getting here!

PR #286, opened 2022-07-25
Rename "pi" to "exner" everywhere

### LINKED ISSUE #265: Changing "pi" to "exner"
We should change `pi` to `exner` when it refers to the Exner pressure function to avoid confusion with the number `pi`.

tommbendall:
This just renames `pi` to `exner` everywhere across Gusto when it refers to the Exner pressure, allowing us to close #265


jshipton (approved):
Very thorough - thanks!

Issue #275, opened 2022-07-22
The time step dt should be a Constant held by state


tommbendall:
`dt` should be a `Constant` held by the `state` object. Then the time step used by time discretisations, the time loop and other parts of the code should all point to this `dt`. This is really important for adaptive time stepping, where there must be only one single source of truth for `dt`.

Additionally, `t` should be updated at the end of the time step (but before outputting) and not at the start.

If no `dt` is passed to state, there should be an error message (or `dt` should be a compulsory argument!)

PR #287, opened 2022-07-25
make dt and t Constants that are owned by state

tommbendall:
This PR ensures that `state.dt` is a `Constant`, and that the rest of Gusto uses this `dt`. If we need adaptive time stepping then updating `state.dt` will ensure that the correct time step is picked up everywhere.

This addresses #275 

Also:
- the time is updated at the end of the time step (but before outputting)
- the `PrescribedTransport` time stepper updates the transporting velocity with a `Projector`
- when a `Constant` is multiplied by zero, it creates a `ufl.constantvalue.Zero` object. I have added handling of this for `LabelledForm`s, which arises in the `test_gw_incompressible.py` test

### DIFF
--- a/gusto/fml/form_manipulation_labelling.py
+++ b/gusto/fml/form_manipulation_labelling.py
@@ -114,6 +114,9 @@ class LabelledForm(object):
     def __mul__(self, other):
         if type(other) in (float, int):
             other = Constant(other)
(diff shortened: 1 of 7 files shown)


jshipton (approved):
great - thanks Tom!

Issue #270, opened 2022-07-22
Adopt new checkpointing infrastructure


tommbendall:
Gusto still uses the `DumbCheckpoint` for checkpointing -- which is now old and will be obsolete, leading to the warnings:
```
DeprecationWarning: DumbCheckpoint class will be deprecated after 01/01/2023; use CheckpointFile class instead.
```

We should update gusto to use the new `CheckpointFile` class.

PR #289, opened 2022-08-08
update checkpointing to use CheckpointFile instead of DumbCheckpoint

tommbendall:
This PR updates Gusto to use the new `CheckpointFile` instead of the `DumbCheckpoint`, and in doing so addresses issue #270


tommbendall:
Closing as this has been superseded by #355

PR #291, opened 2022-08-18
Improve Gusto test suite

tommbendall:
This PR follows up on the work done on the FML branch to tackle some of the weaker points in our test suite.

In general this means that integration-tests don't just check that something runs, they also check that it gives a sensible answer.

Other than the tests themselves, there are three main changes to highlights:
- some changes to `state` to make checkpointing work in some more flexible ways -- this may only be temporary if we reorganise state
- some changes to `state` to allow specific function spaces to be added to state -- this may only be temporary if we reorganise state
- changes to the tracer field used in the `conftest` -- I think using a Gaussian rather than sines slightly helps with spotting errors

Some of the `equations` tests use checkpointing which will alert us to if the results dramatically change.  These will need updating when we start using the new checkpointing capability.

----

Still to do: the equispaced PR to go on first


omarjamil:
Related to testing, I was looking at the `if '--running-tests' in sys.argv` condition and wondered if there was another way to do this. pytest recommends this: https://doc.pytest.org/en/latest/example/simple.html#detect-if-running-from-within-a-pytest-run

tommbendall:
> Related to testing, I was looking at the `if '--running-tests' in sys.argv` condition and wondered if there was another way to do this. pytest recommends this: https://doc.pytest.org/en/latest/example/simple.html#detect-if-running-from-within-a-pytest-run

That's a really helpful suggestion -- I think that could also tie into some thoughts we had about model wrappers (#262). I had thought that we might allow some level of command line interaction for running examples but this would definitely be an improvement

jshipton (approved):
Changes discussed in person and implemented already - thanks Tom!

PR #294, opened 2022-08-23
Tim fml notebooks

ta440:
Added a directory with four example notebooks to demonstrate the functionality of Gusto.


JDBetteridge:
Hopefully changes in https://github.com/firedrakeproject/gusto/pull/297 fix the above issues

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
```diff
@@ -0,0 +1,380 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "664c0cba",
+   "metadata": {},
+   "source": [
+    "The Williamson 2 test case solves the shallow water equations. This corresponds to the following set of momentum and continuity equations, for a velocity $\\textbf{u}$ and free surface, $\\eta$:\n",
+    "$$\\textbf{u}_t + f \\textbf{u}^{\\perp} + g \\nabla \\eta + (\\textbf{u} \\cdot \\nabla) \\textbf{u} = 0$$\n",
+    "$$\\eta_t + H(\\nabla \\cdot \\textbf{u}) + \\nabla [\\textbf{u} (\\eta - b)] = 0$$\n",
```
tommbendall, line 34: I have a very petty suggestion for how we write the shallow-water equations. I prefer to see the transport term (u.grad)u immediately after the time derivative in the momentum equation (so that you can easily see that together these are the material derivative). For the continuity equation, can we just write
`\\eta_t + \\nabla \\cdot [(\\eta - b) \\textbf{u}] = 0`?

Please feel free to ignore if you disagree!

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
```diff
@@ -0,0 +1,380 @@
...
+    "\n",
+    "The parameters we need to specify are the Coriolis force, $f$, gravitational constant, $g$, and mean depth, $H$.\n",
+    "$\\newline$\n",
+    "We can neglect boundary conditions, due to solving on a spherical domain. \\ \n",
+    "Williamson 2 is called 'Global Steady State Nonlinear Zonal Geostrophic Flow', so the solution should not change over time."
```
tommbendall, line 39: Maybe we could rewrite this as "The second Williamson test case models zonal geostrophic flow. This is a steady-state solution of the shallow-water equations, so the true solution does not change with time."

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
```diff
@@ -0,0 +1,380 @@
...
+    "from math import pi\n",
+    "import sys"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "33f71f3a",
+   "metadata": {},
+   "source": [
+    "We now define the time-step size and the duration of the simulation. We will also decide how frequently we want an output to be able to visualise in Paraview, using dumpfreq."
```
tommbendall, line 76: When we have a variable from the code embedded in normal text, we could use code formatting to highlight that it is a variable (like `dumpfreq` here)

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
```diff
@@ -0,0 +1,380 @@
...
+     "output_type": "stream",
+     "text": [
+      "firedrake:WARNING OMP_NUM_THREADS is not set or is set to a value greater than 1, we suggest setting OMP_NUM_THREADS=1 to improve performance\n"
+     ]
+    }
+   ],
+   "source": [
+    "from gusto import *\n",
+    "from firedrake import IcosahedralSphereMesh, SpatialCoordinate, as_vector\n",
+    "from math import pi\n",
```
tommbendall, line 67: I wonder if we could actually just import `pi` from `firedrake`? I think we also might not need to do `import sys`

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
```diff
@@ -0,0 +1,380 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "id": "754708dd",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "output = OutputParameters(dirname=\"sw_W2_ref%s_dt%s\" % (ref_level, dt),\n",
```
tommbendall, line 157: We actually can probably give a clearer directory name -- maybe just:
```
dirname="williamson_2_intro"
```

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
```diff
@@ -0,0 +1,380 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "9918a45e",
+   "metadata": {},
+   "source": [
+    "We are now ready to specify the initial conditions.\n",
+    "Due to our choice of function spaces for the velocity and depth, the intialisations of each variable use projection \n",
+    "and interpolation operations respectively. Williamson 2 specifies initial conditions for a 'Global Steady State Non-linear Zonal Geostrophic Zone' (do I need this detail?). \n",
```
tommbendall, line 230: I think you don't need this detail -- we've explained the initial condition at the top so I think it's OK to leave this out

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
```diff
@@ -0,0 +1,380 @@
...
+    "$$ \\textbf{u}_0 = \\frac{u_{max}}{R} [-y,x,0] $$\n",
+    "$$ h_0 = H - \\frac{\\Omega u_{max} z^2}{g R}  $$\n",
+    "We only require a two-dimensional velocity, so the vertical component is specified to be zero. (Jemma/Tom question: Why is $R/R^2$ used in the code)???"
```
tommbendall, line 233: I think we are embedded in a 3D space so we are specifying all three of the Cartesian components of the velocity. I don't think we need the `R/R^2` so maybe you can cancel one out! Presumably the was originally done to make it clear that the final part is `sin^2(lat)`?

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
```diff
@@ -0,0 +1,380 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "id": "b1642c8b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#Now, construct the time-stepper. We will firstly use a semi-implicit (?) approach.\n",
```
tommbendall, line 279: Yes it's `semi-implicit`!

tommbendall (changes requested):
I think this looks really nice and I'm really excited for us to have these in Gusto. Thanks so much for doing all of this.

I have a few pernickety comments about style of some particular comments and things we can probably omit -- if you disagree with anything please feel free to ignore it. The main thing that I think would be nice is to add lint checking, which I suspect may highlight some more issues.

I've only looked at the first notebook -- I thought I'd come back to that in case you want to make any changes having used the lint checker.

I was also wondering whether we want to include the mountain notebook? Some of the routines in there aren't that obvious so it might be quite challenging as an introduction. Maybe there's something important in that case that I'm missing though that is helpful!

My final suggestion is about the names of the notebooks. I think the first one is great, the second one could do with some capitalisation to make it look consistent (i.e. Book_2_Gravity_Wave_Vertical_Slice) and finally I think the final one might be best as Book_3_Lowest_Order_Rising_Bubble (or book 4 if you keep the mountain!)

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
jshipton: Yes, I agree with grouping the terms in that way and I also think we should try to be consistent with the names in the code - i.e. we use `D` for the prognostic variable, where `D=H+\eta-b`.

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
jshipton: yes, this is probably highlighting that the examples need a bit more tidying...

jshipton:
Thanks Tim, these are a great start to our notebook resources!

tommbendall:
> > 
> 
> I suggested the mountain as it demonstrates how to deform the mesh - so maybe not an introductory notebook but a more advanced one!

Ah OK! Maybe that can be the fourth then? And maybe it should be called Book_4_Flow_Over_A_Mountain

ta440:
Hi Tom and Jemma, thanks for the feedback! I've made the changes you've mentioned and have added the notebooks to the lint checker, so will see what that comes back with.
Let me know when you change the name of the time-stepper from 'Crank-Nicolson', as I'll need to update the notebooks to compensate.

jupyter_notebooks/Book_4_Mountain_Wave_Moving_Mesh.ipynb
```diff
@@ -0,0 +1,508 @@
+{
```
tommbendall, line 1: I'm wondering if it could be confusing to have "moving mesh" in the name of the file here since people might think that it's a time-evolving mesh. How about just "Mountain Wave?

jupyter_notebooks/Book_4_Mountain_Wave_Moving_Mesh.ipynb
```diff
@@ -0,0 +1,508 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "3ec4b5fe",
+   "metadata": {},
+   "source": [
+    "# Notebook 3: Mountain Wave Test Case"
```
tommbendall, line 8: I suppose this is notebook 4 now -- sorry!

jupyter_notebooks/Book_4_Mountain_Wave_Moving_Mesh.ipynb
```diff
@@ -0,0 +1,508 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "eb79138f",
+   "metadata": {},
+   "source": [
+    "This notebook will go through an example of a hydrostatic flow over a mountain. This test case is outlined in the paper: Melvin, T., Dubal, M., Wood, N., Staniforth, A., & Zerroukat, M. (2010). An inherently mass‐conserving iterative semi‐implicit semi‐Lagrangian discretization of the non‐hydrostatic vertical‐slice equations. Quarterly Journal of the Royal Meteorological Society: A journal of the atmospheric sciences, applied meteorology and physical oceanography, 136(648), 799-814."
```
tommbendall, line 16: Again we can probably shorten this citation and give the link:  https://doi.org/10.1002/qj.603

jupyter_notebooks/Book_4_Mountain_Wave_Moving_Mesh.ipynb
```diff
@@ -0,0 +1,508 @@
...
+    "hm = 1.\n",
+    "zs = hm*a**2/((x-xc)**2 + a**2)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2fb1fb2d",
+   "metadata": {},
+   "source": [
+    "We now invoke the smooth_z variable, I assume to smooth out the vertical profile somehow (do we want an explanation of this?? We then apply this smoothing to the defined mesh."
```
tommbendall, line 123: I actually think it's clearest if we remove the `smoother_z` flag? But we can still use the smoother z profile. The description here could then be something like: "Now we transform the mesh to follow the defined terrain. A smoothing is applied in the vertical direction."

jupyter_notebooks/Book_4_Mountain_Wave_Moving_Mesh.ipynb
```diff
@@ -0,0 +1,508 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 14,
+   "id": "7e16a3dc",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "supg = True\n",
+    "if supg:\n",
```
tommbendall, line 427: As in the second book we should probably just pick to use `supg` and get rid of the flag here

jupyter_notebooks/Book_4_Mountain_Wave_Moving_Mesh.ipynb
```diff
@@ -0,0 +1,508 @@
...
+    "# Here's what you would do:\n",
+    "# stepper.run(t=0, tmax=tmax)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "10e50cf7",
+   "metadata": {},
+   "source": [
+    "What other feature do we want to demonstrate in this notebook?"
```
tommbendall, line 483: I think that's all good!

ta440:
Hi Tom,

Sorry for the late reply - I just came back from my NZ trip.
I'll look into making those adjustments to the notebooks today. Thanks for reading through them!

Tim

.github/workflows/build.yml
```diff
@@ -58,8 +58,19 @@ jobs:
...
+      - name: Gusto jupyter_notebooks
+        run: |
+          . /home/firedrake/firedrake/bin/activate
+          export GUSTO_PARALLEL_LOG=FILE
+          python -m pytest \
+            -n 4 --dist loadscope \
+            --timeout=1200 \
+            --timeout-method=thread \
+            -o faulthandler_timeout=1260 \
+            --nbval-lax \
```
JDBetteridge, line 72:
```suggestion
            --nbval \
```
When you're happy with the state of the notebooks, you can drop the `lax` and the test will check that the output doesn't change. Use `# NBVAL_IGNORE_OUTPUT` for specific cells where this is undesirable, see [these docs](https://nbval.readthedocs.io/en/latest/#Avoid-output-comparison-for-specific-cells).

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
```diff
@@ -0,0 +1,3088 @@
...
...c9AC57X7sYy7zhsy88kdPAtIN0+bzykgBsSMUwVmd1RjPkV5J/9eRJhP6GGN3BuvBJppI9odwNxVwfxoTb8u54BBAzFVST8apOlzj55xRvQrN/04yrQ45EiproxpEwQBAy5pRhySzXNsm/P83he+SPWRSehcxQSG2ghNtBKfKiN+FAbwX3JEJxkpTru6cX77j+wVK9DX1A54f0j6PSIJjtyLDzl/ARBRO8sRvJlDl2RY2ECe19WA0ySMJYvouDDX9Ois8ciFWIztuFQkWIE6l/Fu/1x1ZkDAdGWixwcwbn6wxnJM4CpciX+Pc8T2PcaxuJqvG8/jL/mBVBkRHs+hR/+GsEDbxLc9wrRgVZMyWj6tDFmLyPWf4jeP9+AEguhcxRScvXPMOSWqmMBka59GoH2172M57V7SL0fUic9qcp/wu9CtDjHRILD7bffnvnYvc8VaBjthRhP6gVBwOl04nQ6mTt3LvF4HI/Ho9nkffrTn9Yq0K+99hrV1dUUFU393n4/oCgKX/nKV/jUpz7Fq6+++q+eThb/hsgS6CyyOAoMDAxQXl6eMagkhVQjzuHaxY1HSu8cDodxOp0sXz5RP3y4mMoSr6uri4aGBoqKili5ciXHH388Tuc4RwhBxPvW30EQEE155J58NTp7Hu6X/0DCNwQIE3TDsV71srh9zYfVxrn+ZgYe/g6WeceRe9q1xN3dCCar6qGbAZGkfCTnhMsxlS9CCnnxJXWx4eZ3VTeIFOQESiKOoDNMGCcVLmLMICMZi1SDm2P1eRrplWWZcOtOQvvfINrToEpBkprWgYe+jj63DNOsJZjKF2OqWIKhsFJrSkwkm9vGSzjGI0W0xic7TlxQPZkxzV6Oc50aZpOqVMcHWlVCPdhKrL+ZRNADUnTUB1oQEYwW9M5iDEVzMM1ahnXeekRbruZVPRVUi7a+NMKXCHjw7XkW/57nUaJBdRsmG0o0SMmVP9ESCMdDZ81B5yxG8ruQYxECtZvw7ngSOehR5T5zVlJ43leIDXcx9MQPCbfuntDgl4I5qa0PH9pOT/N2NdHRYCb39E/jXHsBAJJ3gOC+VwgdfGsCgU6EvEjJZkAlFsK28lzyP3TTqL9zvvrejPW3oCgyI1vuV73RBTUqfKxzitYXIOjIOekqPJvv0p678cYbM85fluUZN/AdK0xGoMfDYDBQXFxMcXExsizz8
```
atb1995, line 3068: It would be good to have some description of what we are seeing in the plots - and perhaps a comment on the oscillations in the depth perturbation

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
```diff
@@ -0,0 +1,3088 @@
...
+   "source": [
+    "# Shallow water: a geostrophically balanced test case"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5371527b",
+   "metadata": {},
+   "source": [
+    "This notebook provides an introduction to Gusto. We will demonstrate how to set up and run a shallow water model using the geostrophic balance example (test case 2) from [Williamson et. al. (1992)](https://doi.org/10.1016/S0021-9991(05)80016-6)."
```
atb1995, line 33: Formatting has messed the doi up on the actual notebook

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
atb1995, line 173: It would be good to have a description of the discretised equations - rather than it being a magic box. Gusto obviously does a lot of this for us, but I think it is still important to know what discretisation choices were made and why

atb1995 (changes requested):
Just a few suggested changes! Open to discussion on them :)

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
jshipton: Yes, I agree - and I will change the reference plots to be closer to what we're using here...

jshipton: I will add a reference, but I am deliberately not going in to too much detail here.

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
jshipton: I think this only a problem when viewing in the browser.

jupyter_notebooks/Book_1_Intro_Williamson_2.ipynb
jshipton: Unfortunately the best reference for this is the paper I haven't written yet... I have added the original linear shallow water paper but I'll update this with the Gusto paper when it is ready...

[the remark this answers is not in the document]
jshipton: I have changed the wording to avoid all contentious references to implicit or semi-implicit. This is in line with the level of detail I'm giving elsewhere!

PR #298, opened 2022-08-30
Use ufl derivative to generate linearisations

tommbendall:
I got it working! As the name suggests, this uses the `ufl.derivative` feature to generate the linearised terms in the equations.

The main problem I had before was that I was using `trials = TrialFunctions(W)` in the `action` call. `TrialFunctions` is a tuple, but replacing this with `trials = TrialFunction(W)` seemed to make everything happy.

In this PR I have restructured the complicated `label_map` that we use to generate the linear terms. Instead I am explicitly looping through the terms in the equation, and it might be possible to replace this later with a better label map. One challenge is that we want all the labels to be copied over from the original term to the linearisation.

I also realised that some terms didn't have a `prognostic` label. We could of course remove the `prognostic` and attempt to use the index of the test function but this is less user-friendly. Or maybe there is a halfway house in which the `prognostic` label is automatically generated from the test function?

One problem was the Coriolis force, I think because of the perp operator? Something about the linearisation or replace did not work here, so the linear shallow water test only worked when I manually added a linearisation of this term. I have only done this for the shallow-water equations and not the other equation sets.

### DIFF
--- a/gusto/equations.py
+++ b/gusto/equations.py
@@ -4,7 +4,7 @@ from firedrake import (TestFunction, Function, sin, pi, inner, dx, div, cross,
                        TrialFunction, FacetNormal, jump, avg, dS_v,
                        DirichletBC, conditional, SpatialCoordinate,
                        split, Constant, action)
-from gusto.fml.form_manipulation_labelling import Term, all_terms, LabelledForm, drop
+from gusto.fml.form_manipulation_labelling import Term, all_terms, identity, drop
 from gusto.labels import (subject, time_derivative, transport, prognostic,
                           transporting_velocity, replace_subject, linearisation,
                           name, pressure_gradient, coriolis,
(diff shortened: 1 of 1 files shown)


gusto/equations.py
```diff
@@ -276,13 +277,20 @@ def generate_linear_terms(self, residual, terms_to_linearise):
...
+            if (not t.has_label(linearisation)
+                and any(t.has_label(*terms_to_linearise[field], return_tuple=True))):
+                # Linear form is obtained by taking derivative of original term
+                linear_form = action(ufl.derivative(t.form, self.X), self.trials)
+                # Make term by adding labels from original term to form
+                raw_linear_term = Term(linear_form, t.labels)
+                # Replace the prognostic variables with their reference state
+                linear_term = replace_subject(self.X_ref)(raw_linear_term)
+                # Add the linearisation to the original term through a label
+                residual.terms[term_idx] = linearisation(t, linear_term)
```
wence-, line 293: Why can't you use `label_map` any more?

wence-: Ah, I see this is deliberate

wence-:
```python
from functools import partial

def should_linearise(term, field):
    return (
        not term.has_label(linearisation)
        and term.get(prognostic) == field
        and any(term.has_label(*terms_to_linearise[field], return_tuple=True))
   )


def linearise(term, X, X_ref, du):
    linear_term = Term(action(ufl.derivative(term.form, X), du), term.labels)
    linear_term = replace_subject(X_ref, linear_term)
    return linearisation(term, linear_term)


for field in self.field_name:
    residual = residual.label_map(
        partial(should_linearise, field=field),
        partial(linearise, self.X, self.X_ref, self.trials)
    )
```

??

tommbendall: Thanks, that looks great! I'll lift that exactly as is.

I like the `keep` suggestion too.

wence-: (Fully untested btw)

tommbendall: It all works for me, just needed to change `self.field_name` to `self.field_names` (which highlights that it is confusing to have both of these variables!)

jshipton (approved):

PR #305, opened 2022-10-27
Docstrings Part II

tommbendall:
This is the second load of docstrings, which means that we're nearly there -- we just need to do `timeloop.py`.

To check the docstring style, I used:
`pydocstyle gusto/*.py --ignore=D203,D204,D212,D213,D215,D400,D401,D404,D406,D407,D408,D409,D413,D205,D415`

This also fixes the failures in the shallow-water tests which are caused by this Firedrake issue: https://github.com/firedrakeproject/firedrake/issues/2595, by changing the mesh degree to 1. Once this is fixed we should change the mesh degree back (although probably to degree 2)


jshipton (approved):
Massive improvement - thanks Tom!

PR #306, opened 2022-10-29
Tidy timeloop

jshipton:
The main purpose of the PR is to simplify the `Timestepping` classes. This has been done by removing the `problem` argument, which means much less unpicking is necessary in the `Timestepper` class. Where the specification of the problem is more complicated, e.g. for the Strang-split, semi-implicit quasi-Newton solver, this is handled in that class's own methods. This removes the functionality that enabled the user to specify active labels to specific schemes, but in this case the user should write their own time stepping class because other decisions also need to be made (such as the order of application of the schemes and whether to update after each scheme).

Other changes:
* The order of the arguments to the time discretisation apply methods has been changed to enable multilevel methods (which require multiple inputs but still only one output).
* The `CrankNicolson` Timestepper has been renamed to `SemiImplicitQuasiNewton`.
* BDF2 is implemented (to check the multilevel functionality).
* Classes containing fields have been moved to their own file `fields.py`.

Note: docstrings still to do, but wanted to discuss first...


gusto/timeloop.py
```diff
@@ -297,33 +311,87 @@ def timestep(self):
...
+    """
+    def __init__(self, equation, scheme, state, physics_list=None,
                  prescribed_transporting_velocity=None):
+        """
+        Args:
+            equation (:class:`PrognosticEquation`): the prognostic equation
+            scheme (:class:`TimeDiscretisation`): the scheme to use to timestep
+                the prognostic equation
+            state (:class:`State`): the model's state object
+            physics_list: optional list of classes that implement `physics`
```
tommbendall, line 348: I think this should be something like:
```
physics_list (list, optional): a list of :class:`Physics` options describing physical parametrisations. Defaults to None.
```

gusto/timeloop.py
```diff
@@ -297,33 +311,87 @@ def timestep(self):
...
+                schemes
+            prescribed_transporting_velocity: (optional) expression specifying
```
tommbendall, line 350: I think this should be something like:
```
prescribed_transporting_velocity (func, optional): a function, with a single argument representing the time, that returns a :class:`ufl.Expr` for the transporting velocity. This allows the transporting velocity field to be updated with time. If `None` is provided then the equation's velocity field is not updated. Defaults to None.
```

tommbendall (changes requested):
I am really happy with all the code in this and don't actually have any comments about that.

The vast majority of the docstrings look correct, there were a few small things I did spot and have left comments about these. I think some of the code in `time_discretisation.py` still needs doing, and obviously this is a very minor point but in some places the `x_in` and `x_out` documentation needs swapping.

gusto/fields.py
```diff
@@ -0,0 +1,170 @@
...
+
+        self.add_fields(equation, levels)
+        self.previous = [getattr(self, level) for level in previous_levels]
+        self.previous.append(getattr(self, "n"))
+
+    def add_fields(self, equation, levels=None):
+        """
+        Args:
+            equation (:class:`PrognosticEquation`): an equation object.
+            levels (optional, iterable): an iterable containing the names
```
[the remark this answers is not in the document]
jshipton: I really appreciate this level of detailed scrutiny - thanks Tom!

tommbendall (approved):

PR #312, opened 2022-11-07
Improve active tracers

### LINKED ISSUE #303: Duplication in transport options
There is some duplication in the setting of transport equation options. For example:

1. There is a `NoTransport` option in the `TransportEquationType` `Enum` but there is also a `transport_flag` argument that is passed to the `ActiveTracer` class that specified whether the tracer is to be transported or not.
2. The `TracerVariableType` `Enum` probably has the information required to decide what that tracer's `TransportEquationType` should be.

tommbendall:
This removes the redundant `transport_flag` property of the `ActiveTracer`, and converts `is_moisture` into a more generic `chemical` property.

This should allow us to close #303

### DIFF
--- a/gusto/active_tracers.py
+++ b/gusto/active_tracers.py
@@ -80,8 +80,7 @@ class ActiveTracer(object):
         if self.variable_type != TracerVariableType.mixing_ratio:
             raise NotImplementedError('Only mixing ratio tracers are currently implemented')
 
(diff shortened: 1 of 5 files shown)


jshipton (dismissed):

jshipton (approved):
This is great now I've stopped breaking it! Thanks Tom!

PR #317, opened 2022-11-28
Linearisation map

tommbendall:
This PR removes the `terms_to_linearise` and replaces it with a `linearisation_map`, which is a function that determines which terms should be linearised.

This simplifies the way that terms can be linearised and makes it fit much more naturally with FML.


jshipton (approved):
much nicer! thanks!

Issue #315, opened 2022-11-14
should check that physics schemes are explicit


jshipton:
In the new physics framework we can use any time discretisation method from `time_discretisation.py` to timestep the physics terms, as long as it is an explicit method. Therefore we should check that the supplied physics schemes are explicit.

PR #319, opened 2022-11-28
New structure for microphysics schemes using FML

nhartney:
This PR updates physics routines, utilising FML to allow differentiation between the time-stepping scheme applied to physics terms and the time-stepping scheme applied to everything else. 
Changes:
* Adds `ForcedAdvectionEquation` class to `equations.py`.
* Allows for the addition of active tracers to `ShallowWaterEquations` in `equations.py`.
* Adds `InstantRain` class to `physics.py`.
* Introduces `SplitPhysicsTimestepper` to allow the use of a different time-stepping scheme for physics terms.
* Adjusts `DG1Limiter` to work with a subspace of a mixed function space.
* Addresses issue #315 by checking that the scheme specified for time-stepping physics is an instance of `ExplicitTimeDiscretisation`.


jshipton:
This all looks great to me - thanks @nhartney! @tommbendall did you want to check things through before merging?

gusto/physics.py
```diff
@@ -386,96 +450,277 @@ class Evaporation(Physics):
...
+     """
+
+    def __init__(self, equation, saturation_curve, vapour_name="water_vapour",
+                 rain_name=None, parameters=None, convective_feedback=False,
+                 set_tau_to_dt=False):
+        """
+        Args:
+            equation (:class: 'PrognosticEquationSet'): the model's equation.
+                to.
+            saturation_curve (function): the saturation function, above which
```
tommbendall, line 625: I'm not sure, but is `saturation_curve` actually a `ufl.Expr`?

integration-tests/equations/test_forced_advection.py
```diff
@@ -0,0 +1,88 @@
...
+    stepper.run(0, tmax=tmax)
+
+    error = errornorm(r_exact, state.fields("rain"))
+
+    return error
+
+
+def test_forced_advection(tmpdir):
+
+    tol = 0.003
```
tommbendall, line 85: With reduced resolution and increased time step, the test may no longer pass with this tolerance so we might need to change that too

integration-tests/equations/test_forced_advection.py
```diff
@@ -0,0 +1,88 @@
+"""
+Tests discretisations of the forced advection equation. This checks the
```
tommbendall, line 2: Maybe we could mention what the test case is? e.g. something like: "This test describes transport of water vapour which will be converted to rain if it exceeds a prescribed saturation profile. The initial condition and saturation profile are chosen to give an analytic solution. The test then checks ..."

integration-tests/physics/test_instant_rain.py
```diff
@@ -0,0 +1,102 @@
...
+    vapour0.interpolate(vapour_expr)
+
+    initial_vapour = Function(VD).interpolate(vapour_expr)
+
+    # define saturation function
+    saturation = Constant(0.5)
+
+    # define expected solutions; vapour should be equal to saturation and rain
+    # should be (initial vapour - saturation)
+    vapour_true = Function(VD).interpolate(Constant(saturation))
```
tommbendall, line 70: I don't know if we need the extra `Constant` in here?

integration-tests/physics/test_instant_rain.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def test_instant_rain_setup(tmpdir):
+    dirname = str(tmpdir)
+    state, saturation, initial_vapour, vapour_true, rain_true = run_instant_rain(dirname)
+    v = state.fields("water_vapour")
+    r = state.fields("rain")
+
+    # check that the maximum of the vapour field is equal to the saturation
+    assert v.dat.data.max() - saturation.values() < 0.001, "The maximum of the final vapour field should be equal to saturation"
```
tommbendall, line 90: I didn't know you could get the min and max of data like this! Good to know!

integration-tests/equations/test_forced_advection.py
```diff
@@ -0,0 +1,88 @@
...
+
+    return error
+
+
+def test_forced_advection(tmpdir):
+
+    tol = 0.003
+    error = run_forced_advection(tmpdir)
+    print(error)
+    assert error < tol, 'The error in the forced advection equation is greater than the permitted tolerance'
```
tommbendall, line 88: I think this is a really good test, probably better than many of the others that we have

integration-tests/physics/test_instant_rain.py
```diff
@@ -0,0 +1,102 @@
...
+    assert v.dat.data.min() - initial_vapour.dat.data.min() < 0.001, "The minimum of the vapour field should not change"
+
+    # check that the maximum of the excess vapour agrees with the maximum of the
+    # rain
+    VD = Function(v.function_space())
+    excess_vapour = Function(VD).interpolate(initial_vapour - saturation)
+    assert excess_vapour.dat.data.max() - r.dat.data.max() < 0.001
+
+    # check that the minimum of the rain is 0
+    assert r.dat.data.min() < 1e-8
```
tommbendall, line 102: I think this is another really good test

tommbendall (changes requested):
This is looking really really good, it's very exciting to have all of the physics working and I really like the two tests.

I've got a handful of very very minor comments which are mainly to do with whether we can speed up the forced advection test.

integration-tests/physics/test_instant_rain.py
jshipton: I think it is not parallel-safe so should maybe be changed...

nhartney:
Thanks so much Tom! I have hopefully implemented those changes now.

tommbendall (approved):
Thanks Nell, I think this is ready now!

PR #322, opened 2023-01-12
added functionality for replace_subject/test/trial when handling mixed spaces

JHopeCollins:
The replace_* functions in `labels.py` are missing some functionalities that are needed when handling mixed spaces

1. `replace_subject` should correctly handle the new object being a ufl.Indexed instead of a Function/tuple of Functions etc
2. `replace_test_function` and `replace_trial_function` should be able to take an `idx` argument to replace single components of mixed spaces.

### DIFF
--- a/gusto/labels.py
+++ b/gusto/labels.py
@@ -115,6 +115,12 @@ def replace_subject(new, idx=None):
             elif type(new) == ufl.algebra.Sum:
                 replace_dict[subj] = new
 
+            elif isinstance(new, ufl.indexed.Indexed):
+                if idx is None:
+                    raise ValueError('idx must be specified to replace_subject'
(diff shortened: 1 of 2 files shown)


JHopeCollins:
Some extra points:
- I added Exceptions when you should have passed an `idx` to `replace_subject` but didn't.
- When the parameter combination to the `test_replace_subject` unit test are incompatible it was previously returning `True`. pytest was complaining and saying that this would be deprecated soon, so I changed to return `None`. This could also be changed to calling `pytest.skip(<some useful message>)` from inside the test.

jshipton:
This looks great - thanks @JHopeCollins! I have made an issue to remind us to improve the test for `replace_subject`: #323

jshipton (approved):

PR #324, opened 2023-01-13
Neaten up some aspects of FML

tommbendall:
This PR tidies a couple of small parts of FML, and adds a routine to the `equation` object to allow labels to be easily applied to terms.


jshipton (approved):
Is this used anywhere yet? I can see where I would use it in future...

tommbendall:
> Is this used anywhere yet? I can see where I would use it in future...

No, I thought I would use it in the semi-implicit time stepper when we get rid of `forcing.py` but I didn't know if we ready to do that yet!

PR #345, opened 2023-01-20
add perp label to defer application of perp operator until after subj…

jshipton:
…ect is replaced as otherwise replace can no longer find the subject


jshipton:
This wasn't really a PR - I just wanted to run the tests... however, it needs me to merge from main but I would like @ta440 to be able to merge it quickly with his branches so I won't do that just yet...

Issue #272, opened 2022-07-22
The perp operator cannot easily be linearised


tommbendall:
The `perp` term cannot easily be linearised in planar 2D domains or vertical slices. This means that the Coriolis force cannot easily be included in the linearisation of equation sets.

This should be addressed, at least by adding a bespoke linearisation for the perp operator.

tommbendall:
Fixed by #346

PR #346, opened 2023-01-23
Fix 2 d perp

jshipton:
This fixes the issue with `ufl.replace` not finding the perp of a function defined on a subcomponent of a mixed function space. This affects shallow water simulations in the plane. The fix is to label (of course!) the Coriolis term with a label called `perp`, who's value is the function that defines the perp operator. This means that we can delay the application of the perp operator until after the subject of the form has been replaced. Ideally we would find a better fix for this...


gusto/equations.py
```diff
@@ -663,11 +663,15 @@ def __init__(self, domain, parameters, fexpr=None, bexpr=None,
...
-            coriolis_form = coriolis(
-                subject(prognostic(f*inner(domain.perp(u), w)*dx, "u"), self.X))
+            coriolis_form = perp(
+                coriolis(
+                    subject(prognostic(f*inner(u, w)*dx, "u"), self.X)
+                ), domain.perp)
             # Add linearisation
-            linear_coriolis = coriolis(
-                subject(prognostic(f*inner(domain.perp(u_trial), w)*dx, "u"), self.X))
+            linear_coriolis = perp(
```
tommbendall, line 671: Should we put the linearisation behind an if statement, like for the depth transport term? i.e.
```
if self.linearisation_map(coriolis_form.terms[0]):
   linear_coriolis = ...
   coriolis_form = linearisation(coriolis_form, linear_coriolis)
```
Sorry, I realise we weren't doing this previously!

tommbendall (changes requested):
Even though we'd rather not have to fix this, I think it's very neat and good news that we've understood what's going on!

tommbendall (approved):
Thanks, all looks great now. And good to know we're also completing issue #272 !

PR #347, opened 2023-01-27
Fixed Compressible Kinetic Energy Diagnostic

Witt-D:
Fixed the object creation code for the CKE diagnostic and included an example of it being used as a diagnostic in the DCMIP 3,1 compressible example.
Also tidied up some antiquated super calls in the diagnostics file.


tommbendall (approved):
Thanks Daniel, all looks good to me and everything's passing so I'm happy to merge this in.

PR #348, opened 2023-01-27
update Min -> min_value and Max -> max_value

tommbendall:
`Min` and `Max` seem to now be retired.

This PR replaces them with `min_value` and `max_value`


jshipton (approved):
Thanks @tommbendall

PR #351, opened 2023-02-06
Netcdf outputting

tommbendall:
This implements outputting of scalar fields to netCDF files.

Main points:
- a `Coordinates` object has been introduced, which stores the model's coordinates at the DoFs for different spaces.
- the `Coordinates` are created by the domain. We try to work out what the type of the domain is (e.g. `interval`, `vertical_slice`, `spherical_shell`, etc) and choose the appropriate coordinates for that domain.
- coordinates for a particular space are "registered" by the `IO` object
- the domain now also sets up some metadata relating to the size of the physical domain, which is also output to the netCDF file. This might be useful for plotting the fields.
- this all works in parallel, but the parallelisation has been handwritten. Because we only want to write to an output file on a single processor, the field data from other processors must be sent to the first processor. A follow on would be to get this working with `pnetcdf`.
- vector-valued fields should be outputted through scalar components
- there is an integration-test to check that this all works (but only in serial)


jshipton (approved):
Looks good to me - happy to merge this for now and extend to use `pnetcdf` in future...

PR #352, opened 2023-02-09
Thermal sw

nhartney:
This adds an implementation of the thermal shallow water equations to `equations.py`. The thermal shallow water framework is that of Kurganov, Liu and Zeitlin (2020) and Zerroukat and Allen (2015), with the equations written in terms of buoyancy rather than potential temperature. The discretisation results in source terms in the velocity equation and the addition of an advection equation for buoyancy to the shallow water equations set.
Main points:

- `ShallowWaterEquations` now takes as an argument a `thermal=False` flag
- an equations test called `test_thermal_sw.py`has been introduced, which is based on the thermal Williamson 2 example from Zerroukat and Allen


examples/shallow_water/thermal_williamson2.py
```diff
@@ -0,0 +1,89 @@
+from gusto import *
+from firedrake import (IcosahedralSphereMesh, SpatialCoordinate, sin, cos)
+
+# ----------------------------------------------------------------- #
```
jshipton, line 4: it would be good to include the reference and a brief explanation in the comments

gusto/equations.py
```diff
@@ -236,6 +236,8 @@ def __init__(self, field_names, domain, linearisation_map=None,
         # we should not assume this and should instead specify which variable
         # is in which space
         self.spaces = [space for space in domain.compatible_spaces]
+        if len(self.field_names) > len(self.spaces):
```
jshipton, line 239: maybe this should be a more specific check? we're assuming that just one more space is required but we don't check how much longer the list of field names is than the list of spaces... there should at least be a comment that we are assuming that any additional fields live on the DG function space

gusto/equations.py
```diff
@@ -601,6 +611,11 @@ def __init__(self, domain, parameters, fexpr=None, bexpr=None,
         u, D = split(self.X)[0:2]
         u_trial = split(self.trials)[0]
 
+        if self.thermal:
+            gamma = self.tests[2]
+            b = split(self.X)[2]
+            n = FacetNormal(domain.mesh)
```
jshipton, line 617: can we move these definitions closer to where they are used?

tommbendall (commented):
Looks great to me, I have no further suggestions!

nhartney:
Thanks both!
I think I have implemented your suggestions, Jemma, but just to confirm one thing: on the repetition of the diagnostic field checks, is the idea that none of these tests should happen here? I'm not doing any now and I've taken almost everything out of the `diagnostic_fields` list, which might be more than you meant...

gusto/equations.py
```diff
@@ -571,16 +574,23 @@ def __init__(self, domain, parameters, fexpr=None, bexpr=None,
...
         """
 
+        self.thermal = thermal
         field_names = ["u", "D"]
 
         if active_tracers is None:
             active_tracers = []
 
+        if self.thermal:
+            field_names.append("b")
```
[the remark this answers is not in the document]
jshipton: Yes, good point! Why do "topography" and "orography" begin with such unhelpful letters!

integration-tests/equations/test_thermal_sw.py
```diff
@@ -0,0 +1,127 @@
...
+    ul2 = uerr["l2"][-1]/u["l2"][0]
+    assert ul2 < 5.e-3
+
+    berr = data.groups["b_error"]
+    b = data.groups["b"]
+    bl2 = berr["l2"][-1]/b["l2"][0]
+    assert bl2 < 5.e-5
+
+
+@pytest.mark.parametrize("u_transport_option", ["vector_advection_form"])
```
jshipton, line 112: If there is only one option then we don't need to use `pytest.mark.parametrize` - you can just define `u_transport_option` as a normal variable.

jshipton:
> Thanks both!
> I think I have implemented your suggestions, Jemma, but just to confirm one thing: on the repetition of the diagnostic field checks, is the idea that none of these tests should happen here? I'm not doing any now and I've taken almost everything out of the `diagnostic_fields` list, which might be more than you meant...

You've done exactly what I meant - that's fine, all we're testing is that the flow doesn't change, we don't need to also compute the other diagnostics.

integration-tests/equations/test_thermal_sw.py
nhartney: Ah yes - thanks!

jshipton (approved):
Looks great - thanks Nell!

PR #354, opened 2023-03-07
Fix instant rain saturation

nhartney:
This allows the saturation curve for `InstantRain` to be a function of one of the equation variables, or a constant in time. If it is a function, the argument `saturation_variable` with the field name must be passed to `InstantRain`. 
It also corrects the erroneous omission of tracer transport in `ForcedAdvectionEquation`.

### DIFF
--- a/gusto/physics.py
+++ b/gusto/physics.py
@@ -633,18 +633,19 @@ class InstantRain(Physics):
     timestep dt or over a specified time interval tau.
      """
 
-    def __init__(self, equation, saturation_curve, saturation_variable=None,
+    def __init__(self, equation, saturation_curve, saturation_dependency=None,
                  vapour_name="water_vapour", rain_name=None,
                  convective_feedback=False, set_tau_to_dt=False,
                  parameters=None):
         """
         Args:
             equation (:class: 'PrognosticEquationSet'): the model's equation.
-            saturation_curve (ufl.Expr): the saturation function, above which
-                excess moisture is converted.
-            saturation_variable (str, optional): name of the variable that the
+            saturation_curve (ufl.Expr or :class: `function`): the saturation
+                function, above which excess moisture is converted. Can be
+                constant in time or a function of one of the equation fields.
+            saturation_dependency (str, optional): name of the field that the
                 saturation curve depends on, if the saturation curve changes in
-                time.
+                time. This field must be one of the equation variables.
             vapour_name (str, optional): name of the water vapour variable.
                 Defaults to "water_vapour".
             rain_name (str, optional): name of the rain variable. Defaults to
@@ -665,8 +666,7 @@ class InstantRain(Physics):
         parameters = self.parameters
         self.convective_feedback = convective_feedback
         self.set_tau_to_dt = set_tau_to_dt
-        self.saturation = saturation_curve
-        self.saturation_variable = saturation_variable
+        self.saturation_curve = saturation_curve
 
         # check for the correct fields
         assert vapour_name in equation.field_names, f"Field {vapour_name} does not exist in the equation set"
@@ -685,19 +685,20 @@ class InstantRain(Physics):
         test_v = equation.tests[self.Vv_idx]
 
         # Check if saturation is a function, and if so what of.
-        # This works if saturation is function of one of the equation variables.
-        if isinstance(self.saturation, FunctionType):
-            if self.saturation_variable in equation.field_names:
-                self.variable_idx = equation.field_names.index(
-                    self.saturation_variable)
-                Vvar_space = W.sub(self.variable_idx)
-                self.variable = Function(Vvar_space)
-                self.saturation_function = Function(Vvar_space)
+        # The dependency must be one of the equation variables.
+        if isinstance(self.saturation_curve, FunctionType):
+            if saturation_dependency in equation.field_names:
+                self.dependency_idx = equation.field_names.index(
+                    saturation_dependency)
+                dependency_space = W.sub(self.dependency_idx)
+                self.saturation_dependency = Function(dependency_space)
+                # make a Firedrake function to store the saturation function
+                self.saturation_function = Function(dependency_space)
             else:
                 raise NotImplementedError(
                     "Saturation function must be either constant in time or a function of an equation variable")
         else:
-            self.saturation_function = saturation_curve
+            self.saturation_function = self.saturation_curve
 
         # depth needed if convective feedback
         if self.convective_feedback:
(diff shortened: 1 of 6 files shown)


integration-tests/physics/test_instant_rain.py
```diff
@@ -49,7 +49,8 @@ def run_instant_rain(dirname):
 
     # Physics schemes
     # define saturation function
-    saturation = Constant(0.5)
+    VD = FunctionSpace(mesh, "DG", 1)
+    saturation = Function(VD).assign(0.5)
```
jshipton, line 53: This shouldn't be necessary - it should also work when `saturation` is not a function.

gusto/physics.py
```diff
@@ -677,6 +684,22 @@ def __init__(self, equation, saturation_curve, vapour_name="water_vapour",
         Vv = W.sub(self.Vv_idx)
         test_v = equation.tests[self.Vv_idx]
 
+        # Check if saturation is a function, and if so what of.
+        # Currently this works only when saturation is depth-dependent or not
+        # changing in time.
+        # TODO: allow saturation to be dependent on other variables
+        if isinstance(self.saturation, FunctionType):
+            if self.saturation_variable == "depth":
```
jshipton, line 692: I would prefer to use the field name rather than "depth" - then you can assert that the string is in the list of field names.

gusto/physics.py
```diff
@@ -677,6 +684,22 @@ def __init__(self, equation, saturation_curve, vapour_name="water_vapour",
...
+                self.variable_idx = equation.field_names.index("D")
+                Vvar_space = W.sub(self.variable_idx)
+                self.variable = Function(Vvar_space)
+                self.sat_func = Function(Vvar_space)
```
jshipton, line 696: would prefer not to abbreviate too much - so `saturation_function` instead of `sat_func`

integration-tests/physics/test_instant_rain.py
jshipton: That depends on what the error was...

jshipton (changes requested):
1. Some tidying and renaming required: saturation curve can be an expression or a function - docstring needs to be updated, `variable` should be renamed, don't store things in `self` that aren't used.
2. Test should check both cases for saturation curve.

nhartney:
Thanks Jemma!
In terms of testing, the moist convective test based on Williamson 5 (which isn't working yet but hopefully will be soon!) will use a saturation function that varies with height. The forced advection test uses `InstantRain`and has a space-dependent saturation function, and this instant rain test has a saturation function that is a `Constant`. In other words, between those two and the new moist convective test we will be testing three types of saturation curve. Do you think, then, that it is alright to have this test test only the `Constant` case? I think including a height-varying saturation curve here would just be a repeat of the moist convective test.

gusto/physics.py
```diff
@@ -633,18 +633,19 @@ class InstantRain(Physics):
...
                  vapour_name="water_vapour", rain_name=None,
                  convective_feedback=False, set_tau_to_dt=False,
                  parameters=None):
         """
         Args:
             equation (:class: 'PrognosticEquationSet'): the model's equation.
-            saturation_curve (ufl.Expr): the saturation function, above which
-                excess moisture is converted.
-            saturation_variable (str, optional): name of the variable that the
+            saturation_curve (ufl.Expr or :class: `function`): the saturation
```
jshipton, line 643: Should also note that this argument is optional

gusto/physics.py
```diff
@@ -633,18 +633,19 @@ class InstantRain(Physics):
...
+                function, above which excess moisture is converted. Can be
```
jshipton, line 644: is it possible to rephrase without using the word "function" or "expression" - just to avoid confusion with the type of the argument...

gusto/physics.py
```diff
@@ -633,18 +633,19 @@ class InstantRain(Physics):
...
+                constant in time or a function of one of the equation fields.
+            saturation_dependency (str, optional): name of the field that the
                 saturation curve depends on, if the saturation curve changes in
-                time.
+                time. This field must be one of the equation variables.
```
jshipton, line 648: "prognostic fields" instead of "variables"

gusto/physics.py
```diff
@@ -685,19 +685,20 @@ def __init__(self, equation, saturation_curve, saturation_variable=None,
...
+                dependency_space = W.sub(self.dependency_idx)
+                self.saturation_dependency = Function(dependency_space)
+                # make a Firedrake function to store the saturation function
+                self.saturation_function = Function(dependency_space)
             else:
                 raise NotImplementedError(
                     "Saturation function must be either constant in time or a function of an equation variable")
         else:
-            self.saturation_function = saturation_curve
+            self.saturation_function = self.saturation_curve
```
jshipton, line 701: is `self.saturation_curve` ever used? I'm not sure it needs to be stored in `self`.

jshipton (commented):
Some very minor comments on doc strings.

I think I'm still a bit confused by the logic... is it nicer to get rid of `self.saturation_function` and just have `self.saturation_curve`?

jshipton:
> Thanks Jemma!
> In terms of testing, the moist convective test based on Williamson 5 (which isn't working yet but hopefully will be soon!) will use a saturation function that varies with height. The forced advection test uses `InstantRain`and has a space-dependent saturation function, and this instant rain test has a saturation function that is a `Constant`. In other words, between those two and the new moist convective test we will be testing three types of saturation curve. Do you think, then, that it is alright to have this test test only the `Constant` case? I think including a height-varying saturation curve here would just be a repeat of the moist convective test.

Yes, I think you're right.

gusto/physics.py
```diff
@@ -632,14 +633,21 @@ class InstantRain(Physics):
...
+                 time_varying_saturation=False,
+                 vapour_name="water_vapour", rain_name=None,
+                 convective_feedback=False, set_tau_to_dt=False,
                  parameters=None):
         """
         Args:
             equation (:class: 'PrognosticEquationSet'): the model's equation.
-            saturation_curve (ufl.Expr): the saturation function, above which
-                excess moisture is converted.
+            saturation_curve (ufl.Expr or :class: `function`, optional): the
```
jshipton, line 644: this isn't optional!

gusto/physics.py
```diff
@@ -696,6 +705,16 @@ def __init__(self, equation, saturation_curve, vapour_name="water_vapour",
...
 
+        if self.time_varying_saturation:
+            if isinstance(saturation_curve, FunctionType):
+                self.saturation_computation = saturation_curve
+                self.saturation_curve = Function(Vv)
+            else:
+                raise NotImplementedError(
+                    "If the saturation curve is varying in time then currently it should depend on a prognostic field")
+        else:
+            self.saturation_curve = saturation_curve
```
jshipton, line 716: you need to assert that the saturation_curve is a UFL expression

jshipton (approved):
looks much tidier now - thanks!

PR #355, opened 2023-03-27
Update checkpointing

tommbendall:
This updates the checkpointing to:
- pick up the new `CheckpointFile` capability
- checkpoint initial and reference profiles, allowing diagnostics using these to be repeatable
- checkpoint multiple time steps, when using a multi-time level scheme
- it revamps the checkpointing test
- for now, both checkpointing methods are maintained as the new checkpointing doesn't work with HDiv fields on a 2D sphere (see https://github.com/firedrakeproject/firedrake/discussions/2843)


jshipton (approved):
Great - thanks Tom!

PR #356, opened 2023-04-11
Some general spherical meshes

tommbendall:
This adds versions of the icosahedral and cubed sphere meshes in which the number of cells per panel can be chosen much more generally, rather than being fixed to be a power of 2.


colinjcotter:
can we push these upstream to Firedrake?

tommbendall:
> can we push these upstream to Firedrake?

Thanks Colin, good idea! It will need some neatening up for that but I'm hoping these meshes can lurk in Gusto until then

jshipton (approved):
This will be really useful - thanks Tom!

jshipton (approved):

Issue #307, opened 2022-10-31
Need to change shallow-water examples to use degree=2


tommbendall:
We currently can't run spherical tests which use meshes with degree > 1 without throwing an error.

Once https://github.com/firedrakeproject/firedrake/issues/2595 is fixed we should change the shallow-water examples back to using degree = 2

PR #359, opened 2023-04-11
Make outward normal be a field store domain

tommbendall:
This is a work-around for issue #307, and makes the `outward_normal` to be a field stored in the `domain`


jshipton (approved):
Great - thanks so much @tommbendall !

PR #325, opened 2023-01-13
WIP: Allow indexing in `replace_test_function` and `replace_trial_function` maps.

JHopeCollins:
This will allow passing an `idx` argument to `replace_test_function` and `replace_trial_function` in the same way as `replace_subject`, which will help when dealing with mixed spaces.

PR #362, opened 2023-04-14
Another perp fix

jshipton:
This fixes the issue that ufl.replace doesn't replace perp(u) on the plane (because it doesn't see the relationship between [u0, u1] and [-u1, u0]). The fix is to label the terms for which this is a problem and then the Gusto replace functions pass perp(u) to ufl.replace.

This includes a test comparing to benchmarked known good values.

This PR also includes the edits introduced in #325. That PR refactors the logic in `replace_{test,trial,subject}` for building the dictionary to pass to `ufl.replace`. This simplifies the logic a bit (most type checking is deferred to `ufl` instead of being carried out by gusto), and allows it to be share between the three functions. It also adds the ability to pass an index to `replace_{test,trial}` when the new and/or old value is from a mixed space, which was previously only possible with `replace_subject`.


gusto/labels.py
```diff
@@ -65,14 +130,34 @@ def repl(t):
...
+        old_trial = t.form.arguments()[1]
+        replace_dict = _replace_dict(old_trial, new_trial, idx, 'trial')
+
+        try:
+            new_form = ufl.replace(t.form, replace_dict)
+        except Exception as err:
+            error_message = f"{type(err)} raised by ufl.replace when trying to" \
+                            + f" replace_trial_function with {new_trial}"
+            raise type(err)(error_message) from err
+
```
tommbendall, line 142: I suggest putting a comment here about why there is special handling of the `perp` -- maybe just copy the comment from `replace_subject`?

unit-tests/fml_tests/test_replace_perp.py
```diff
@@ -0,0 +1,48 @@
...
+    L = form.label_map(all_terms, replace_subject(U1, 0))
+    U2 = Function(W)
+    solve(a == L.form, U2)
+
+    u2, _ = U2.split()
+    U3 = Function(W)
+    u3, _ = U3.split()
+    u3.interpolate(as_vector([-2, 1]))
+
+    assert errornorm(u2, u3) < 1e-14
```
tommbendall, line 48: Test looks great -- I'm convinced that this is working!

gusto/labels.py
```diff
@@ -30,14 +87,22 @@ def repl(t):
...
+        old_test = t.form.arguments()[0]
+        replace_dict = _replace_dict(old_test, new_test, idx, 'test')
+
+        try:
+            new_form = ufl.replace(t.form, replace_dict)
+        except Exception as err:
+            error_message = f"{type(err)} raised by ufl.replace when trying to" \
+                            + f" replace_test_function with {new_test}"
+            raise type(err)(error_message) from err
+
```
tommbendall, line 99: Are we confident that we'll never have a perp on a test function?

gusto/labels.py
jshipton: done! in fact, the comment in `replace_subject` also needed updating.

gusto/labels.py
jshipton: nope! but this would require some extra thought so I have made an issue: #364

jshipton:
@JHopeCollins are you able to edit the PR description to include a bit about what changed in your branch?

JHopeCollins:
> @JHopeCollins are you able to edit the PR description to include a bit about what changed in your branch?

I've added a description of those additions and a link to my original PR.

tommbendall (approved):
Thanks, this also looks good now!

PR #373, opened 2023-06-12
JDBetteridge/deadlocks

JDBetteridge:
Fixes some nasty deadlocks found when running
```
examples/shallow_water/williamson_2.py
```

The workaround in `gusto/io.py` needs a proper fix. Using Python's `Set()` in parallel is not a good idea since there is no guaranteed order (see comment in file).

### DIFF
--- a/gusto/domain.py
+++ b/gusto/domain.py
@@ -165,7 +165,7 @@ class Domain(object):
                 else:
                     metadata_value = None
                 for procid in range(1, comm_size):
-                    my_tag = comm_size*j + my_rank
+                    my_tag = comm_size*j + procid
                     comm.send((metadata_key, metadata_value), dest=procid, tag=my_tag)
         else:
             # Need to receive information and store in metadata
--- a/gusto/io.py
+++ b/gusto/io.py
@@ -310,7 +310,10 @@ class IO(object):
 
         if self.output.dump_vtus or self.output.dump_nc:
             # make list of fields to dump
-            self.to_dump = [f for f in state_fields.fields if f.name() in state_fields.to_dump]
+            # FIXME: This is a hack, it shouldn't be necessary to sort this list,
+            # It shouldn't be possible to create the list in different orders on different ranks!!!
+            # Problem: `state_fields.to_dump` is a set()! No guaranteed order
+            self.to_dump = sorted([f for f in state_fields.fields if f.name() in state_fields.to_dump], key=lambda x: x.name())
 
         # make dump counter
         self.dumpcount = itertools.count()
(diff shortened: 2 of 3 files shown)


gusto/io.py
```diff
@@ -310,7 +310,10 @@ def setup_dump(self, state_fields, t, pick_up=False):
 
         if self.output.dump_vtus or self.output.dump_nc:
             # make list of fields to dump
-            self.to_dump = [f for f in state_fields.fields if f.name() in state_fields.to_dump]
+            # FIXME: This is a hack, it shouldn't be necessary to sort this list,
```
JDBetteridge, line 313: Please don't commit this!

gusto/io.py
```diff
@@ -310,7 +310,10 @@ def setup_dump(self, state_fields, t, pick_up=False):
...
+            # It shouldn't be possible to create the list in different orders on different ranks!!!
+            # Problem: `state_fields.to_dump` is a set()! No guaranteed order
+            self.to_dump = sorted([f for f in state_fields.fields if f.name() in state_fields.to_dump], key=lambda x: x.name())
```
tommbendall, line 316: Can we resolve this by changing `self.to_dump` where it is created? I think this is line 111 of `fields.py`:
```
class StateFields(Fields):
    ...
    def __init__(self, prognostic_fields, prescribed_fields, *fields_to_dump):
        ...
        self.fields = []
        output_specified = len(fields_to_dump) > 0
        self.to_dump = set((fields_to_dump))
        self.to_pick_up = set(())
```
I don't actually see why this (or `state_fields.to_pick_up` on the line below_ needs to be a set, so can we just change them both to be lists?

gusto/domain.py
```diff
@@ -165,7 +165,7 @@ def __init__(self, mesh, dt, family, degree=None,
                 else:
                     metadata_value = None
                 for procid in range(1, comm_size):
-                    my_tag = comm_size*j + my_rank
+                    my_tag = comm_size*j + procid
```
tommbendall, line 168: Thanks very much for spotting this!! I've checked the other times that we manually send/receive and I think this is the only one that was wrong.

JDBetteridge: Should they be updated to gathers/scatters rather than manual send-receives? It feels like that is what is supposed to be happening here.

gusto/io.py
JDBetteridge: So the only reason I can think for these being `set()`s rather than `list()`s is to avoid duplication. It might be that some fields are added more than once and duplicates need to be removed. If that's the case you need some sort of sorted set that is consistent over all MPI ranks.

gusto/domain.py
tommbendall: Yes that almost certainly would be better. In the long run we should use pnetcdf and hopefully this should be handled for us ... I'm happy to follow up in a separate PR and change these

JDBetteridge: Feel free to keep adding commits to this PR if you like, I don't think I'll be making any new changes.

jshipton (approved):
Fantastic! Thanks for spotting these issues @JDBetteridge  and thanks for the fixes @tommbendall! Merging despite issues #377 and #378 because running in parallel is becoming very necessary.

PR #379, opened 2023-06-19
Actually get netcdf output working in parallel

tommbendall:
This addresses some issues with our netCDF outputting by:
- fixing another small bug related to using a `set` instead of a `sorted` list for the space names used in outputting
- allowing the model to run when there are diagnostic fields that can't be outputted (e.g. the HDiv wind field or fields in a `VectorFunctionSpace`. Instead of throwing an error with these fields, a warning is printed


jshipton (approved):
Thanks @tommbendall !

PR #371, opened 2023-06-01
Change default checkpointing

tommbendall:
This changes the default checkpointing to use the new `CheckpointFile` style instead of the `DumbCheckpoint`.

I have:
- changed the name of our checkpointing method to make clearer what they are doing ('old' -> 'dumbcheckpoint', 'new' -> 'checkpointfile')
- changed the default checkpointing method to be 'checkpointfile'
- updated the data files used for KGO checking the equation sets, and changed these tests to be appropriate for the `checkpointfile` method

PR #380, opened 2023-06-19
Use new checkpointing for the mountain examples

tommbendall:
Our two outstanding examples using the `DumbCheckpoint` method of checkpointing were the mountain examples -- as these threw an error (described in PR #371) when we tried to use `CheckpointFile`.

The fix suggested by @ksagiyam is to ensure that the mountain meshes are told that their base mesh is the original 1D base mesh.

This PR makes that change, changes the checkpointing for these two examples and also neatens up a bit of the code at the top of these examples.

### DIFF
--- a/examples/compressible/mountain_hydrostatic.py
+++ b/examples/compressible/mountain_hydrostatic.py
@@ -32,25 +32,26 @@ else:
 # ---------------------------------------------------------------------------- #
 
 # Domain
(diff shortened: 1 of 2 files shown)


jshipton (approved):
Thanks @tommbendall and @ksagiyam - I have made an issue #381 to remind us to check if this gets fixed somewhere within Firedrake and then we can remove those couple of lines.

Issue #332, opened 2023-01-20
Linear transport forms need fixing


tommbendall:
Both the linear transport forms look a bit odd -- the `advective` one seems to only do the vertical part of the transport, while the `continuity` form has an option for facet terms -- when would we use this?

tommbendall:
Fixed by PR #382

PR #382, opened 2023-06-23
Introduce wrappers and separate transport from time discretisations

tommbendall:
This PR:
- introduces the `Wrapper` object for time discretisations
- introduces a `SpatialMethod` object to describe specific spatial discretisations, similar to physics schemes.
- reorganises code relating to these things

This allows a bunch of specific code to be pulled out from the time discretisations, and gives us new ways of combining and specifying our different options.

**Breakdown of changes**
1. A new `Wrapper` object is introduced in `wrappers.py`, which has pre- and post- apply methods. There are three applications of this:
- `EmbeddedDGWrapper`
- `RecoveryWrapper`
- `SUPGWrapper` (which does nothing in pre- and post- apply methods, but does introduce a different test function to be used)
I have tried to keep all of this code the same, but we should instantly gain the functionality to apply these wrappers to physics and diffusion terms.
2. `TimeDiscretisation` objects can now set up a `Wrapper` object when they are passed wrapper options. All of the specific wrapper code is now completely pulled out of `time_discretisation.py`.
3. A new `SpatialMethod` object is defined in `spatial_methods.py`, which is the base object for specifying a spatial discretisation of some term. Why bother to introduce this? It allows us to separate the generic term in an equation [e.g. `test*dot(u,grad(q))*dx`], from its discretisation [e.g. an upwind discretisation]. When the `stepper` sets up all of the methods/schemes, it will replace the term in the equation with the discretisation in the `SpatialMethod`.
4. Two general child classes of `SpatialMethod` have been introduced: `TransportMethod` in `transport_methods.py` and `DiffusionMethod` in `diffusion_methods.py`. These are very simple and pretty much just call the methods of `SpatialMethod` but with the specific `transport`/`diffusion` labels. However the `TransportMethod` requires more care as it needs to ensure that the `transporting_velocity` is replaced correctly.
5. `diffusion.py` has been removed, and code for the `InteriorPenaltyDiffusion` method has been moved to `diffusion_methods.py`
6. Two types of `TransportMethod` have been implemented: `DGUpwind` (which is basically our work-horse at the moment) and `DefaultTransport` (which calls the basic form originally set up in the equation -- I'm open to better names for this!). The forms used in `DGUpwind` are stored in `transport_methods.py`
7. NB: SUPG transport much of the same code as `DGUpwind`, so this has been implemented as a `Wrapper` (to ensure that the test function in the time discretisation gets replaced). The appropriate `IntegrateByParts` option should be passed to the `DGUpwind` object.
8. `transport_forms.py` has been renamed as `common_forms.py`, and the intention is that this should contain forms that are used by `PrognosticEquation`s *and not* `SpatialMethod`s. The basic transport and diffusion forms have been added to this.
9. I haven't been able to remove the linearisations of the transport forms, which was my hope. I think this will require some special care with the `transporting_velocity` label. I have however fixed the `linear_advection_form` (issue #332).
10. I have made some changes to the `circulation` form of the vector transport equation. I was a bit confused by what was happening before, and hopefully this clears things up:
- `"vector_invariant"`: (u.∇)u = (∇×u)×u + (1/2)∇(u.u), implemented as a *single* transport term, with separate transported and transporting velocities
- `"circulation"`: (u.∇)u = (∇×u)×u + (1/2)∇(u.u), implemented as  *two* terms, a transport term (∇×u)×u and a K.E. term (1/2)∇(u.u) [which would now be evaluated as a forcing rather than as a transport term]
11. All integration-tests and examples with transport or diffusion terms now need some extra code to specify which transport method is needed. This pretty much means just adding `transport_methods = [DGUpwind(eqn, variable_name), ...]` to all of these, and ensuring that it is passed to the time stepper.


jshipton (approved):
Fantastic - thanks @tommbendall

PR #383, opened 2023-07-21
New solver options for backwards euler - otherwise it fails for shall…

atb1995:
When using BackwardEuler as the main time stepper for the shallow water test cases I noticed the solver failed in the first time step. If solver settings are changed to the same as ThetaMethod then this fixes the failure. This is just a quick fix at the moment, but in the long term we should probably investigate the best solver settings for each time stepping method.


jshipton (approved):
thanks Alex!

PR #384, opened 2023-07-22
Only call mesh.init_cell_orientations if it hasn't been called before

JHopeCollins:
`Domain` assumes that if the mesh is on the sphere then it needs to call `mesh.init_cell_orientations`. However, if the cells orientations have already been initialised before the mesh is passed to `Domain` then Firedrake will raise an error.

This PR just adds a check before calling `init_cell_orientations` to avoid this situation.


tommbendall (approved):
Thanks Josh! This seems very sensible

PR #390, opened 2023-07-24
Moist thermal sw

nhartney:
This PR adds the capability for the moist thermal shallow water equations. It:

- Adds a new physics class, `SWSaturationAdjustment`, to do condensation and evaporation between two moist species in the shallow water equations. This class takes `saturation_curve` as an argument, and where the water vapour exceeds the saturation curve a specified fraction of the excess vapour is converted to cloud. Conversely, in a sub-saturated atmosphere where cloud is present, a specified fraction of the cloud will evaporate. If `thermal_feedback` is true then these processes affect the buoyancy equation - evaporation should cause an increase in buoyancy and condensation a decrease.
- Includes an integration test for `SWSaturationAdjustment` which tests both the evaporation and condensation ability by running two different scenarios.
- Adds a `moist_thermal_williamson5` example to the `shallow_water` directory. This is based on test 2 in Zerroukat and Allen (2015), which is a version of test 5 from Williamson et al (flow over a mountain) with additional thermodynamic variables. 
- Renames the `gamma` parameter in `InstantRain` to `beta1` to make clearer the distinction between convective feedback (`beta1`) and thermal feedback (`beta2`).  Adds a `gamma_r` parameter to `InstantRain` to allow the conversion of a specified fraction of vapour above the saturation threshold to rain. 
- Corrects a mistake in the specification of the initial conditions for the thermal Williamson 2 example.


tommbendall (commented):
Thanks for opening this! Very exciting to have this on main.

In terms of which examples to have, I think it would be lovely to have just a `moist_thermal_williamson_5.py` to ensure that we have the moist-thermal combination tested somewhere

gusto/physics.py
```diff
@@ -765,3 +767,201 @@ def evaluate(self, x_in, dt):
...
+            saturation_curve (ufl expression or :class: `function`): the
+                curve which dictates when phase changes occur. In a saturated
+                atmosphere vapour above the saturation curve becomes cloud,
+                and if the atmosphere is sub-saturated and there is cloud
+                present cloud will become vapour until the saturation curve is
+                reached. The saturation curve is either prescribed or
+                dependent on a prognostic field.
+            time_varying_saturation (bool, optional): set this to True if the
+                saturation curve is changing in time. Defaults to False.
+            L (float, optional): The air expansion factor multiplied by the
```
tommbendall, line 799: Should we call this `L_v` instead of `L`?

gusto/physics.py
```diff
@@ -765,3 +767,201 @@ def evaluate(self, x_in, dt):
...
+                latent heat due to phase change divided by the specific heat
+                capacity. For the atmosphere we take L to be 10, following A.2
+                in Zerroukat and Allen (2015). Defaults to None but must be
```
tommbendall, line 802: Just to check, should this be (2015) or (2020) as the comment above is their paper from (2020)?

gusto/physics.py
```diff
@@ -765,3 +767,201 @@ def evaluate(self, x_in, dt):
...
+                fraction avoids a two-timestep oscillation between vapour and
+                cloud when saturation is tempertature/height-dependent.
+            time_varying_gamma_v (bool, optional): set this to True
+                if the fraction of moist species converted changes in time
+                (if gamma_v is temperature/height-dependent).
+            tau (float, optional): Timescale for condensation and evaporation.
+                Defaults to None, in which case the timestep dt is used.
+            parameters (:class:`Configuration`, optional): parameters containing
+                the values of constants. Defaults to None, in which case the
+                parameters are obtained from the equation.
```
tommbendall, line 833: This is very nicely commented, thanks!

integration-tests/physics/test_sw_condensation.py
```diff
@@ -0,0 +1,155 @@
...
+    # Check that total moisture conserved
+    filename = path.join(dirname, "sw_cond_evap/diagnostics.nc")
+    data = Dataset(filename, "r")
+    water = data.groups["water_vapour_plus_cloud_water"]
+    total = water.variables["total"]
+    water_t_0 = total[0]
+    water_t_T = total[-1]
+
+    assert abs(water_t_0 - water_t_T) / water_t_0 < 1e-12, \
+        f'Total amount of water should be conserved by {process}'
```
tommbendall, line 155: Test looks really good! My only question is if we should called it `test_sw_saturation_adjustment.py` instead? (And also rename `test_condensation.py` as `test_saturation_adjustment.py`?

tommbendall (changes requested):
Generally looks really good, thanks so much for adding this!

gusto/physics.py
nhartney: Yes, I think that makes more sense.

gusto/physics.py
nhartney: Should be 2015 - thanks!

integration-tests/physics/test_sw_condensation.py
nhartney: Yes, I think that's a better name. I've changed the name of both tests.

tommbendall (approved):
Brilliant, thanks Nell!

PR #391, opened 2023-07-24
Numerical Integrator added to SRC  along with the __init__() file

Witt-D:
Added a numerical integrator to Gusto SRC along with a corrosponding initialisation call into the __innit__() file

### DIFF
--- a/gusto/__init__.py
+++ b/gusto/__init__.py
@@ -13,6 +13,7 @@ from gusto.labels import *                           # noqa
 from gusto.limiters import *                         # noqa
 from gusto.linear_solvers import *                   # noqa
 from gusto.meshes import *                           # noqa
+from gusto.numerical_integrator import *             # noqa
 from gusto.physics import *                          # noqa
 from gusto.preconditioners import *                  # noqa
 from gusto.recovery import *                         # noqa
@@ -21,4 +22,3 @@ from gusto.time_discretisation import *              # noqa
 from gusto.timeloop import *                         # noqa
 from gusto.transport_methods import *                # noqa
 from gusto.wrappers import *                         # noqa
-from gusto.numerical_integrator import *             # noqa 
--- a/gusto/__init__.py
+++ b/gusto/__init__.py
@@ -21,3 +21,4 @@ from gusto.time_discretisation import *              # noqa
 from gusto.timeloop import *                         # noqa
 from gusto.transport_methods import *                # noqa
 from gusto.wrappers import *                         # noqa
+from gusto.numerical_integrator import *             # noqa 
--- a/gusto/numerical_integrator.py
+++ b/gusto/numerical_integrator.py
@@ -1,0 +1,56 @@
+import numpy as np
+
+
+class NumericalIntegral(object):
+    """
+    A class for numerically evaluating and tabulating some 1D integral.
+    :arg lower_bound: lower bound of integral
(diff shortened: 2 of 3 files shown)


gusto/numerical_integrator.py
```diff
@@ -0,0 +1,56 @@
+import numpy as np
+
+
+class NumericalIntegral(object):
+    """
+    A class for numerically evaluating and tabulating some 1D integral.
+    :arg lower_bound: lower bound of integral
```
tommbendall, line 7: These docstrings don't match the style now used through the rest of Gusto -- would you be able to be able to update this?

gusto/numerical_integrator.py
```diff
@@ -0,0 +1,56 @@
...
+        self.x_double = np.linspace(lower_bound, upper_bound, 2*num_points-1)
+        self.lower_bound = lower_bound
+        self.upper_bound = upper_bound
+        self.num_points = num_points
+        self.tabulated = False
+
+    def tabulate(self, expression):
+        """
+        Tabulate some integral expression using Simpson's rule.
+        :arg expression: a function representing the integrand to be evaluated.
```
tommbendall, line 25: Same with these docstrings

tommbendall (changes requested):

tommbendall (changes requested):
Thanks for doing this! The main thing that we need is a unit-test for this, just to check some analytic integral

unit-tests/test_numerical_integrator.py
```diff
@@ -0,0 +1,34 @@
...
+
+@pytest.mark.parametrize("integrand_name", ["quadratic", "sine"])
+def test_numerical_integrator(integrand_name):
+    if integrand_name == "quadratic":
+        integrand = quadratic
+        upperbound = 3
+        answer = 9
+    elif integrand_name == "sine":
+        integrand = sine
+        upperbound = 2*pi
```
tommbendall, line 25: Could you integrate this to `pi` instead of `2*pi` so that the answer isn't 0? Just in case the whole thing is doing nothing we would get 0 as the answer so this makes the test extra-safe

gusto/numerical_integrator.py
```diff
@@ -0,0 +1,60 @@
...
+            self.cumulative[i+1] = self.cumulative[i] + self.interval_areas[i]
+
+        self.tabulated = True
+
+    def evaluate_at(self, points):
+        """
+        Evaluates the integral at some point using linear interpolation.
+        Args:
+            points (float or iter) the point value, or array of point values to
+                evaluate the integral at.
```
tommbendall, line 53: This docstring should also have a `Returns:` statement, as the function routines something

gusto/__init__.py
```diff
@@ -21,3 +21,4 @@
 from gusto.timeloop import *                         # noqa
 from gusto.transport_methods import *                # noqa
 from gusto.wrappers import *                         # noqa
+from gusto.numerical_integrator import *             # noqa
```
tommbendall, line 24: These imports are ordered alphabetically so could you put it in its correct place?

tommbendall (changes requested):
Looking really good and it's very nearly ready to go on. Just a few tiny things left to sort. Thanks for doing this!

tommbendall (approved):
Looks great now, thanks for taking this on!

PR #392, opened 2023-07-24
Move around parts of FML

tommbendall:
This:
- renames `form_manipulation_labelling` as `form_manipulation_language`
- moves the replacement routines from `labels.py` to `fml.replacement.py`
- puts two special labels (`subject` and  `name`) into FML
- introduces Gusto's `Perp` operator to give us properly working replacement using `perp`. This is also a PR in UFL: https://github.com/FEniCS/ufl/pull/184, so once that is in then we can remove our `perp` and our monkey patch from the __init__
- removes some type checking from FML
- overhauls the replacement, allowing new and old functions to be indexed separately
- makes a couple of changes to the `NullTerm`
- adds a workflow demo test


jshipton (dismissed):
This looks good to go!

jshipton (approved):

PR #395, opened 2023-07-25
Split apart the index argument to the `replace_subject,trial,test` functions into `idx_new` and `idx_out`

JHopeCollins:
The `replace_*` functions previously only accepted a single index argument. If only one of the new/old things were indexable (i.e. a mixed space or a tuple) then the index was used for that one. If both were indexable then the index was used for both. If neither was indexable then it was silently ignored. This is error prone because the index argument can mean different things depending on the other arguments, and it is silently ignored rather than warning/erroring when it isn't used.

This PR changes the `replace_*` functions to take separate `old_idx` and `new_idx` for indexing the thing being replaced (old) or the thing replacing (new). They will also error if an index is given but not used.


gusto/fml/replacement.py
```diff
@@ -13,67 +13,132 @@
...
+
+    if indexable_new and not indexable_old:
+        if new_idx is None:
+            raise ValueError(f"new_idx must be specified to replace_{replace_type} when new"
+                             + f" {replace_type} of type {new} is mixed or indexable and"
+                             + f" old {replace_type} of type {old} is not mixed.")
+
+    if indexable_old and indexable_new:
+        # must be both True or both False
+        if old_idx ^ new_idx:
```
jshipton, line 53: change to `old_idx is None ^ new_idx is None`

PR #397, opened 2023-07-25
Move theta input and change error check

atb1995:
theta is changed from an optional argument in `ThetaMethod` to required and limited to `0 < theta < 1`. `ImplicitMidpoint` is changed to be called `TrapeziumRule`. Note, in future we could add the actual Implicit Method in future.

### DIFF
--- a/examples/compressible/dcmip_3_1_meanflow_quads.py
+++ b/examples/compressible/dcmip_3_1_meanflow_quads.py
@@ -84,7 +84,7 @@ diagnostic_fields = [Perturbation('theta'), Perturbation('rho'), CompressibleKin
 io = IO(domain, output, diagnostic_fields=diagnostic_fields)
 
 # Transport schemes
-transported_fields = [ImplicitMidpoint(domain, "u"),
+transported_fields = [TrapeziumRule(domain, "u"),
                       SSPRK3(domain, "rho", subcycles=2),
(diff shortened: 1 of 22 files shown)


tommbendall (commented):
The code changes are exactly right so as it stands this could be merged onto main. But...
1. Could you write a short description of the PR (under the "conversation" tab)
2. Do we want to think about renaming this? I've always wanted to clarify the difference between the following two schemes:
```
q^{n+1} = q^n + 0.5*dt*F(q^n) + 0.5*dt*F(q^{n+1})
```
and
```
q^{n+1} = q^n + dt*F(0.5*[q^n+q^{n+1}])
```
Since this one is the first, I had wondered if this should be called something like the `TrapezoidalThetaMethod`? What do you think?

atb1995:
I've changed the name of ImplicitMidpoint

tommbendall:
I'm afraid the branch is still failing! I also wondered if we should open an issue to implement `GeneralisedMidpoint` and `ImplicitMidpoint` discretisations?

atb1995:
Yep I missed an `ImplicitMidpoint` and had some trailing whitespace! I think that would be good, although I think it will be tied in to having implicit multistage schemes with a butcher tableau (after explicit is done). I could still create an issue?

tommbendall (approved):
Looks good now, thanks!

PR #398, opened 2023-07-26
Updated logging capabilities

### LINKED ISSUE #394: Gusto logger should produce a log file

JDBetteridge:
Fixes #394

### DIFF
--- a/gusto/configuration.py
+++ b/gusto/configuration.py
@@ -1,35 +1,16 @@
 """Some simple tools for configuring the model."""
 from abc import ABCMeta, abstractproperty
 from enum import Enum
-import logging
-from logging import DEBUG, INFO, WARNING
 from firedrake import sqrt, Constant
+from gusto.logging import logger
 
 
-__all__ = ["WARNING", "INFO", "DEBUG", "IntegrateByParts",
-           "TransportEquationType", "OutputParameters",
-           "CompressibleParameters", "ShallowWaterParameters",
-           "logger", "EmbeddedDGOptions", "RecoveryOptions", "SUPGOptions",
-           "SpongeLayerParameters", "DiffusionParameters"]
-
-logger = logging.getLogger("gusto")
-
-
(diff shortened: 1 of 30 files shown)


gusto/timeloop.py
```diff
@@ -3,14 +3,14 @@
 from abc import ABCMeta, abstractmethod, abstractproperty
 from firedrake import Function, Projector, Constant, split
 from pyop2.profiling import timed_stage
-from gusto.configuration import logger
 from gusto.equations import PrognosticEquationSet
-from gusto.forcing import Forcing
+from gusto.fields import TimeLevelFields, StateFields
```
tommbendall, line 7: Thanks for alphabetising this!

gusto/configuration.py
```diff
@@ -87,12 +68,15 @@ def __setattr__(self, name, value):
             AttributeError: if the :class:`Configuration` object does not have
                 this attribute pre-defined.
         """
+        if name == 'log_level':
+            logger.warning("Set log level with environemnt variables, see `logging.py` for details")
```
tommbendall, line 72: Typo here

tommbendall, line 72:
```suggestion
            logger.warning("Set log level with environment variables, see `logging.py` for details")
```

JDBetteridge:
There's more to come, I think I can get KSP residuals to be logged. I think what we need is a custom KSP monitor, like I have written [here](https://github.com/JDBetteridge/gusto_linesmoother/blob/master/monitor.py) (but _much_ simpler), which logs the exact same information that is output by `ksp_monitor_true_residual` to the Python logger and then turn off the PETSc internal logger to avoid duplication.

JDBetteridge:
I think this is ready for another review :slightly_smiling_face:

tommbendall (approved):
This is really good. I can confirm that I have run an example in serial and in parallel on my machine and that the log files appear in the results directory for that example.

Thanks very much!

PR #399, opened 2023-07-26
Evaluate physics at each stage of SSPRK3, RK4 and Heun's method

nhartney:
This PR fixes a bug where physics schemes were only being called once at the beginning of the SSPRK3, RK4 and Heun's method time-stepping schemes. Physics source terms are now evaluated at the beginning of each stage of the timestep.

### DIFF
--- a/gusto/time_discretisation.py
+++ b/gusto/time_discretisation.py
@@ -374,13 +374,13 @@ class SSPRK3(ExplicitTimeDiscretisation):
 
         elif stage == 1:
             for evaluate in self.evaluate_source:
-                evaluate(x_in, self.dt)
+                evaluate(self.x1, self.dt)
             self.solver.solve()
             self.x1.assign(0.75*x_in + 0.25*self.x_out)
 
         elif stage == 2:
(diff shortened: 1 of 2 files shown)


gusto/time_discretisation.py
```diff
@@ -470,16 +474,22 @@ def solve_stage(self, x_in, stage):
             self.x1.assign(x_in + 0.5 * self.dt * self.k1)
 
         elif stage == 1:
+            for evaluate in self.evaluate_source:
+                evaluate(x_in, self.dt)
```
JHopeCollins, line 478: Should the source term be evaluated at `x1` instead of `xin`?
If I've read it correctly , `xin` is the solution at the beginning of the subcycle and doesn't get updated at each stage, whereas `x1` is the solution at each stage (ie the argument to the function F in the class doctring)

nhartney: Yes I think so! Thanks.

tommbendall (approved):
Looks good to me now, thanks Nell!

Issue #342, opened 2023-01-20
It is assumed that the compatible finite element spaces have the same ordering as the prognostic variables


tommbendall:
The equation sets are written so that the order of the prognostic variables matches the order of the compatible finite element spaces. We should not assume this and should come up with a better way to specify which variable is in which space.

jshipton:
Taken care of by #402

Issue #338, opened 2023-01-20
CG space used for vorticity field


tommbendall:
Which CG space should be used for the vorticity diagnostic? We are requesting degree + 2 but should it be degree + 1?

Issue #254, opened 2022-07-19
Implement BDFM family


tommbendall:
The BDFM family aren't covered by the existing space creator. It would be great to enhance it to cover these spaces.

jshipton:
Done in #402

PR #402, opened 2023-07-27
Build the whole de Rham complex in the domain

tommbendall:
This PR:
- expands the space building so that we always build the H1 space and HCurl space (if this exists), thus giving us the whole de Rham complex
- specifies what the appropriate H1 space is for a particular de Rham complex, which should be used for any vorticity field, thus addressing #338 
- adds the correct handling for the BDFM space, thus addressing #254 
- there is a unit-test for space creation
- introduces `space_names` as an argument and property for equations sets. This means that the space for prognostic variables can now be specified by the user (although the defaults are of course the same). This addresses #342 
- the space that was registered as `"DG"` is also registered `"L2"`
- I changed some `"` to `'` in `equations.py` which just makes the code a bit more consistent

### DIFF
--- a/gusto/diagnostics.py
+++ b/gusto/diagnostics.py
@@ -1,6 +1,6 @@
 """Common diagnostic fields."""
 
-from firedrake import op2, assemble, dot, dx, FunctionSpace, Function, sqrt, \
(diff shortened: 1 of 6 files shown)


jshipton (approved):
This looks fantastic - thanks @tommbendall !!

Issue #335, opened 2023-01-20
Spherical coordinate transform routines should be organised


tommbendall:
There are multiple implementations of coordinate transforms between spherical and Cartesian coordinates / components. These should be organised and made available from a single module.

tommbendall:
Closing as this was fixed by #409

PR #409, opened 2023-07-27
Organise and enhance spherical coordinate transforms

tommbendall:
Adds some handy routines for spherical coordinate transforms, and reorganises some of our existing ones. This addresses #335 .

There are quite aspects to this change:

1. There are new routines to:
- convert between (x,y,z) and (lon,lat,r) coordinates
- convert between the (x,y,z) and (lon,lat,r) components of vectors
- calculate distances between points on (a) periodic Cartesian meshes and (b) on the surface of a sphere
- compute rotated (lon,lat,r) coordinates, effectively allowing spherical meshes to be easily rotated

Underpinning these new routines is the aim to have a set of routines that can be used both with Firedrake expressions and numpy arrays.

2. The model's internal coordinates can now be rotated (lon,lat,r) coordinates. The rotation of coordinates is threaded through the model by passing it to the `Domain` and `Coordinates` objects. The rotated coordinates are now used in the Williamson 2 test.

3. There is now a calculation in the `Domain` object to evaluate the radius of the mesh that has been provided. When outputting for a 3D spherical domain, the vertical coordinate is now `r-radius` instead of `r`.

4. The vector component diagnostics have been consolidated and use the rotation capability.

5. There are unit-tests for all of the new routines.


jshipton (approved):
Looks great!

PR #410, opened 2023-07-28
Tidy examples

jshipton:
The aim of this PR is to reduce the run time of the tests while retaining sufficient coverage of functionality.
* edit the thermal shallow water example to run for just one timestep when running via pytest
* reduce spatial resolution of DCMIP test when running via pytest
* remove some examples that duplicate functionality tested by running other examples - these will go to the case studies repository


tommbendall (approved):
This is great thanks! We can now move the deleted examples to our `case_studies`

PR #411, opened 2023-07-28
Fix a parallel bug in the domain extents used in the domain metadata

tommbendall:
The problem was a `my_rank` should have been a `procid`. But it's better to use a `bcast` instead of `send` and `recv`


atb1995:
I'm happy with these changes - seems like a much neater way of doing it!

atb1995 (approved):

PR #412, opened 2023-07-28
Moist thermal sw

nhartney:
This PR adds the capability for the moist thermal shallow water equations. It:

- Adds a new physics class, `SWSaturationAdjustment`, to do condensation and evaporation between two moist species in the shallow water equations. This class takes `saturation_curve` as an argument, and where the water vapour exceeds the saturation curve a specified fraction of the excess vapour is converted to cloud. Conversely, in a sub-saturated atmosphere where cloud is present, a specified fraction of the cloud will evaporate. If `thermal_feedback` is true then these processes affect the buoyancy equation - evaporation should cause an increase in buoyancy and condensation a decrease.
- Includes an integration test for `SWSaturationAdjustment` which tests both the evaporation and condensation ability by running two different scenarios.
- Adds a `moist_thermal_williamson5` example to the `shallow_water` directory. This is based on test 2 in Zerroukat and Allen (2015), which is a version of test 5 from Williamson et al (flow over a mountain) with additional thermodynamic variables. 
- Renames the `gamma` parameter in `InstantRain` to `beta1` to make clearer the distinction between convective feedback (`beta1`) and thermal feedback (`beta2`).  Adds a `gamma_r` parameter to `InstantRain` to allow the conversion of a specified fraction of vapour above the saturation threshold to rain. 
- Corrects a mistake in the specification of the initial conditions for the thermal Williamson 2 example.


tommbendall (approved):
This was already approved in #390 and has just been moved across as something odd happened with Github.

Looks great and thanks for doing this!

Issue #334, opened 2023-01-20
Unclear if precipitation is correct


tommbendall:
The `Fallout` parametrisation is tricky because it is a physics term which involves a transport term. I'm not sure if it is being correctly, so at some point this should be thought about properly. It may be easiest to investigate using the split physics timestepper if we can properly specify the splitting order here.

atb1995:
There are two bugs that cancel each other out:

1. The transport term is included in the prognostic equation, meaning if it were to be transported by the physics scheme it would be double counted
2. The transport term somehow isn't included in the physics scheme, this means in the case of the `test_precipitation.py` test the physics scheme isn't doing anything.

This showed itself when I used a physics time stepper that uses no `time_derivative` terms in the `rhs` `residual`. This threw a zero term error. I have `xfail`ed this test in my pull request.

tommbendall:
I think it's very clear that this is incorrect! Thanks for helping to investigate this, I think I now understand what has been going on here. Several issues around this need addressing:

1. It is confusing that the `PrescribedTransport` timestepper is allowed to have physics terms when the `Timestepper` is not. I think both should be allowed to take physics terms -- the difference between these two steppers should just be whether the transporting wind needs updating as a function of time.
2. The way that `PrescribedTransport` applies the physics schemes is confusing. We pair physics schemes with time discretisations, but the `Timestepper` and `PrescribedTransport` should evaluate all of the terms (including physics terms) at the same time, so physics schemes should not be allowed to have separate time discretisations for these time steppers. In contract, a `SplitTimestepper` or `SemiImplicitQuasiNewton` should allow different time discretisations to be used for different terms.
3. Physics schemes should also have a label to identify the physics scheme, and not just have a general `physics` label. Otherwise the time discretisation that is paired with a particular physics scheme will act on all physics schemes, as we will be unable to filter the terms associated with individual physics schemes. This would be a problem for us using multiple physics schemes.
4. There is a bug in the residual used in `Fallout`, https://github.com/firedrakeproject/gusto/blob/main/gusto/physics.py#L302:
```
equation.residual += physics(subject(adv_term, equation.X), self.evaluate)
```
should include the `prognostic` to become
```
equation.residual += physics(prognostic(subject(adv_term, equation.X), rain_name), self.evaluate)
```
5. the contribution to `equation.residual` from `Fallout` should not include `transport` label so that this term does not get picked up by the general dynamics transport. This should be done just before line 302 in `physics.py`

[...]

PR #413, opened 2023-07-31
Butcher Tableau

atb1995:
We have created a new time stepper class `ExplicitMultistage` which takes in a Butcher Tableau in a numpy array of the form: [a_00   0   .       0  ]
[a_10 a_11  .       0  ]
[  .    .   .       .  ]
[ b_0  b_1  .   b_{s-1}]
to run single time step multistage explicit time discretisation's. This allows time steppers such as `ForwardEuler`, `RK4`, `SSPRK3` and `Heun` to be rewritten by the use of a single Butcher Tableau. It will also allow explicit multistage schemes to be easily written in future. 

Unfortunately two bugs were found in the process. The first is that the `Fallout` physics scheme has a few bugs in, meaning `test_precipitation.py` fails. I have xfailed this pytest for now. This problem is detailed in issue #334.

The second was that `RK4` on main fails in some pytests when using the `NonlinearVariationalSolver`. This failure translated to all `ExplicitMultistage` as they are written in the same way as `RK4`. I have changed to `LinearVariationalSolver` in `ExplicitMultistage`, and have created a minimal failing example outside of Gusto, which I will raise with Firedrake team.


gusto/time_discretisation.py
```diff
@@ -286,6 +284,24 @@ def setup(self, equation, apply_bcs=True, *active_labels):
...
+
+        return l.form
+
+    @cached_property
+    def solver(self):
+        """Set up the problem and the solver."""
+        # setup linear solver using lhs and rhs defined in derived class
+        problem = NonlinearVariationalProblem(self.lhs - self.rhs, self.x_out, bcs=self.bcs)
+        solver_name = self.field_name+self.__class__.__name__
+        return NonlinearVariationalSolver(problem, solver_parameters={'snes_type': 'ksponly'} | self.solver_parameters, options_prefix=solver_name)
```
tommbendall, line 303: The `|` here is cool python to combine the dictionaries, so I'm glad to have learnt something new!

I wonder if you want to add a comment in the line above to explain why we need this and to explain the cool python. e.g:
```
# If snes_type not specified by user, set this to ksp only to avoid outer Newton iteration
```

gusto/time_discretisation.py
```diff
@@ -315,123 +331,55 @@ def apply(self, x_out, x_in):
         x_out.assign(self.x1)
 
 
-class ForwardEuler(ExplicitTimeDiscretisation):
+class ExplicitMultistage(ExplicitTimeDiscretisation):
```
tommbendall, line 334: I don't mind `ExplicitMultistage`, but have also wondered if `ExplicitRungeKutta` is clearer?

gusto/time_discretisation.py
```diff
@@ -471,39 +411,23 @@ def rhs(self):
...
-        Args:
-            x_in (:class:`Function`): field at the start of the stage.
-            stage (int): index of the stage.
-        """
-        if stage == 0:
-            self.solver.solve()
-            self.k1.assign(self.x_out)
-            self.x1.assign(x_in + 0.5 * self.dt * self.k1)
+    def solve_stage(self, x0, stage):
+        self.x1.assign(x0)
```
tommbendall, line 415: I think we need to include the call to evaluate physics terms here (this was included after you branched off from trunk!). But you should check if makes sense to evaluate the physics sources at exactly this point...
```suggestion
        self.x1.assign(x0)
        for evaluate in self.evaluate_source:
            evaluate(self.x1, self.dt)
```

gusto/time_discretisation.py
```diff
@@ -471,39 +411,23 @@ def rhs(self):
...
 
-        elif stage == 1:
-            for evaluate in self.evaluate_source:
-                evaluate(self.x1, self.dt)
-            self.solver.solve()
-            self.k2.assign(self.x_out)
-            self.x1.assign(x_in + 0.5 * self.dt * self.k2)
+        if (stage == self.nStages - 1):
+            self.x1.assign(x0)
+            for i in range(self.nStages):
```
tommbendall, line 425: I guess we need to add the physics source evaluation here too?

gusto/time_discretisation.py
```diff
@@ -518,12 +442,69 @@ def apply_cycle(self, x_out, x_in):
...
+    k0 = F[y^n]
+    k1 = F[y^n + dt*k1]
+    k2 = F[y^n + (1/4)*dt*(k0+k1)]
+    y^(n+1) = y^n + (1/6)*dt*(k0 + k1 + 4*k2)
+    """
+    def __init__(self, domain, field_name=None, subcycles=None, solver_parameters=None, limiter=None, options=None, butcher_matrix=None):
+        super().__init__(domain, field_name=field_name, subcycles=subcycles,
+                         solver_parameters=solver_parameters,
+                         limiter=limiter, options=options, butcher_matrix=butcher_matrix)
+        self.butcher_matrix = np.array([[1., 0., 0.], [1./4., 1./4., 0.], [1./6., 1./6., 2./3.]])
```
tommbendall, line 480: I really like how clear this is now

tommbendall (changes requested):
This looks really neatly done to me. I think the main thing is to make sure that we're including the physics source evaluations (which were added in after you started this branch)

gusto/time_discretisation.py
atb1995: I think I may veto this.. I'm keener on multistage (especially in contrast to multistep)

gusto/time_discretisation.py
atb1995: I've added this!

gusto/time_discretisation.py
atb1995: Not required here, as this is the final timestep evaluation of field (after solve)

gusto/time_discretisation.py
```diff
@@ -315,123 +332,55 @@ def apply(self, x_out, x_in):
         x_out.assign(self.x1)
 
 
-class ForwardEuler(ExplicitTimeDiscretisation):
+class ExplicitMultistage(ExplicitTimeDiscretisation):
     """
-    Implements the forward Euler timestepping scheme.
+    A class for implementing general explicit multistage methods based on its Butcher tableau.
```
tommbendall, line 337:
```suggestion
    A class for implementing general explicit multistage (Runge-Kutta) methods based on its Butcher tableau.
```

tommbendall (approved):
I'm happy with this, but suggest including "Runge-Kutta" in the docstring for the new class

jshipton (changes requested):
The only thing is to take care with line length, otherwise this looks great - thanks @atb1995 !

PR #415, opened 2023-08-02
Coupled transport

ta440:
This branch adds a CoupledTransportEquation class which will transport multiple active tracers of varying types and equation forms.
The ForcedAdvectionEquation() class is removed. If a single variable is to be transported, then the simpler AdvectionEquation() can be called instead.
This new equation class is tested in the /integration-tests/equations/test_coupled_transport.py file.
The integration test, equations/test_forced_advection.py is changed to work with the new equation type, as this test has two tracers, of rain and water vapour.


gusto/equations.py
```diff
@@ -81,6 +81,7 @@ def __init__(self, domain, function_space, field_name, Vu=None):
                 velocity field. If this is not specified, uses the HDiv spaces
                 set up by the domain. Defaults to None.
             **kwargs: any keyword arguments to be passed to the advection form.
+            (SHOULD THIS BE AN ARGUMENT IN THE ABOVE FUNCTION????)
```
tommbendall, line 84: I think you can delete the `**kwargs` where it appears here, thanks!

gusto/equations.py
```diff
@@ -484,36 +485,37 @@ def generate_tracer_transport_terms(self, active_tracers):
         return adv_form
 
 
-class ForcedAdvectionEquation(PrognosticEquationSet):
+class CoupledTransportEquation(PrognosticEquationSet):
     u"""
-    Discretises the advection equation with a source/sink term,               \n
+    Discretises a transport equation with a source/sink term,               \n
```
tommbendall, line 490: I would be tempted to remove the direct reference to source/sink in the comments here, but also to add a comment at the end of the description here saying "This equation can also be enhanced through the addition of sources or sinks by using it with physics schemes."

gusto/equations.py
```diff
@@ -484,36 +485,37 @@ def generate_tracer_transport_terms(self, active_tracers):
...
         Args:
             domain (:class:`Domain`): the model's domain object, containing the
                 mesh and the compatible function spaces.
-            function_space (:class:`FunctionSpace`): the function space that the
-                equation's prognostic is defined on.
-            field_name (str): name of the prognostic field.
+            active_tracers (list): a list of `ActiveTracer` objects
+                that encode the metadata for any active tracers to be included
+                in the equations. This is required for using this class; if there
+                is only a field to be advected, use ForcedAvectionEquation()
```
tommbendall, line 505: So I think we're now wondering about not having a `ForcedAdvectionEquation`, so this could say user "AdvectionEquation" or "TransportEquation" instead

integration-tests/equations/test_coupled_transport.py
```diff
@@ -0,0 +1,63 @@
+"""
+Tests the CoupledTransportEquation() class.
```
tommbendall, line 2: I think you don't need the brackets here

integration-tests/equations/test_coupled_transport.py
```diff
@@ -0,0 +1,63 @@
...
+def run(timestepper, tmax, f_end):
+    timestepper.run(0, tmax)
+    norm1 = norm(timestepper.fields("f1") - f_end) / norm(f_end)
+    norm2 = norm(timestepper.fields("f2") - f_end) / norm(f_end)
+    return (norm1 + norm2)/2
+
+
+@pytest.mark.parametrize("geometry", ["slice", "sphere"])
+@pytest.mark.parametrize("equation_form1", ["advective", "continuity"])
+@pytest.mark.parametrize("equation_form2", ["advective", "continuity"])
```
tommbendall, line 23: This is great, I think it's the right idea to test these combinations

integration-tests/equations/test_coupled_transport.py
```diff
@@ -0,0 +1,63 @@
...
+    setup = tracer_setup(tmpdir, geometry)
+    domain = setup.domain
+
+    if equation_form1 == "advective":
+        tracer1 = ActiveTracer(name='f1', space='DG',
+                               variable_type=TracerVariableType.density,
+                               transport_eqn=TransportEquationType.advective)
+    else:
+        tracer1 = ActiveTracer(name='f1', space='DG',
+                               variable_type=TracerVariableType.density,
```
tommbendall, line 34: I suggest that you keep `mixing_ratio` always with `transport_eqn=TransportEquationType.advective` and `density` with `transport_eqn=TransportEquationType.conservative`. Still test a mixture of advective and conservative variables, but since mixing ratios obey the advective form it would be better to keep that in this test

integration-tests/equations/test_coupled_transport.py
```diff
@@ -0,0 +1,63 @@
...
+    transport_method = [DGUpwind(eqn, 'f1'), DGUpwind(eqn, 'f2')]
+
+    timestepper = PrescribedTransport(eqn, transport_scheme, setup.io, transport_method)
+
+    # Initial conditions
+    timestepper.fields("f1").interpolate(setup.f_init)
+    timestepper.fields("f2").interpolate(setup.f_init)
+    timestepper.fields("u").project(setup.uexpr)
+
+    error = run(timestepper, setup.tmax, setup.f_end)
```
tommbendall, line 61: Something I'm not sure about is how this error is calculated? I think it might be better for us to separately check the error for each of the tracers, by doing something like (warning I didn't test this!):
```
_ = run(timestepper, setup.tmax, setup.f_end)  # Run the model but don't use returned error
error_1 = errornorm(timestepper.fields("f1"), setup.f_end)
error_2 = errornorm(timestepper.fields("f2"), setup.f_end)

assert error_1 < setup.tol, \
    'The transport error for tracer 1 is greater than the permitted tolerance'
assert error_2 < setup.tol, \
    'The transport error for tracer 2 is greater than the permitted tolerance'
```

tommbendall (changes requested):
Thanks very much for this! I have just a few small suggestions

Issue #416, opened 2023-08-02
Cannot run Non-hydrostatic Skamarock Klemp in parallel


JDBetteridge:
The non-hydrostatic Skamarock Klemp example does not run in parallel, is this expected?

Full traceback below:

```
$ mpiexec -n 2 python examples/compressible/skamarock_klemp_nonhydrostatic.py --running-tests
firedrake:WARNING OMP_NUM_THREADS is not set or is set to a value greater than 1, we suggest setting OMP_NUM_THREADS=1 to improve performance
gusto:INFO Physical parameters that take non-default values:
gusto:INFO 
Traceback (most recent call last):
  File "/scratch/jbetteri/firedrake_py311_opt/src/gusto/examples/compressible/skamarock_klemp_nonhydrostatic.py", line 125, in <module>
Traceback (most recent call last):
  File "/scratch/jbetteri/firedrake_py311_opt/src/gusto/examples/compressible/skamarock_klemp_nonhydrostatic.py", line 125, in <module>
    stepper.run(t=0, tmax=tmax)
    stepper.run(t=0, tmax=tmax)
  File "/scratch/jbetteri/firedrake_py311_opt/src/gusto/gusto/timeloop.py", line 582, in run
  File "/scratch/jbetteri/firedrake_py311_opt/src/gusto/gusto/timeloop.py", line 582, in run
    super().run(t, tmax, pick_up=pick_up)
    super().run(t, tmax, pick_up=pick_up)
  File "/scratch/jbetteri/firedrake_py311_opt/src/gusto/gusto/timeloop.py", line 162, in run
  File "/scratch/jbetteri/firedrake_py311_opt/src/gusto/gusto/timeloop.py", line 162, in run
    self.io.setup_dump(self.fields, t, pick_up)
    self.io.setup_dump(self.fields, t, pick_up)
  File "/scratch/jbetteri/firedrake_py311_opt/src/gusto/gusto/io.py", line 412, in setup_dump
  File "/scratch/jbetteri/firedrake_py311_opt/src/gusto/gusto/io.py", line 412, in setup_dump
    self.dump(state_fields, t)
  File "/scratch/jbetteri/firedrake_py311_opt/src/gusto/gusto/io.py", line 549, in dump
    self.dump(state_fields, t)
  File "/scratch/jbetteri/firedrake_py311_opt/src/gusto/gusto/io.py", line 549, in dump
    self.pointdata_output.dump(state_fields, t)
[...]

PR #417, opened 2023-08-02
Test examples in parallel

### LINKED ISSUE #377: there is no test for running in parallel

JDBetteridge:
Given that most users will probably try and run an example from the `examples` directory in parallel, enable parallel testing for all examples.

I do not anticipate this adding significant time to CI runs as the parallel tests will benefit from having a hot cache.

Any test in `unit-tests` or `integration-tests` can be made to run in parallel by using the Python decorator:
```python
@pytest.mark.parallel(nprocs=4)
```
The above example would run the test with 4 MPI ranks.

Fixes #377.

Requires #416 to be addressed.


JDBetteridge:
Tests take ~3min longer, seems like a win

gusto/io.py
```diff
@@ -15,6 +15,10 @@
 __all__ = ["pick_up_mesh", "IO"]
 
 
+class GustoIOError(IOError):
```
tommbendall, line 18: OK this is fun! Should we use this in place of our other `IOError`?

tommbendall: Which is on line 315...

tommbendall (changes requested):
This is great, thanks! It raises the question of which of the other unit tests or integration tests should be in parallel. The nice thing about the examples is that they cover most of the capabilities so this should give us reasonable coverage.

Just one small suggestion...

gusto/io.py
JDBetteridge: Done!

Yes, this is a neat trick if you want to catch your own IO exceptions, but not general IO exceptions.

For instance if the disk is full you probably want to let that error get raised as an `IOError` and crash the program, but if it's a known Gusto shortcoming that you may wish to handle (like `PointDataOutput` does not work in parallel) you can raise a `GustoIOError` and catch & handle the exception.

If you like the pattern, it might be worth having an `exceptions.py` in the repo with all of your custom exceptions, but I didn't think it was worth it for 2 (there's also a `LoggingError` which I should have probably called `GustoLoggingError`).

JDBetteridge:
I agree, it's probably worth having a discussion at some point which of the unit/integration tests should be run in parallel. Using the `pytest` marker makes it very easy to update existing tests.

A short term suggestion would be: If you find any parallel bugs add a parallel test, or update an existing test to run in parallel.

JDBetteridge:
Twice in a row! Nice!

I will pull out my debugging code and then this is ready for review

tommbendall:
I'd really like us to start parallel testing by default, so I thought it would be helpful to update the branch.

But as we saw a few months ago, we can still get failures. On this latest run it was the `skamarock_klemp_hydrostatic` test, and it looks like other failures we've seen:

On the first call to the mixed solver, one rank seems to fall behind the others and not begin the `ImplicitSolver_solve`. Then all the ranks hang. Output from most log files:
```
2024-03-13 08:36:20,335 INFO     Semi-implicit Quasi-Newton: Mixed solve (0, 1)
2024-03-13 08:36:20,335 INFO     Compressible linear solver: rho average solve
2024-03-13 08:36:20,340 INFO     Compressible linear solver: Exner average solve
2024-03-13 08:36:20,345 INFO     Compressible linear solver: hybridized solve
2024-03-13 08:36:20,359 DEBUG            Residual norms for ImplicitSolver_condensed_field_ solve
2024-03-13 08:36:20,359 DEBUG            0 KSP unpreconditioned resid norm 4.413507124086e+07 true resid norm 4.413507124086e+07 ||r(i)||/||b|| 1.000000000000e+00
2024-03-13 08:36:20,362 DEBUG            1 KSP unpreconditioned resid norm 5.900637824684e+01 true resid norm 5.900637824491e+01 ||r(i)||/||b|| 1.336949881035e-06
2024-03-13 08:36:20,364 DEBUG            2 KSP unpreconditioned resid norm 5.041852127746e-05 true resid norm 5.042033608880e-05 ||r(i)||/||b|| 1.142409758753e-12
2024-03-13 08:36:20,376 DEBUG        Residual norms for ImplicitSolver_ solve
2024-03-13 08:36:20,376 DEBUG        0 KSP no resid norm 4.765315244582e+07 true resid norm 5.046385477193e-05 ||r(i)||/||b|| 1.058982505498e-12
```

Output from one log file:
```
2024-03-13 08:36:20,335 INFO     Semi-implicit Quasi-Newton: Mixed solve (0, 1)
2024-03-13 08:36:20,335 INFO     Compressible linear solver: rho average solve
2024-03-13 08:36:20,340 INFO     Compressible linear solver: Exner average solve
2024-03-13 08:36:20,345 INFO     Compressible linear solver: hybridized solve
2024-03-13 08:36:20,359 DEBUG            Residual norms for ImplicitSolver_condensed_field_ solve
2024-03-13 08:36:20,359 DEBUG            0 KSP unpreconditioned resid norm 4.413507124086e+07 true resid norm 4.413507124086e+07 ||r(i)||/||b|| 1.000000000000e+00
2024-03-13 08:36:20,362 DEBUG            1 KSP unpreconditioned resid norm 5.900637824684e+01 true resid norm 5.900637824491e+01 ||r(i)||/||b|| 1.336949881035e-06
2024-03-13 08:36:20,364 DEBUG            2 KSP unpreconditioned resid norm 5.041852127746e-05 true resid norm 5.042033608880e-05 ||r(i)||/||b|| 1.142409758753e-12
```

tommbendall:
In the failing example that I posted above, the hanging comes between the `condensed_field` solve (the preconditioning step for the hybridized solver) -- which suggests to me that it could be worth reviewing whether the solver options are valid in parallel for our domains:
```
    solver_parameters = {'mat_type': 'matfree',
                         'ksp_type': 'preonly',
                         'pc_type': 'python',
                         'pc_python_type': 'firedrake.SCPC',
                         'pc_sc_eliminate_fields': '0, 1',
                         # The reduced operator is not symmetric
                         'condensed_field': {'ksp_type': 'fgmres',
                                             'ksp_rtol': 1.0e-8,
                                             'ksp_atol': 1.0e-8,
                                             'ksp_max_it': 100,
                                             'pc_type': 'gamg',
                                             'pc_gamg_sym_graph': None,
                                             'mg_levels': {'ksp_type': 'gmres',
                                                           'ksp_max_it': 5,
                                                           'pc_type': 'bjacobi',
                                                           'sub_pc_type': 'ilu'}}}
```

tommbendall:
I don't think we actually use a mesh hierarchy in any tests -- we only use algebraic multigrid and not geometric, although this is definitely something that we need to do to improve performance.

I agree about the minimum failing example, the hard thing is that sometimes it will pass!!

JDBetteridge:
Some recent changes in Firedrake my have fixed some parallel bugs. I'm running CI again and crossing my fingers :crossed_fingers: .

If this still doesn't work we should try tackling this at the hackathon.

tommbendall (approved):
This has been talked about extensively offline. Thanks Jack and I approve!

PR #418, opened 2023-08-07
Fix precipitation

tommbendall:
This attempts to fix precipitation (issue #334) by:
- ensuring that the advection term associated with `Fallout` has its `transport` label removed
- adding the `prognostic` label to that term

It also attempts to resolve some outstanding physics problems by:
- allowing the `Timestepper` to take physics schemes
- the `PrescribedTransport` time stepper is now only different in allowing the wind to be time-varying

I also implemented a routine to return an equation's `active_tracer` object for a particular species, which should help with error checking in the physics.


tommbendall:
Thanks Alex! Yes that's right about the physics not wanting a separate time discretisation for the `PrescribedTransport`

PR #420, opened 2023-08-08
Fix subtle parallel logging bug

JDBetteridge:


tommbendall (approved):
This works for me, thanks for putting this in!

PR #423, opened 2023-08-15
Rexi updated

jshipton:
This contains code to compute the exponential of the linear operator defined by linearising an equation.

The code is tested for shallow water and compressible Boussinesq and there are also some tests for the REXI approximation.


jshipton (commented):
@nhartney could you update the checkpoint data file to contain known correct output from REXI so that we can then make the test tolerances smaller? Thanks!

gusto/core/io.py
```diff
@@ -21,7 +21,7 @@ class GustoIOError(IOError):
     pass
 
 
-def pick_up_mesh(output, mesh_name):
+def pick_up_mesh(output, mesh_name, comm=COMM_WORLD):
```
tommbendall, line 24: We had a brief chat about the addition of the `comm` argument here last week -- this makes sense to me now.

tommbendall (approved):
I am happy with everything here -- we just need to merge in main and resolve conflicts

tommbendall (approved):
Thanks Jemma, exciting!

PR #424, opened 2023-08-23
Log the Courant number

tommbendall:
This introduces a system to log the Courant number, which will be done by default.

Main points:
- a `log_courant` option is passed to the output configuration
- the `CourantNumber` diagnostic is augmented so that it can be evaluated on an expression, and not just the field named "u" (this is used when the transporting velocity isn't exactly the wind field)
- some plumbing has been added to the IO, so that the Courant number diagnostics are set up if needed. The diagnostics are used to evaluated the maximum value (which is logged)
- the calls to log the Courant number are added to the time loop. The `SemiImplicitQuasiNewton` time stepper logs the Courant number every outer loop

### DIFF
--- a/gusto/io.py
+++ b/gusto/io.py
@@ -256,10 +256,10 @@ class IO(object):
             courant_name = None if name == 'u' else name
 
             # Set up diagnostic if it hasn't already been
-            if courant_name not in diagnostic_names:
+            if courant_name not in diagnostic_names and 'u' in state_fields._field_names:
                 if expression is None:
                     diagnostic = CourantNumber(to_dump=False)
-                else:
+                elif expression is not None:
                     diagnostic = CourantNumber(velocity=expression, name=courant_name, to_dump=False)
 
                 self.diagnostic_fields.append(diagnostic)
@@ -278,7 +278,7 @@ class IO(object):
                 None.
         """
 
-        if self.output.log_courant:
+        if self.output.log_courant and 'u' in state_fields._field_names:
             diagnostic_names = [diagnostic.name for diagnostic in self.diagnostic_fields]
             courant_name = 'CourantNumber' if name == 'u' else 'CourantNumber_'+name
             courant_idx = diagnostic_names.index(courant_name)
--- a/gusto/io.py
+++ b/gusto/io.py
@@ -5,7 +5,7 @@ import itertools
 from netCDF4 import Dataset
 import sys
 import time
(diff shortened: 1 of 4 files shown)


tommbendall:
> The DG0 Courant number based on DG fluxes is a bit more useful for pinpointing exactly which cell things went bad in.

Oh interesting, thanks Colin! How would we go about calculating that?

dham:
> > The DG0 Courant number based on DG fluxes is a bit more useful for pinpointing exactly which cell things went bad in.
> 
> Oh interesting, thanks Colin! How would we go about calculating that?

Looks like your code is already set up to do it. You just set the output space to a DG0 `FunctionSpace`.

gusto/io.py
```diff
@@ -238,6 +238,60 @@ def log_parameters(self, equation):
...
+        """
+        Sets up Courant number diagnostics to be logged.
+
+        Args:
+            state_fields (:class:`StateFields`): the model's field container.
+            name (str, optional): the name of the field to log the Courant
+                number of. Defaults to 'u'.
+            expression (:class:`ufl.Expr`, optional): expression of velocity
+                field to take Courant number of. Defaults to None, in which case
+
```
jshipton, line 251: unfinished comment...

gusto/io.py
```diff
@@ -238,6 +238,60 @@ def log_parameters(self, equation):
...
+    def log_courant(self, state_fields, name='u', message=None):
+        """
+        Logs the maximum Courant number value.
+
+        Args:
+            state_fields (:class:`StateFields`): the model's field container.
+            name (str, optional): the name of the field to log the Courant
+                number of. Defaults to 'u'.
+            message (str, optional): an extra message to be logged. Defaults to
```
jshipton, line 277: I like the option to add a message!

jshipton (changes requested):
looks good, and very useful - thanks! Just needs that comment finishing off...

jshipton (approved):

colinjcotter:
On 24 Aug 2023, at 15:26, David A. Ham ***@***.***> wrote:


Looks like your code is already set up to do it. You just set the output space to a DG0 FunctionSpace.

Message ID: ***@***.***>

No, it’s different.
See: https://github.com/firedrakeproject/asQ/blob/master/utils/diagnostics.py<https://github.com/firedrakeproject/asQ/blob/master/utils/diagnostics.py>

PR #425, opened 2023-08-23
A package of idealised physics routines

tommbendall:
This PR adds a package of idealised physics routines, that are largely taken from the DCMIP (2016) test case documentation.

1. A general `SourceSink` physics scheme, which takes some expression and evaluates it as a source or sink of a variable.
2. A `SurfaceFlux` scheme, which allows prescribed surface fluxes to be applied to theta and water vapour.
3. A `WindDrag` scheme.
4. A `StaticAdjustment` scheme
5. A `BoundaryLayerMixing` scheme

Additionally it implements a Firedrake "escape hatch", allowing us to break out of the standard structure of Firedrake `Function`s and to handle column-wise data instead. This is the only way to implement some physics schemes (such as dry static adjustment) and opens the route to us doing other non-FEM algorithms in the future.

Lots of other features have been pulled off this branch and brought onto main through separate PRs, but this branch still includes:
- a bit of a hack to allow `ExplicitMultiStage` schemes to work when there is no right-hand side


gusto/logging.py
```diff
@@ -186,7 +186,6 @@ def update_logfile_location(new_path, comm):
         fh.close()
         logger.removeHandler(fh)
 
-        os.makedirs(new_path, exist_ok=True)
```
jshipton, line 189: this needs to go back in

gusto/timeloop.py
```diff
@@ -345,12 +346,12 @@ def __init__(self, equation, scheme, io, spatial_methods=None,
         else:
             self.physics_schemes = []
 
-        for _, phys_scheme in self.physics_schemes:
+        for parametrisation, phys_scheme in self.physics_schemes:
             # check that the supplied schemes for physics are explicit
-            assert isinstance(phys_scheme, ExplicitTimeDiscretisation), \
-                "Only explicit time discretisations can be used for physics"
+            # assert isinstance(phys_scheme, ExplicitTimeDiscretisation), \
```
jshipton, line 351: This check needs to become a check for whether the time discretisation is explicit and whether this is allowed for the physics scheme (add a logical to the physics scheme).

gusto/timeloop.py
```diff
@@ -460,8 +462,8 @@ def __init__(self, equation_set, io, transport_schemes, spatial_methods,
 
         for _, scheme in self.all_physics_schemes:
             assert scheme.nlevels == 1, "multilevel schemes not supported as part of this timestepping loop"
-            assert isinstance(scheme, ExplicitTimeDiscretisation), \
-                "Only explicit time discretisations can be used for physics"
+            # assert isinstance(scheme, ExplicitTimeDiscretisation), \
```
jshipton, line 465: This check needs to become a check for whether the time discretisation is explicit and whether this is allowed for the physics scheme (add a logical to the physics scheme).

gusto/physics.py
```diff
@@ -152,7 +262,7 @@ def __init__(self, equation, vapour_name='water_vapour',
                 liquid_water += self.X.subfunctions[liq_idx]
 
         # define some parameters as attributes
-        self.dt = Constant(0.0)
+        self.dt = Constant(1.0)
```
jshipton, line 265: keep as zero as more likely to fail if not updated

gusto/physics.py
```diff
@@ -1002,3 +1119,624 @@ def evaluate(self, x_in, dt):
...
+        Args:
+            equation (:class:`PrognosticEquationSet`): the model's equation.
+            T_surface_expr (:class:`ufl.Expr`): the surface temperature.
+            vapour_name (str, optional): name of the water vapour variable.
+                Defaults to None, in which case no moisture fluxes are applied.
+            implicit_formulation (bool, optional): whether the scheme is already
+                put into a Backwards Euler formulation (which allows this scheme
+                to actually be used with a Forwards Euler or other explicit time
+                discretisation. Otherwise, this is formulated more generally and
+                can be used with any time stepper. Defaults to False.
```
jshipton, line 1148: close bracket

jshipton (changes requested):
just a few minor comments - this opens up a lot of cool stuff!

jshipton (approved):

PR #428, opened 2023-08-24
Refactor to avoid str.removeprefix

### LINKED ISSUE #427: Does not run on python 3.8
Whilst trying to run the gusto test suite on isambard (python 3.8), we received errors relating to [PEP616](https://peps.python.org/pep-0616/).

It looks like the main branch only has one such call to `str.removeprefix`:
https://github.com/firedrakeproject/gusto/blob/482ccbc56640590d39650c2d673a12721af738d6/gusto/logging.py#L180.

stephenpcook:
Closes #427

Method `str.removeprefix` only available from python 3.9.


EdHone:
Looks good to me: I was thinking that a fix like:

```python
filename = Path(old_path.name.replace("temp-", ""))
```
might be more consistent with what's currently there, but I think your way is probably better as it ensures that the ```temp-``` string is definitely a prefix.

jshipton (approved):

PR #432, opened 2023-08-29
Improve the Courant number calculation

tommbendall:
This implements the better Courant number calculation suggested by @colinjcotter, basically copying the code from https://github.com/firedrakeproject/asQ/blob/master/utils/diagnostics.py

It also allows us to separately print horizontal and vertical Courant numbers


jshipton (approved):
Thanks for doing this so quickly @tommbendall !

PR #434, opened 2023-09-01
Tracer density

ta440:
Added a tracer density diagnostic. This is useful when examining conservation of mass when transporting species via the dry density and mixing ratio.
Here are examples of plotting the normalised difference in the tracer density diagnostic over time, using the planar_nair_gaussian case as examined the Bendall, 'Trilemma' paper.
tracer_density_consistency
tracer_density_convergence


gusto/diagnostics.py
```diff
@@ -1581,3 +1582,34 @@ def setup(self, domain, state_fields):
...
         super().setup(domain, state_fields, vorticity_type="relative")
+
+
+class TracerDensity(DiagnosticField):
+    """Diagnostic for computing the density of a tracer. This is
+    computed as the product of a mixing ratio and dry density"""
+
+    name = "TracerDensity"
+
+    def __init__(self, m_X, rho_d):
```
tommbendall, line 1593: I think this `__init__` method should still have the arguments of the parent class, because the user might still want to specify the function space and the computation method.

We also need to make clear whether the `m_X` and `rho_d` arguments should be strings (the names of the fields) or the `Function` variables. I think if we change the name of the arguments to `mixing_ratio_name` and `density_name` then it is clearer?

So I suggest something like:
```
def __init__(self, mixing_ratio, density_name, space=None, method='interpolate', required_fields=()):
```

gusto/diagnostics.py
```diff
@@ -1581,3 +1582,34 @@ def setup(self, domain, state_fields):
...
+        """
+        Args:
+            m_X: the mixing ratio of the tracer
```
tommbendall, line 1596: We need to include the type of the argument in the docstrings, so this should be something like:
```
mixing_ratio_name (str), the name of the tracer mixing ratio variable.
```
and so on for the other variables

gusto/diagnostics.py
```diff
@@ -1581,3 +1582,34 @@ def setup(self, domain, state_fields):
...
+
+        Args:
+            domain (:class:`Domain`): the model's domain object.
+            state_fields (:class:`StateFields`): the model's field container.
+        """
+        m_X = state_fields(self.m_X)
+        rho_d = state_fields(self.rho_d)
+        space = m_X.function_space()
+        self.expr = m_X*rho_d
+        super().setup(domain, state_fields, space=space)
```
tommbendall, line 1615: If we include `space` as an argument to the `__init__` then we won't need to include it here

tommbendall (changes requested):

gusto/diagnostics.py
ta440: I agree with having more descriptive variable names. 
Is the required_fields=() just to initialise this variable before using it in:
 _super().__init__(method='interpolate', required_fields=(mixing_ratio_name, density_name))_
?

I've tested the four different methods with this diagnostic. It works when specifying 'interpolate' or 'project'. However, there are issues when specifying 'assign': 
_ValueError: All functions in the expression must have the same element as the assignee_
and 'solve':
_NotImplementedError: Solve method has not been implemented for diagnostic TracerDensity_

Should this diagnostic be able to be used with these methods, or is it to be expected that these aren't compatible?

tommbendall: Ah I think you're right, we don't need `required_fields` in the `__init__`, as this should just be set in the call to `super().__init__`.

For methods, `'assign'` only makes sense when the density and the mixing ratio field in the same space (what `assign` does is evaluate the values at the degrees of freedom). The solve method is there if we need to solve a more complicated equation to get a diagnostic, rather than interpolating/projecting/assigning. In this case I don't think we need a solve method so it should throw an error.

ta440: Ok, thanks.
I'll put in the function header that:
Valid options are 'interpolate', 'project' and 'assign', so that one doesn't try to use 'solve' with this

tommbendall (approved):
This all looks good to me, thanks Tim!

PR #438, opened 2023-09-04
Tiny fix to prevent all ranks trying to make directory

tommbendall:
I have been testing out our parallel logging which has been super useful. To get it to work, I had to ensure that we only create the results directory in one place (`io.setup_dump`), so this removes a line to create this directory when the log files are moved.


tommbendall:
Closing as this has been fixed by #445

PR #440, opened 2023-09-07
Diagnonally Implicit Multistage Class

atb1995:
This code change adds an Implicit Multistage class which allows for general diagonally implicit Runge-Kutta schemes by passing in a Butcher Tableau. It also adds the Implicit Midpoint and a two stage scheme developed by Qin and Zhang. The butcher tableau is of the form:

[a_00   0   .          0  ]
[a_10 a_11  .       0  ]
[  .        .     .          .  ]
[ b_0  b_1  .       b_s]

I have tested a Trapezium Rule version vs the regular TrapeziumRule for Williamson 1 & 6, and the errors are of the order 10^-9. QinZhang and ImplicitMidpoint also run nicely for Williamson 1 and 6.


gusto/time_discretisation.py
```diff
@@ -234,6 +234,128 @@ def apply(self, x_out, x_in):
...
 
+class ImplicitMultistage(TimeDiscretisation):
+    """
+    A class for implementing general diagonally implicit multistage (Runge-Kutta)
+    methods based on its Butcher tableau.
+
+    A Butcher tableau is formed in the following way for a s-th order
+    diagonally implicit scheme:
+
+    c_0 | a_00 a_01  .    a_0s
```
tommbendall, line 245: This looks really nice in the code. Unfortunately I suspect it will look rubbish in the documentation, which doesn't know that these need to be on new lines!

For the documentation to be created nicely, I would add a `\n` at the end of each line when you know you want a line break. For an example, see here: https://github.com/firedrakeproject/gusto/blob/main/gusto/equations.py#L579

You might also want to fix this for the `ExplicitMultistage` scheme (sorry!)

gusto/time_discretisation.py
```diff
@@ -234,6 +234,128 @@ def apply(self, x_out, x_in):
...
+      At iteration i (from 0 to s-1)
+        An intermediate location is computed as y_i = y^n + sum{over j less than or equal to i} (dt*a_ij*k_i)
+        Compute the gradient at the intermediate location, k_i = F(y_i)
+
+    At the last stage, compute the new solution by:
+    y^{n+1} = y^n + sum_{j from 0 to s} (dt*b_i*k_i)
+
+    """
+
+    def __init__(self, domain, field_name=None, solver_parameters=None, limiter=None, options=None, butcher_matrix=None):
```
tommbendall, line 276: I think `butcher_matrix` needs to be a compulsory argument here (and so the second argument), since I think this scheme won't work unless we've specified the Butcher matrix. I'm afraid the same thing again applies to the `ExplicitMultistage` scheme...

gusto/time_discretisation.py
```diff
@@ -234,6 +234,128 @@ def apply(self, x_out, x_in):
...
+    [  .    .   .       .  ]
+    [ b_0  b_1  .       b_s]
+
+    Unlike the explicit method, all upper diagonal a_ij elements are non-zero for implicit methods.
+
+    There are three steps to move from the current solution, y^n, to the new one, y^{n+1}
+
+    For an s stage method,
+      At iteration i (from 0 to s-1)
+        An intermediate location is computed as y_i = y^n + sum{over j less than or equal to i} (dt*a_ij*k_i)
```
tommbendall, line 268: This looks like you're trying to write pseudocode? Again I think this won't format too well in the documentation! We should also try to avoid making lines that are this long (in most of the code we stick to less than 80 characters per line). 

I'm afraid this all also the case for the `ExplicitMultistage` scheme

gusto/time_discretisation.py
```diff
@@ -234,6 +234,128 @@ def apply(self, x_out, x_in):
...
+    At the last stage, compute the new solution by:
+    y^{n+1} = y^n + sum_{j from 0 to s} (dt*b_i*k_i)
+
+    """
+
+    def __init__(self, domain, field_name=None, solver_parameters=None, limiter=None, options=None, butcher_matrix=None):
+        super().__init__(domain, field_name=field_name,
+                         solver_parameters=solver_parameters,
+                         limiter=limiter, options=options)
+        if butcher_matrix is not None:
```
tommbendall, line 280: I don't think this should be within an `if` statement when `butcher_matrix` is compulsory

gusto/time_discretisation.py
```diff
@@ -1341,3 +1465,40 @@ def apply(self, x_out, *x_in):
             self.x[n].assign(x_in[n])
         solver.solve()
         x_out.assign(self.x_out)
+
+
+class ImplicitMidpoint(ImplicitMultistage):
+    u"""
+    Implements the Implicit Midpoint method as a 1-stage Runge Kutta method.
+
+    The method, for solving
```
tommbendall, line 1474: I'm afraid more `\n`s are probably needed here, and in other schemes...

gusto/time_discretisation.py
```diff
@@ -1341,3 +1465,40 @@ def apply(self, x_out, *x_in):
...
+    ∂y/∂t = F(y), can be written as:
+
+    k0 = F[y^n + 0.5*dt*k0]
+    y^(n+1) = y^n + dt*k0
+    """
+    def __init__(self, domain, field_name=None, solver_parameters=None, limiter=None, options=None, butcher_matrix=None):
```
tommbendall, line 1480: I think `butcher_matrix` shouldn't be an argument here, since there is only one `butcher_matrix` for this scheme.

The same is true for the explicit schemes, e.g. `RK4`, `SSPRK3`, etc.

gusto/time_discretisation.py
```diff
@@ -1341,3 +1465,40 @@ def apply(self, x_out, *x_in):
...
+        super().__init__(domain, field_name,
+                         solver_parameters=solver_parameters,
+                         limiter=limiter, options=options)
+        self.butcher_matrix = np.array([[0.5], [1.]])
```
tommbendall, line 1484: It might be neater to define `butcher_matrix` above the `super().__init__` call, then pass it as an argument. (Again, same for the other schemes)

gusto/time_discretisation.py
tommbendall, line 1480: And another thing, unfortunately we need full docstrings here underneath the `__init__`, explaining what each of the arguments are (you can basically copy and paste from other places)

gusto/time_discretisation.py
tommbendall, line 276: We need some docstrings for this `__init__`, explaining what the arguments should be. You can basically copy and paste from `TimeDiscretisation`, but you need to add what `butcher_matrix` should be. The same is true for `ExplicitMultistage`

gusto/time_discretisation.py
```diff
@@ -1341,3 +1465,40 @@ def apply(self, x_out, *x_in):
...
+    Implements Qin and Zhang's two-stage, 2nd order, implicit Runge–Kutta method.
+
+    The method, for solving
+    ∂y/∂t = F(y), can be written as:
+
+    k0 = F[y^n + 0.25*dt*k0]
+    k1 = F[y^n + 0.5*dt*k0 + 0.25*dt*k1]
+    y^(n+1) = y^n + 0.5*dt*(k0 + k1)
+    """
+    def __init__(self, domain, field_name=None, solver_parameters=None, limiter=None, options=None, butcher_matrix=None):
```
tommbendall, line 1499: Same here about:
- `butcher_matrix` not an argument
- `__init__` needs a docstring
- the summary docstring for the class should have linebreak characters included
- `butcher_matrix` should be passed through to `super().__init__`

integration-tests/model/test_time_discretisation.py
```diff
@@ -27,8 +27,12 @@ def test_time_discretisation(tmpdir, scheme, tracer_setup):
 
     if scheme == "ssprk":
         transport_scheme = SSPRK3(domain)
-    elif scheme == "implicit_midpoint":
```
tommbendall, line 30: Thanks for fixing this!

tommbendall (changes requested):
Looks like it has implemented nicely and it's great to have `ImplicitMidpoint` through this method! I have a bunch of picky issues, which I'm afraid also apply to the `ExplicitMultistage` schemes (and I'm sorry for missing them the first time round...)

gusto/time_discretisation.py
atb1995: I have now done this one, separated out the butcher matrix comments as requested.

tommbendall (approved):
This looks great to me. Thanks for all the effort in making the documentation look nice!!

PR #445, opened 2023-09-19
Alternative parallel log fix

JDBetteridge:
Some cherry-picked commits from https://github.com/firedrakeproject/gusto/pull/417 to save waiting


tommbendall (changes requested):
This looks good to me, and is a better alternative to #438.

I am happy that it is better to have control over the creation of the directories, *before* the different logs are moved.

I have checked that this works locally with:
`env GUSTO_PARALLEL_LOG=FILE mpiexec -n 4 python3 examples/shallow_water/williamson_2.py`, and I also tried this example when picking up from a checkpoint just in case that might cause a problem.

My only suggestion is to include a couple of comments so that I can remember our logic.

tommbendall:
> This looks good to me, and is a better alternative to #438.
> 
> I am happy that it is better to have control over the creation of the directories, _before_ the different logs are moved.
> 
> I have checked that this works locally with: `env GUSTO_PARALLEL_LOG=FILE mpiexec -n 4 python3 examples/shallow_water/williamson_2.py`, and I also tried this example when picking up from a checkpoint just in case that might cause a problem.
> 
> My only suggestion is to include a couple of comments so that I can remember our logic.

Actually the picking up from checkpoint failed, but has only highlighted a separate parallel bug ...

tommbendall (approved):
Thanks Jack!

PR #449, opened 2023-09-22
Dual space updates to hybrid hydrostatic PC and precipitation diagnostic

tommbendall:
This fixes the hydrostatic balance routines and the precipitation diagnostic following updates to Firedrake's dual space code.


jshipton (approved):
Thanks for this @tommbendall !!

PR #450, opened 2023-09-25
Rewrite spaces again

tommbendall:
I have had another attempt at writing `function_spaces.py` and the `Spaces` object. In its current guise we can only one `"HDiv"` or `"HCurl"` space, but when using recovery options we might want more than one. The aim here was to facilitate that.

There are two main changes:
1. The building of the base and compatible spaces has been pulled out to a new `DeRhamComplex` object. The code here is essentially the same as before.
2. I have broken up the `Spaces.__call__` routine. Previously this was a single interface that (a) returned a field already created, (b) created a field otherwise or (c) added an existing field to the `Spaces` container. I have now separated these into three different routines. I am hoping that this makes the code much simpler.

I have also added in the BDMCE/BDMCF spaces...

### DIFF
--- a/gusto/function_spaces.py
+++ b/gusto/function_spaces.py
@@ -66,7 +66,16 @@ class Spaces(object):
         """
         self.mesh = mesh
         self.extruded_mesh = hasattr(mesh, "_base_mesh")
-        self._initialised_base_spaces = False
+
+        # Make dictionaries to store base elements for extruded meshes
+        # Here the keys are the horizontal and vertical degrees
+        self._initialised_base_spaces = {}
+        self.base_elt_hori_hdiv = {}
+        self.base_elt_hori_hcurl = {}
+        self.base_elt_hori_dg = {}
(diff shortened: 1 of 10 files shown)


gusto/function_spaces.py
```diff
@@ -66,11 +70,46 @@ def __init__(self, mesh):
...
         self.extruded_mesh = hasattr(mesh, "_base_mesh")
-        self._initialised_base_spaces = False
+        self.de_rham_complex = {}
 
-    def __call__(self, name, family=None, degree=None,
-                 horizontal_degree=None, vertical_degree=None,
-                 V=None, overwrite_space=False):
+    def __call__(self, name):
+        """
+        Returns a space, and also creates it if it is not created yet.
```
jshipton, line 77: docstring needs to change as space is no longer created here

jshipton (changes requested):
This all looks great, just one tiny change to the docstring.

jshipton (approved):

PR #451, opened 2023-10-06
Eady equations and problems

tommbendall:
The aim of this is to finally get the Eady equations/problems working again.

For now I've just re-added the code (from before FML was on main!), it won't work as many of the model objects have changed.

### DIFF
--- a/gusto/configuration.py
+++ b/gusto/configuration.py
@@ -6,9 +6,9 @@ from firedrake import sqrt, Constant
 
 __all__ = [
     "IntegrateByParts", "TransportEquationType", "OutputParameters",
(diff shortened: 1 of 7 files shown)


gusto/configuration.py
```diff
@@ -134,6 +135,33 @@ class ShallowWaterParameters(Configuration):
     H = None  # mean depth
 
 
+class EadyParameters(Configuration):
+
+    """
+    Physical parameters for Incompressible Eady
+    """
+    Nsq = 2.5e-05  # squared Brunt-Vaisala frequency (1/s)
```
colinjcotter, line 143: It's probably a bit dangerous to have N and Nsq both in the parameters with inconsistent values.

PR #452, opened 2023-10-10
Fix Diagnostics of Diagnostics

Witt-D:
I went through the thermodynamic diagnostics and added the self_setup_theromdyancs line where appropriate. 
Additionatlly I added Toms hack fix for why the potential energy is failing


examples/compressible/skamarock_klemp_nonhydrostatic.py
```diff
@@ -27,10 +27,10 @@
     tmax = dt
     dumpfreq = 1
 else:
-    nlayers = 10
-    columns = 150
-    tmax = 3600.
```
jshipton, line 32: Did you mean to add all of these changes?

examples/compressible/skamarock_klemp_nonhydrostatic.py
```diff
@@ -69,13 +69,11 @@
...
-        dumplist=['u'],
-    )
+        dump_nc=True,
+        dump_vtus=False
+          )
 
-diagnostic_fields = [CourantNumber(), Gradient("u"), Perturbation('theta'),
-                     Gradient("theta_perturbation"), Perturbation('rho'),
-                     RichardsonNumber("theta", parameters.g/Tsurf), Gradient("theta")]
+diagnostic_fields = [CourantNumber(), Pressure(eqns), SteadyStateError('Pressure_Vt')]
```
jshipton, line 76: Are any of the changes to this file meant to be added to this PR?

gusto/diagnostics.py
```diff
@@ -1033,12 +1033,25 @@ def setup(self, domain, state_fields):
             field1 = state_fields(self.field_name1)
             field2 = state_fields(self.field_name2, space=field1.function_space(),
                                   pick_up=True, dump=False)
-            # By default set this new field to the current value
-            # This may be overwritten if picking up from a checkpoint
-            field2.assign(field1)
+            self.init_field_set = False
```
jshipton, line 1036: Looks like we can get rid of `field2`  - is that right?

jshipton (changes requested):

examples/compressible/skamarock_klemp_nonhydrostatic.py
Witt-D: yes im not sure why the examples have been added as i was unaware that these had even been pushed from my local area?

tommbendall (changes requested):
Thanks for this Dan.

I've left a handful of comments, but actually I think there are some other things that we could do in this PR, that are harder to put in comments:
- we should add a `SteadyStateError` of a diagnostic to another test (maybe Williamson 2)
- we should add the `PotentialEnergy` diagnostic to an example to make sure that it's tested. Maybe the compressible DCMIP test?
- while you've fixed the `SteadyStateError` of a diagnostic, we haven't fixed the `Perturbation` of a diagnostic. I think we should include a fix to that in this PR too.
- a minor thing, but maybe you could update the name of the PR? I think it's more like "Fix diagnostics of diagnostics" -- you might have started by trying to making a minimum failing example but we don't want failing examples on main!

examples/compressible/skamarock_klemp_nonhydrostatic.py
```diff
@@ -44,22 +44,22 @@
...
+dirname = 'testing'
 
 # Dumping point data using legacy PointDataOutput is not supported in parallel
 if COMM_WORLD.size == 1:
     output = OutputParameters(
         dirname=dirname,
         dumpfreq=dumpfreq,
-        pddumpfreq=dumpfreq,
-        dumplist=['u'],
-        point_data=[('theta_perturbation', points)],
```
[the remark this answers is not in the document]
Witt-D: So i had no intention of pushing the change to this example and will move to testing through the examples suggested above. For the reference when including point data the examples wouldn't run so will most likely revert a lot of the changes to skamarock and move to testing in Williams 2 and dcmip etc

examples/compressible/skamarock_klemp_nonhydrostatic.py
Witt-D: Removed the changes to the skamarock test and implemented it in Williamson2

examples/compressible/skamarock_klemp_nonhydrostatic.py
```diff
@@ -44,22 +44,24 @@
 # Equation
 Tsurf = 300.
 parameters = CompressibleParameters()
-eqns = CompressibleEulerEquations(domain, parameters)
+eqns = CompressibleEulerEquations(domain, parameters, u_transport_option='vector_advection_form')
```
tommbendall, line 47: I don't think we should change the equation here

examples/compressible/skamarock_klemp_nonhydrostatic.py
```diff
@@ -44,22 +44,24 @@
...
+print(f'Number of DOFs = {eqns.X.function_space().dim()}')
```
tommbendall, line 48: This print statement needs removing!

examples/compressible/skamarock_klemp_nonhydrostatic.py
```diff
@@ -44,22 +44,24 @@
...
 
 # I/O
 points_x = np.linspace(0., L, 100)
 points_z = [H/2.]
 points = np.array([p for p in itertools.product(points_x, points_z)])
-dirname = 'skamarock_klemp_nonlinear'
+dirname = 'testing'
```
tommbendall, line 54: I don't think you mean to change the `dirname` here!

examples/compressible/skamarock_klemp_nonhydrostatic.py
```diff
@@ -69,13 +71,13 @@
     output = OutputParameters(
         dirname=dirname,
         dumpfreq=dumpfreq,
-        pddumpfreq=dumpfreq,
-        dumplist=['u'],
+        dump_nc=True,
+        dump_vtus=False
```
tommbendall, line 75: I think I'd prefer for these changes not to be made here to avoid confusing things, so it'd be good to revert back to the previous outputting settings

examples/compressible/skamarock_klemp_nonhydrostatic.py
```diff
@@ -44,22 +44,24 @@
...
+dirname = 'testing'
 
 # Dumping point data using legacy PointDataOutput is not supported in parallel
 if COMM_WORLD.size == 1:
     output = OutputParameters(
         dirname=dirname,
         dumpfreq=dumpfreq,
-        pddumpfreq=dumpfreq,
+        dump_nc=True,
+        dump_vtus=True,
```
tommbendall, line 62: I think I'd prefer for these changes not to be made here to avoid confusing things, so it'd be good to revert back to the previous outputting settings

gusto/diagnostics.py
```diff
@@ -1033,12 +1033,30 @@ def setup(self, domain, state_fields):
             field1 = state_fields(self.field_name1)
             field2 = state_fields(self.field_name2, space=field1.function_space(),
                                   pick_up=True, dump=False)
-            # By default set this new field to the current value
-            # This may be overwritten if picking up from a checkpoint
```
tommbendall, line 1037: I think we still want these deleted comments in here...
and this raises the point that we should make sure to comment our changes clearly so that we can still understand this change in 6 months time...

I suggest adding something like (you can use your own words!):
```
# The initial value for this field may not have been set yet, so we should postpone setting
# the initial field until the first time that the compute method is called
```

gusto/diagnostics.py
```diff
@@ -1033,12 +1033,30 @@ def setup(self, domain, state_fields):
...
+            field2 = state_fields(self.field_name2, space=field1.function_space(),
+                                  pick_up=True, dump=False)
             field2.assign(field1)
+            self.state_fields = state_fields
+            self.init_field_set = True
 
         super().setup(domain, state_fields)
 
+    def compute(self):
+        if not self.init_field_set:
```
tommbendall, line 1050: Add comment here along the lines of:
```
# The first time the compute method is called, we should ensure the initial field is set
# However we don't want to do this if picking up from a checkpoint
```

gusto/io.py
```diff
@@ -630,6 +630,10 @@ def pick_up_from_checkpoint(self, state_fields):
             self.output.checkpoint_pickup_filename = None
         else:
             raise ValueError("Must set checkpoint True if picking up")
+        # Prevent any steady-state diagnostics overwriting their original fields
```
tommbendall, line 633: Very picky, but it would be good to have a blank line above this comment to help illustrate that the next segment of code is doing something separate from the above bits of code

tommbendall (changes requested):
Thanks Dan! Good that the tests now pass.

A few more minor things to sort.

Do we also want to fix the `Perturbation` diagnostic as part of this PR?

PS: you might want to fix the typo in the name of the pull request!

gusto/io.py
Witt-D: So ive addressed the points regarding the example and some of the comments but was a bit uncertain on my explaining comments so they will be worth checking.

I think including perturbation in this will be a good idea too so will start having a dig around in that now

gusto/diagnostics.py
```diff
@@ -1032,21 +1032,27 @@ def setup(self, domain, state_fields):
         if not hasattr(state_fields, self.field_name2):
             field1 = state_fields(self.field_name1)
             field2 = state_fields(self.field_name2, space=field1.function_space(),
-                                  pick_up=True, dump=False)
+                                  pick_up=True, dump=False)         
             # Attach state fields to self so that we can pick it up in compute
             self.state_fields = state_fields
+            # The initial value for fields may not have alreay been set yet so we
```
tommbendall, line 1038:
```suggestion
            # The initial value for fields may not have already been set yet so we
```

gusto/diagnostics.py
```diff
@@ -1032,21 +1032,27 @@ def setup(self, domain, state_fields):
...
+            # postpone setting it untill the compute method is called
```
tommbendall, line 1039:
```suggestion
            # postpone setting it until the compute method is called
```

gusto/diagnostics.py
```diff
@@ -1032,21 +1032,27 @@ def setup(self, domain, state_fields):
...
+            # This may be overwritten if picking up from a checkpoint 
             field2.assign(field1)
             self.state_fields = state_fields
             self.init_field_set = True
 
         super().setup(domain, state_fields)
 
     def compute(self):
+        # The first time the compute method is called we set the initial field.
+        # We do not want to do this is picking up from a checkpoint
```
tommbendall, line 1055:
```suggestion
        # We do not want to do this if picking up from a checkpoint
```

tommbendall (commented):
Some typos to fix

gusto/timeloop.py
```diff
@@ -216,6 +216,7 @@ def set_reference_profiles(self, reference_profiles):
             set the reference field.
         """
         for field_name, profile in reference_profiles:
+            breakpoint()
```
jshipton, line 219: this is why you're breaking the tests

gusto/timeloop.py
```diff
@@ -228,16 +228,19 @@ def set_reference_profiles(self, reference_profiles):
...
-
+                if field_name in self.equation.field_names:
+                    idx = self.equation.field_names.index(field_name)
+                    X_ref = self.equation.X_ref.subfunctions[idx]
+                    X_ref.assign(ref)
+                else:
+                    # reference profile of a diagnostic
+                    # warn user in case they made a typo
+                    logger.warning(f'Setting reference profile for diagnostic {field_name}')
+                    # dont need to do anythin else as value in field container has already been set
```
tommbendall, line 243: Could you fix the typos in this comment?
```suggestion
                    # Don't need to do anything else as value in field container has already been set
```

tommbendall (changes requested):
This is very very nearly ready, you just need to fix the typos in a comment

tommbendall (approved):
Thanks for doing this Dan! Good to go now

PR #453, opened 2023-10-19
IMEX Multistage

atb1995:
This code change adds IMEX Multistage class which allows for general split diagonally implicit and explicit Runge-Kutta schemes by passing in a pair of Butcher Tableaus. It also adds IMEX Euler, ARK2, Trap2, SSP3 and ARS3.  The butcher tableaus are of the form:

[a_00 0 . 0 ]
[a_10 a_11 . 0 ]
[ . . . . ]
[ b_0 b_1 . b_s]

and 

[0  0 . 0 ]
[d_10 0 . 0 ]
[ . . . . ]
[ e_0 e_1 . e_s]

I ran convergence tests for Williamson 5 & 6 (see attached). This pull request is blocked by #440.


atb1995:
w6_imex_rk
s://github.com/firedrakeproject/gusto/assets/81297297/10d864db-7d19-4f48-975d-19150c9613b4)

integration-tests/model/test_time_discretisation.py
```diff
@@ -8,8 +8,9 @@ def run(timestepper, tmax, f_end):
     return norm(timestepper.fields("f") - f_end) / norm(f_end)
 
 
-@pytest.mark.parametrize("scheme", ["ssprk", "implicit_midpoint",
-                                    "RK4", "Heun", "BDF2", "TR_BDF2", "AdamsBashforth", "Leapfrog", "AdamsMoulton"])
+@pytest.mark.parametrize("scheme", ["ssprk", "TrapeziumRule",
+                                    "RK4", "Heun", "BDF2", "TR_BDF2", "AdamsBashforth",
```
tommbendall, line 12: Should we add some of the IMEX schemes here to test them?

gusto/time_discretisation.py
```diff
@@ -238,6 +239,202 @@ def apply(self, x_out, x_in):
...
+
+    def res(self, stage):
+        mass_form = self.residual.label_map(
+            lambda t: t.has_label(time_derivative),
+            map_if_false=drop)
+        residual = mass_form.label_map(all_terms,
+                                       map_if_true=replace_subject(self.x_out, old_idx=self.idx))
+        residual -= mass_form.label_map(all_terms,
+                                        map_if_true=replace_subject(self.x1, old_idx=self.idx))
+        for i in range(stage):
```
tommbendall, line 354: I can follow what is going on in this for loop but maybe you could add some comments to explain it?

gusto/time_discretisation.py
```diff
@@ -238,6 +239,202 @@ def apply(self, x_out, x_in):
...
+        """Set up the discretisation's left hand side (the time derivative)."""
+        return super(IMEXMultistage, self).lhs
+
+    @cached_property
+    def rhs(self):
+        """Set up the discretisation's right hand side (the time derivative)."""
+        return super(IMEXMultistage, self).rhs
+
+    def res(self, stage):
+        mass_form = self.residual.label_map(
```
tommbendall, line 347: Could you add a docstring here to explain this routine?

gusto/time_discretisation.py
```diff
@@ -238,6 +239,202 @@ def apply(self, x_out, x_in):
...
+            map_if_false=drop)
+        r_imp = r_imp.label_map(
+            lambda t: t.has_label(time_derivative),
+            map_if_false=lambda t: Constant(self.butcher_imp[stage, stage])*self.dt*t)
+        residual += r_imp
+        return residual.form
+
+    @property
+    def final_res(self):
+        mass_form = self.residual.label_map(lambda t: t.has_label(time_derivative),
```
tommbendall, line 383: I can't spot what is different about this compared with the normal `res` routine!

tommbendall (changes requested):
Looks very exciting, thanks Alex. My suggestions are mainly about adding comments

integration-tests/model/test_time_discretisation.py
atb1995: The issue here is the test is a prescribed transport test, hence no option for implicit and explicit terms.. Could change or add another test?

gusto/time_discretisation.py
```diff
@@ -238,6 +239,202 @@ def apply(self, x_out, x_in):
...
+                a butcher tableau defining a given implicit Runge Kutta time discretisation.
+            butcher_exp (numpy array): A matrix containing the coefficients of
+                a butcher tableau defining a given explicit Runge Kutta time discretisation.
+            field_name (str, optional): name of the field to be evolved.
+                Defaults to None.
+            subcycles (int, optional): the number of sub-steps to perform.
+                Defaults to None.
+            solver_parameters (dict, optional): dictionary of parameters to
+                pass to the underlying solver. Defaults to None.
+            limiter (:class:`Limiter` object, optional): a limiter to apply to
```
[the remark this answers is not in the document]
atb1995: It does not, I have removed!

gusto/time_discretisation.py
```diff
@@ -238,6 +239,202 @@ def apply(self, x_out, x_in):
...
+        Args:
+            equation (:class:`PrognosticEquation`): the model's equation.
+            *active_labels (:class:`Label`): labels indicating which terms of
+                the equation to include.
+        """
+
+        super().setup(equation, apply_bcs, *active_labels)
+
+        self.residual = self.residual.label_map(
+            lambda t: any(t.has_label(time_derivative, transport)),
```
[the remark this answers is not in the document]
atb1995: Yes, it is a map if false, so all non-time derivative and transport terms are implicit

gusto/time_discretisation.py
```diff
@@ -238,6 +239,202 @@ def apply(self, x_out, x_in):
...
+            map_if_false=lambda t: implicit(t))
+
+        self.residual = self.residual.label_map(
```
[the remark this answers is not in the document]
atb1995: This is a good point, it could be left up to the user to label terms as explicit or implicit at the script level.. What do you think?

gusto/time_discretisation.py
atb1995: Good point! Comments added

gusto/time_discretisation.py
atb1995: I will do! Will also add for final_res and the various solvers

gusto/time_discretisation.py
atb1995: The final residual loops across all stages, and also loops up to the final stage for explicit terms.

gusto/time_discretisation.py
```diff
@@ -238,6 +239,202 @@ def apply(self, x_out, x_in):
...
+        Args:
+            domain (:class:`Domain`): the model's domain object, containing the
+                mesh and the compatible function spaces.
+            butcher_imp (numpy array): A matrix containing the coefficients of
+                a butcher tableau defining a given implicit Runge Kutta time discretisation.
+            butcher_exp (numpy array): A matrix containing the coefficients of
+                a butcher tableau defining a given explicit Runge Kutta time discretisation.
+            field_name (str, optional): name of the field to be evolved.
+                Defaults to None.
+            subcycles (int, optional): the number of sub-steps to perform.
```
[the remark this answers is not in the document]
atb1995: Removed!

gusto/time_discretisation.py
```diff
@@ -323,13 +323,37 @@ def setup(self, equation, apply_bcs=True, *active_labels):
...
+                    # uadv = split(self.equation.X)[u_idx]
+                    # breakpoint()
+                    uadv = t.get(transporting_velocity)
+                    breakpoint()
+                    new_transport_term = prognostic(subject(advection_form(test, transported_field, uadv) + test*transported_field*div(uadv)*dx, subj, prognostic_field_name))
+                    # Add onto residual and drop old term
+                    self.residual = self.residual.label_map(
+                        lambda t: t.get(transport) == TransportEquationType.conservative,
+                        map_if_true=drop)
+                    self.residual += new_transport_term.form
```
jshipton, line 345: you should just add the labelled form to the residual, not just the form part, i.e. `self.residual += new_transport_term`

gusto/time_discretisation.py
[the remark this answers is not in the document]
tommbendall: We've discussed this offline: we realised that we can label terms as `explicit` or `implicit` in any example file using the `equation.label_map(...)` routine.

But here we should have a check that throws an error if a term hasn't been labelled as implicit/explicit (or is the `time_derivative`)

gusto/time_discretisation.py
```diff
@@ -323,13 +323,37 @@ def setup(self, equation, apply_bcs=True, *active_labels):
 
         super().setup(equation, apply_bcs, *active_labels)
 
+        # Get continuity form transport term
+        for t in self.residual:
```
tommbendall, line 333: As we discussed offline: it might be best to move this code to break up the continuity form out of time discretistations to a new routine (we suggested that it would be in `common_forms.py`). Then a `TimeDiscretisation` can be agnostic to the terms that it works on, and this code than then be reused for other things.

integration-tests/model/test_time_discretisation.py
tommbendall: We discussed offline: adding a new `test_imex.py` which could test these on something simple like a wave equation with an analytic solution

tommbendall (changes requested):
Sorry to be asking for more changes!

integration-tests/model/test_time_discretisation.py
atb1995: I've added test_IMEX.py. It uses a continuity form and splits the term to treat q*div(u) implicitly and u.grad(q) explicitly.

gusto/time_discretisation.py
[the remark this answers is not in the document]
atb1995: An error checker has been added. I raised a "NotImplementedError" - is this the correct choice..?

gusto/common_forms.py
```diff
@@ -194,3 +198,65 @@ def diffusion_form(test, q, kappa):
...
+
+    for t in equation.residual:
+        if (t.get(transport) == TransportEquationType.conservative):
+            # Get fields and test functions
+            subj = t.get(subject)
+            prognostic_field_name = t.get(prognostic)
+            if hasattr(equation, "field_names"):
+                idx = equation.field_names.index(prognostic_field_name)
+            else:
+                idx = equation.field_name.index(prognostic_field_name)
```
tommbendall, line 226: I think this won't give what we expect! Is `field_name` a string and not a list?

In the case that we don't have `field_names`, we don't want an index and don't want to split anything up (because we won't have a mixed function space)

tommbendall (changes requested):
Test looks really good, just some questions about what the `split_continuity_form` routine does when we don't have a mixed function space

gusto/common_forms.py
```diff
@@ -194,3 +198,67 @@ def diffusion_form(test, q, kappa):
     form = -inner(test, div(kappa*grad(q)))*dx
 
     return diffusion(form)
+
+
+def split_continuity_form(equation):
+    u"""
+    Loops through terms in a given equation, and splits continuity terms into
```
tommbendall, line 205: Could you mention that this will split up *all* conservative terms in the equation?

gusto/common_forms.py
```diff
@@ -194,3 +198,67 @@ def diffusion_form(test, q, kappa):
...
+    advective and divergence terms.
+
+    This describes splitting ∇.(u*q) into u.∇q and q(∇.u),
+    for transporting velocity u and transported q.
+
+    Args:
+        equation (:class:`PrognosticEquation`): the model's equation.
+
+    Returns:
+        equation (:class:`PrognosticEquation`): the model's equation.
```
tommbendall, line 215: The format for the docstrings is a bit different for the Returns section. I think this should be:
```suggestion
        :class:`PrognosticEquation`: the model's equation.
```

tommbendall (changes requested):
Two tiny changes and then I think that this is ready!

gusto/common_forms.py
atb1995: Done! Thanks Tom

tommbendall (approved):

jshipton:
[celebrate]     Shipton, Jemma reacted to your message:

PR #457, opened 2023-11-02
Terminator toy

ta440:
This adds the physics scheme for the Terminator Toy problem, which tests the interaction of two species, defined through mixing ratios, at the same time that a density field is being advected. A script to run this test this is provided in the case_studies repository. As this physics scheme needs to be solved with Backwards Euler, due to very sharp gradients in the ODEs governing the chemical interaction, the physics and dynamics need to be solved separately. Additionally, we prescribe the transporting velocity (a Nair Lauritzen setup) instead of having the velocity as a prognostic variable. A new SplitPrescribedTransport timestepper is written to enable this.


gusto/timeloop.py
```diff
@@ -378,6 +377,89 @@ def timestep(self):
...
+                methods to use for discretising transport or diffusion terms
+                for each transported/diffused variable. Defaults to None,
+                in which case the terms follow the original discretisation in
+                the equation.
+            physics_schemes: (list, optional): a list of tuples of the form
+                (:class:`PhysicsParametrisation`, :class:`TimeDiscretisation`),
+                pairing physics parametrisations and timestepping schemes to use
+                for each. Timestepping schemes for physics can be explicit
+                or implicit. Defaults to None.
+            prescribed_transporting_velocity: (field, optional): A known
```
tommbendall, line 407: Could you add a bit more description to this to explain that it can be a python function, taking time as an argument?

I see that the docstrings are wrong for the `PrescribedTransport` stepper, sorry about that!

gusto/physics.py
```diff
@@ -1751,3 +1751,72 @@ def evaluate(self, x_in, dt):
...
+    def evaluate(self, x_in, dt):
+        """
+        Evaluates the source/sink for the coalescence process.
+
+        Args:
+            x_in (:class:`Function`): the (mixed) field to be evolved.
+            dt (:class:`Constant`): the time interval for the scheme.
+        """
+
+        logger.info(f'Evaluating physics parametrisation {self.label.label}')
```
tommbendall, line 1822: Maybe add a `pass` underneath this to make it obvious that this evaluate method does nothing

gusto/physics.py
```diff
@@ -1751,3 +1751,72 @@ def evaluate(self, x_in, dt):
...
+            species1_name(str, optional): Name of the first interacting species. Defaults
+              to 'X'.
+            species2_name(str, optional): Name of the second interacting species. Defaults
+              to 'X2'.
+        """
+
+        label_name = 'terminator_toy'
+        super().__init__(equation, label_name, parameters=None)
+
+        assert species1_name in equation.field_names, f"Field {species1_name} does not exist in the equation set"
```
tommbendall, line 1788: Very picky point here. It might be slightly better if we change this error (and the one below) to something like:
```
if species1_name not in equation.field_names:
    raise ValueError(...)
```
as this is the normal type of python error when the wrong argument is given to a routine

tommbendall (changes requested):
This looks really good Tim, thanks! Just three really minor comments

gusto/physics.py
ta440: Cool, I've done that for the two name checks. Is the check for the function spaces ok as it is, as this is not an argument error, but a compatibility error?

tommbendall: I think the check is OK as is, as you say it is not about having an incorrect argument but that the model set up isn't correct

tommbendall (approved):
Thanks very much Tim! We've come up with a plan of follow-up activities for this to get the new things tested, that I'll post here for posterity:
- a test for the `TerminatorToy`
- add `SplitPrescribedTransport` as an parametrized option in `test_prescribed_transport.py`
- use `SplitPrescribedTransport` in one of the physics tests, maybe `test_precipitation.py`

PR #459, opened 2023-11-14
Mixed fs limiter

ta440:
This pull request adds the ability to apply individual limiters to specific prognostic variables. It does so through a MixedFSLimiter object in gusto/limiters.py which stores a dictionary with user defined sublimiters for certain prognostic variables. A test is of this new functionality is added to the integration-tests/transport/test_limiters.py file.


gusto/limiters.py
```diff
@@ -178,3 +178,35 @@ def apply(self, field):
...
+
+class MixedFSLimiter(object):
+    """An object to hold a dictionary that defines limiters for transported prognostic
+    variables. Different limiters may be applied to different fields and not every transported variable needs a defined limiter.
+    """
+
+    def __init__(self, equation, sublimiters):
+        """
+        Args:
+            equation (:class: 'PrognosticEquationSet'): the prognostic equation(s)
```
tommbendall, line 191:
```suggestion
            equation (:class:`PrognosticEquationSet`): the prognostic equation(s)
```
I think we need a back quotation here and not forward ones!

tommbendall (changes requested):
Thanks Tim, this looks really good and very neatly implemented.

I think the choice of DG1 and DG1 equispaced for the different spaces in the test makes sense, and it's good to demonstrate that this will in theory work when the tracers are in different spaces.

I have two very very trivial suggestions about docstrings, then this will be good to go on

tommbendall (approved):
Looks great Tim, thanks and sorry it's taken a while to get this approved!

PR #464, opened 2023-11-15
Fix issues from update to ufl (deleting brackets!)

tommbendall:
This fixes things that were broken by changes to FInAT and UFL. It:
- deletes some brackets from methods called from `space.ufl_element()`, as these now seem to be properties.
- changes the degree 1 Vtheta limiter as the DoF ordering changed
- updates the KGO files for some tests given this


jshipton (approved):
well done for tracking all those down @tommbendall !!

PR #465, opened 2023-11-21
Tests for Terminator Toy and SplitPrescribedTransport

ta440:
This pull request follows up on PR#457 to add the following tests:

1. A test for the Terminator Toy physics scheme. The aim of this test is to ensure that the coupled ODEs for the species interaction are being solved as expected. As such, the test takes very large time steps in a case with no flow, starting from X(0) = 4e-6, X2(0)=0. The analytical steady state for this case is given in the Lauritzen paper:
steady_state_lauritzen_picture
An error norm is computed relative to this analytical solution. As we aren't running sufficiently long to reach this state, the test ensures that the system is moving towards this state and that the normalised difference between the numerical and steady-state fields is less than 0.25.

2. A test of the prescribed wind velocity feature of SplitPrescribedTransport. This is implemented as an additional parameterised option in the test_prescribed_transport.py file.

3. Using SplitPrescribedTransport within the test_precipitation physics test.


integration-tests/physics/test_wind_drag.py
```diff
@@ -37,7 +37,7 @@ def run_wind_drag(dirname, implicit_formulation):
     eqn = CompressibleEulerEquations(domain, parameters)
 
     # I/O
-    output = OutputParameters(dirname=dirname+"/surface_fluxes",
+    output = OutputParameters(dirname=dirname+"/wind_drag",
```
tommbendall, line 40: Thanks for fixing this!

integration-tests/physics/test_terminator_toy.py
```diff
@@ -0,0 +1,112 @@
...
+    lamda_c = -np.pi/3.
+
+    k1 = max_value(0, sin(theta)*sin(theta_c) + cos(theta)*cos(theta_c)*cos(lamda-lamda_c))
+    k2 = 1
+
+    physics_schemes = [(TerminatorToy(eqn, k1=k1, k2=k2, species1_name='X',
+                        species2_name='X2'), BackwardEuler(domain))]
+
+    # Set up a non-divergent, time-varying, velocity field
+    def u_t(t):
```
tommbendall, line 66: Since the wind is zero, do we need a function prescribing the velocity? Or could we just set the initial velocity to zero?

integration-tests/physics/test_terminator_toy.py
```diff
@@ -0,0 +1,112 @@
...
+    # ------------------------------------------------------------------------ #
+
+    # A much larger timestep than in proper simulations, as this
+    # tests moving towards a steady state with no flow.
+    dt = 100000.
+
+    # Make the mesh and domain
+    R = 6371220.
+    mesh = IcosahedralSphereMesh(radius=R,
+                                 refinement_level=3, degree=2)
```
tommbendall, line 26: As this is just a test and the spatial discretisation is not important, could we use a coarser grid? e.g. refinement level 2?

integration-tests/physics/test_terminator_toy.py
```diff
@@ -0,0 +1,112 @@
...
+    return stepper, X_steady, X2_steady
+
+
+def test_terminator_toy_setup(tmpdir):
+    dirname = str(tmpdir)
+    stepper, X_steady, X2_steady = run_terminator_toy(dirname)
+    X_field = stepper.fields("X")
+    X2_field = stepper.fields("X2")
+
+    # Assert that the physics scheme has sufficiently moved
```
tommbendall, line 109: This seems like a sensible test to me!

tommbendall (changes requested):
Thanks again Tim for doing this.

I think the terminator toy test looks sensible. I had a couple of suggestions to slightly simplify it. The other changes look good.

integration-tests/physics/test_terminator_toy.py
ta440: That's a good point. Using a refinement level of 2 still successfully tests the ODEs moving the system towards the steady state, although I will need to halve the timestep size and slacken the tolerance on the test (error less than 0.4) to adjust.

integration-tests/physics/test_terminator_toy.py
ta440: I can't find a way to get the prescribed_transporting_velocity to work without giving it a function.

tommbendall (approved):
As everything is passing I think this is now ready! Thanks Tim

PR #468, opened 2023-12-07
Fix minor broken tests on main

tommbendall:
This fixes the two tests broken on main (presumably from recent Firedrake updates):
- by slightly slackening a tolerance in one test
- updating the KGO for the linear SW wave test


jshipton (approved):
Thanks @tommbendall

PR #470, opened 2023-12-21
Include b in the list of fields to linearise in the ShallowWaterEquations

nhartney:
This pull request fixes a bug that meant that the buoyancy variable `b` in the thermal shallow water equations was being treated as an active tracer. 
It adds b to the list of fields in the linearisation in the `ShallowWaterEquations` class so that `b` is recognised as a prognostic field.


tommbendall (approved):
Good spot! This looks right to me.

PR #471, opened 2024-01-09
Wrapper of options

ta440:
This adds the ability to apply the different wrapper options of EmbeddedDG and Recovery to different tracers/prognostic variables in the transport step. This will enable tracers to live in different spaces and will fix a current compatibility issue where a DG and theta tracer can't both be used within the CoupledTransportEquation with respective sublimiters, as the ThetaLimiter requires the use of EmbeddedDG.
 
This is implemented by defining a new dictionary wrapper 'MixedOptions' which will store the individual wrapper options of the transported variables. Tracers have different options, or may have none defined at all. Currently, SUPG has not been implemented as an option for a tracer.

A new test file, 'test_mixed_fs_options.py' is made to test combinations of these different options for two tracers solved with a CoupledTransportEquation. Limiters are also used on all these tests. As this new script includes the DG1-DG1_equispaced combination, this test has been removed from the 'test_limiters.py' file.


gusto/time_discretisation.py
```diff
@@ -87,16 +87,32 @@ def __init__(self, domain, field_name=None, solver_parameters=None,
...
         if options is not None:
-            self.wrapper_name = options.name
-            if self.wrapper_name == "embedded_dg":
-                self.wrapper = EmbeddedDGWrapper(self, options)
-            elif self.wrapper_name == "recovered":
-                self.wrapper = RecoveryWrapper(self, options)
-            elif self.wrapper_name == "supg":
-                self.wrapper = SUPGWrapper(self, options)
+            if type(options) == MixedOptions:
+                self.wrapper = options
```
tommbendall, line 91: Not that what you've done isn't a sensible way of implementing this, but it's occurred to me that it isn't quite consistent with how the existing wrappers work.

At the moment we have two broad types of objects: `Options` and `Wrappers`, and I think what we've done here is mix up the two a bit.

So now I'm thinking the best way of mirroring this structure for mixed function spaces would be to have:
1. `MixedFSOptions` which is defined in `configuration.py` and is essentially a glorified dictionary
2. `MixedFSWrapper` which is defined in `wrappers.py` and has the same methods as other `Wrapper` objects

gusto/wrappers.py
```diff
@@ -161,13 +171,19 @@ class RecoveryWrapper(Wrapper):
     def setup(self):
         """Sets up function spaces and fields needed for this wrapper."""
 
+        print(self.options)
+
         assert isinstance(self.options, RecoveryOptions), \
-            'Embedded DG wrapper can only be used with Recovery Options'
+            'Recovery wrapper can only be used with Recovery Options'
```
tommbendall, line 177: Thanks for correcting this!

gusto/wrappers.py
```diff
@@ -361,3 +384,73 @@ def label_terms(self, residual):
...
+            x_in_sub = self.x_in.subfunctions[field_idx]
+
+            if field_name in self.subwrappers:
+                subwrapper = self.subwrappers[field_name]
+                subwrapper.pre_apply(field)
+                x_in_sub.assign(subwrapper.x_in)
+            else:
+                x_in_sub.assign(field)
+
+    def post_apply(self, x_out):
```
tommbendall, line 440: This method looks exactly right to me

gusto/wrappers.py
```diff
@@ -361,3 +384,73 @@ def label_terms(self, residual):
...
+
+    def setup(self):
+        """ Compute the new mixed function space from the subwrappers """
+
+        self.function_space = MixedFunctionSpace(self.wrapper_spaces)
+        self.x_in = Function(self.function_space)
+        self.x_out = Function(self.function_space)
+
+    def pre_apply(self, x_in):
+        """
```
tommbendall, line 423: This method looks exactly right to me

gusto/wrappers.py
```diff
@@ -33,6 +33,9 @@ def __init__(self, time_discretisation, wrapper_options):
         self.time_discretisation = time_discretisation
         self.options = wrapper_options
         self.solver_parameters = None
+        self.idx = None
+        self.mixed_options = False
```
tommbendall, line 37: I'm a little bit uncertain about having a general `Wrapper` object know about `mixed_options`...

It feels like the inheritance of properties could be the wrong way around, as a `MixedFSWrapper` has several subwrappers, but the subwrappers then need to know that they are subwrapper. Ideally it would be better if they don't know that they are subwrappers!

Looking below at why you've had to do this, it's to set the function space. I'm just wondering if we can find a cleaner way to set `original_space`? Maybe `idx` could be an argument to the `__init__` or `setup` method of the base `Wrapper` class?

tommbendall (changes requested):
Thanks very much for doing this Tim, I think this will enable us to lots of exciting things!

I think the methods for the `MixedOptions` wrapper look perfect and I think it's the right thing to have a new test, which also looks good.

However (as I've written in a detailed comment) I think we actually might want to separate the new object into two objects:
1. `MixedFSOptions` to live in `configuration.py`
2. `MixedFSWrapper` to live in `wrappers.py` and would be very close to what you've implemented already

The reason for this is that it keeps the code structure used for the existing `Options`/`Wrappers`.

My other thought (as I've written in another comment) is about whether we can avoid having the `mixed_options` property for wrapper objects?

I'm very happy to discuss these things offline.

gusto/wrappers.py
```diff
@@ -161,13 +169,19 @@ class RecoveryWrapper(Wrapper):
     def setup(self):
         """Sets up function spaces and fields needed for this wrapper."""
 
+        print(self.options)
```
tommbendall, line 172: This just needs removing!

gusto/wrappers.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, time_discretisation, wrapper_options):
         self.time_discretisation = time_discretisation
         self.options = wrapper_options
         self.solver_parameters = None
+        self.tracer_fs = None
```
tommbendall, line 36:
```suggestion
        self.original_space = None
```

gusto/time_discretisation.py
```diff
@@ -159,21 +173,52 @@ def setup(self, equation, apply_bcs=True, *active_labels):
...
+                    if field not in equation.field_names:
+                        raise ValueError(f"The option defined for {field} is for a field that does not exist in the equation set")
+
+                    field_idx = equation.field_names.index(field)
+
+                    # Store the original space of the tracer
+                    subwrapper.tracer_fs = self.equation.spaces[field_idx]
 
-            self.residual = self.wrapper.label_terms(self.residual)
+                    subwrapper.setup()
```
tommbendall, line 191:
```suggestion
                    subwrapper.setup(self.equation.spaces[field_idx])
```

gusto/time_discretisation.py
```diff
@@ -159,21 +173,52 @@ def setup(self, equation, apply_bcs=True, *active_labels):
...
+
+                for field, subwrapper in self.wrapper.subwrappers.items():
+
+                    if field not in equation.field_names:
+                        raise ValueError(f"The option defined for {field} is for a field that does not exist in the equation set")
+
+                    field_idx = equation.field_names.index(field)
+
+                    # Store the original space of the tracer
+                    subwrapper.tracer_fs = self.equation.spaces[field_idx]
```
tommbendall, line 189:
```suggestion
```

gusto/time_discretisation.py
```diff
@@ -159,21 +173,52 @@ def setup(self, equation, apply_bcs=True, *active_labels):
...
+                new_test_mixed = TestFunctions(self.fs)
+
+                # Replace the original test function with one from the new
+                # function space defined by the subwrappers
+                self.residual = self.residual.label_map(
+                    all_terms,
+                    map_if_true=replace_test_function(new_test_mixed))
+
+            else:
+                self.wrapper.setup()
```
tommbendall, line 207:
```suggestion
                self.wrapper.setup(self.fs)
```
I think this should be `self.fs` but I'm not sure...? Maybe it is just `None`

gusto/wrappers.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, time_discretisation, wrapper_options):
...
 
     @abstractmethod
     def setup(self):
```
tommbendall, line 39: Add `original_space` as an argument here

gusto/wrappers.py
```diff
@@ -82,10 +83,14 @@ def setup(self):
...
 
-        original_space = self.time_discretisation.fs
         domain = self.time_discretisation.domain
         equation = self.time_discretisation.equation
 
+        if self.tracer_fs is not None:
+            original_space = self.tracer_fs
+        else:
+            original_space = self.time_discretisation.fs
+
```
tommbendall, line 93: I think this snippet changes to be:
`self.original_space = original_space`

gusto/wrappers.py
```diff
@@ -104,7 +109,10 @@ def setup(self):
 
         self.x_in = Function(self.function_space)
         self.x_out = Function(self.function_space)
-        if self.time_discretisation.idx is None:
+
+        if self.tracer_fs is not None:
```
tommbendall, line 113:
```suggestion
        if self.original_space is not None:
```

gusto/wrappers.py
```diff
@@ -104,7 +109,10 @@ def setup(self):
...
+            self.x_projected = Function(self.tracer_fs)
```
tommbendall, line 114:
```suggestion
            self.x_projected = Function(self.original_space)
```
?

gusto/wrappers.py
```diff
@@ -161,13 +169,19 @@ class RecoveryWrapper(Wrapper):
...
 
-        original_space = self.time_discretisation.fs
         domain = self.time_discretisation.domain
         equation = self.time_discretisation.equation
 
+        if self.tracer_fs is not None:
+            original_space = self.tracer_fs
+        else:
+            original_space = self.time_discretisation.fs
+
```
tommbendall, line 184: Again this snippet changes to
```
self.original_space = original_space
```

gusto/wrappers.py
```diff
@@ -184,10 +198,17 @@ def setup(self):
         # Internal variables to be used
         # -------------------------------------------------------------------- #
 
-        self.x_in_tmp = Function(self.time_discretisation.fs)
+        if self.tracer_fs is not None:
```
tommbendall, line 201:
```suggestion
        if self.original_space is not None:
```

gusto/wrappers.py
```diff
@@ -184,10 +198,17 @@ def setup(self):
...
+            self.x_in_tmp = Function(self.tracer_fs)
```
tommbendall, line 202:
```suggestion
            self.x_in_tmp = Function(self.original_space)
```

gusto/wrappers.py
```diff
@@ -184,10 +198,17 @@ def setup(self):
...
+        else:
+            self.x_in_tmp = Function(self.time_discretisation.fs)
+
         self.x_in = Function(self.function_space)
         self.x_out = Function(self.function_space)
-        if self.time_discretisation.idx is None:
+
+        if self.tracer_fs is not None:
```
tommbendall, line 209:
```suggestion
        if self.original_space is not None:
```

gusto/wrappers.py
```diff
@@ -184,10 +198,17 @@ def setup(self):
...
+            self.x_projected = Function(self.tracer_fs)
```
tommbendall, line 210:
```suggestion
            self.x_projected = Function(self.original_space)
```

tommbendall (changes requested):
Thanks again for this Tim. I think we're nearly there.

I understand why you needed to introduce the `tracer_fs` to the other wrappers. I think we could go a bit further with this by:
- renaming it as `original_space` (since it might not always be used with tracers!)
- having it set more explicitly through the `setup` method, which will help make the code clearer

gusto/time_discretisation.py
ta440: I think we would want this to be None, as we want self.original_space to be None when not using the mixed wrapper.

ta440:
Thanks for the comments, Tom!
By having the `original_space` set in the `setup`, I assume you mean for the individual wrappers (EmbeddedDG, Recovery), as if it was in the original wrapper object, we still have to pass `original space` as an argument into the individual wrapper instances. If so, I don't think we need original_space to actually be a property of wrappers, as we pass it in if using `MixedFSWrapper`, and it will default to `None` if not.
I think this requires the argument to the setup() function being something other than `original_space`, to avoid confusion when we define it later on. I've put in `previous_space` as a placeholder, but can change it to a better name? 
Please let me know if this wasn't the implementation you were thinking of! Perhaps there is a simpler way to set `self.original_space` before performing the `subwrapper.setup()`?

tommbendall (approved):
Thanks Tim for doing all of this! I think this is ready now.

PR #473, opened 2024-01-23
Fix SWSaturationAdjustment

nhartney:
This removes the redundant `L_v` parameter from the `SWSaturationAdjustment` physics class. The condensation proportionality parameter in Zerroukat and Allen (2015) is called `L_v` but is in fact the same thing as `beta2`.


tommbendall (approved):
This looks completely fine to me!

PR #477, opened 2024-02-19
Thermal linear solver fixes

nhartney:
This PR makes three changes to the `ThermalSWSolver` class in `linear_solvers.py`.

1. Checks that buoyancy is one of the fields in the equations set when using the `ThermalSWSolver`. The solver should only be used in equations sets with a thermal field and this will raise a `NotImplementedError` if the solver is being used and there is no buoyancy field.
2. Checks that a reference profile for buoyancy has been set. The solver will still work without a reference buoyancy profile but will work on the basis that the reference buoyancy is zero. This will log a warning to the user that no buoyancy reference profile has been set and to set one if the profile should be a non-zero one.
3. Considers `bbar` to be in the DG space and discretises the terms involving derivatives of `bbar` in the thermal linear solver using integration by parts.

### DIFF
--- a/gusto/linear_solvers.py
+++ b/gusto/linear_solvers.py
@@ -586,6 +586,10 @@ class ThermalSWSolver(TimesteppingSolver):
         VD = equation.domain.spaces("DG")
         Vb = equation.domain.spaces("DG")
 
+        # Check that the third field is buoyancy
+        if not equation.field_names[2] == 'b':
+            raise NotImplementedError("Field 'b' must exist to use the thermal linear solver in the SIQN scheme")
+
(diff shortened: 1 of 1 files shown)


jshipton (changes requested):
Looks good! Please just update PR description now that we are not interpolating b to the CG space (and carefully check where this line has gone!)

nhartney:
> Looks good! Please just update PR description now that we are not interpolating b to the CG space (and carefully check where this line has gone!)

I think it looks like the line isn't there because it is comparing to what is on main rather than comparing to previous commits on this branch. The lines do appear in the commit history of the branch.

jshipton (approved):

PR #478, opened 2024-02-20
Fix moist thermal example

nhartney:
This adds specified transport methods to the moist thermal example. Their omission was a mistake left over from when the spatial discretisation was prescribed with the equation. This specifies that the method is `DGupwind` for all transported fields.
This PR also adjusts the threshold for rain, `qprecip`, and the proportion converted when rain becomes cloud, `gamma_r`, so that these parameters are in line with the amount of cloud produced and rain will be generated.

### DIFF
--- a/examples/shallow_water/moist_thermal_williamson5.py
+++ b/examples/shallow_water/moist_thermal_williamson5.py
@@ -34,8 +34,8 @@ mu1 = 0.05
 mu2 = 0.98
 q0 = 135  # chosen to give an initial max vapour of approx 0.02
 beta2 = 10
-qprecip = 10e-4
-gamma_r = 10e-3
+qprecip = 1e-4
+gamma_r = 1e-3
 # topography parameters
 R0 = pi/9.
 R0sq = R0**2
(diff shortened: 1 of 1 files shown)


examples/shallow_water/moist_thermal_williamson5.py
```diff
@@ -34,8 +34,8 @@
 mu2 = 0.98
 q0 = 135  # chosen to give an initial max vapour of approx 0.02
 beta2 = 10
-qprecip = 10e-4
-gamma_r = 10e-3
+qprecip = 1e-4
```
jshipton, line 37: can you comment on why these have changed?

jshipton (changes requested):
please just comment on the change of thresholds...

jshipton (approved):

examples/shallow_water/moist_thermal_williamson5.py
nhartney: The original thresholds came from Zerroukat and Allen (2015) but in our model there is less cloud produced than in that paper. Using the threshold for rain from Zerroukat and Allen (2015) would mean that our model would never have sufficient cloud to generate rain, so we have reduced the threshold so that we do get cloud exceeding it and being converted to rain.

PR #479, opened 2024-02-27
Limiter to clip fields to zero

tommbendall:
This adds a simple limiter to clip fields to remove any negative values.

There is an option to perform the clipping on the field as is (which prevents DoF values from being negative), but also to transform to some equispaced element and clip there (which when used for DG1 may prevent the polynomial expansion from going negative).


jshipton (approved):
Fantastic - thanks @tommbendall !!

PR #480, opened 2024-02-29
Change default iterations of SIQN to be 2x2

tommbendall:
I'm not sure if we can run with 2x2 but it would be much better if we could!

So this changes the default to be 2x2 and we'll see which tests fail...


jshipton (approved):
Great - thanks @tommbendall !

PR #483, opened 2024-03-07
Fix to linearisation of contiunuity split, and solver parameters now passed through for IMEX Multistage

atb1995:
IMEX Multistage was using PetSc default solver settings (very inefficient) so now passed through default solver parameters. Longer term it might be wise to have two sets of solver parameters "linear solver parameters" for linear solves (mass matrix inversions etc.) and one for nonlinear solvers (for implicit time stepping).
A fix to a bug in the linearisation in common form continuity split has also been added


jshipton (approved):
great - thanks Alex!!

PR #486, opened 2024-03-07
Siqn improvements

jshipton:
This makes the SIQN timeloop closer to the LFRic timeloop in a couple of very minor ways:

1. The explicit forcing is now applied to `xn` rather than `x_after_slow_phys`. Since we haven't yet run anything slow physics this should not change anything.
2. Add the option to offcentre the transporting velocity but default to using a centred transporting velocity. This will only affect any runs that use an offcentring value of `alpha` that is not equal to a half.


tommbendall (approved):
Thanks Jemma!

PR #487, opened 2024-03-08
Fix sponge layer

tommbendall:
I think the sponge layer wasn't working in the mountain test case because a rogue dt factor. Removing this it looks like it's doing the right thing


jshipton (approved):
Good spot!

PR #489, opened 2024-03-14
A couple of small fixes

jshipton:
* add the missing word "term" to a warning comment in the logger
* fix to enable user to specify not to log the Courant number


tommbendall (approved):
The changes are clear so this can go on, thanks Jemma!

PR #490, opened 2024-03-20
implements compressible Boussinesq option by generalising the incompr…

jshipton:
Implements compressible Boussinesq option by generalising the incompressible Boussinesq equation class.

* test for boussinesq balance added
* KGOs added for incompressible and compressible Boussinesq equations (with 2x2 SIQN)


gusto/linear_solvers.py
```diff
@@ -442,6 +442,11 @@ class IncompressibleSolver(TimesteppingSolver):
     @timed_function("Gusto:SolverSetup")
     def _setup_solver(self):
         equation = self.equations      # just cutting down line length a bit
+
+        # flag to indicate whether we're solving the compressible
```
tommbendall, line 446: Very minor point, but since you only use `compressible` in one place below, it's probably not worth creating a new variable?

tommbendall (changes requested):
Looks really good, I think it's the right thing to do to implement this in the existing `IncompressibleBoussinesqEquations` (and to change the name to `BoussinesqEquations`)

Thoughts:
- I have a trivial suggestion about a variable that could be removed
- as you flagged in the PR description, you need to test the compressible equations and have a checkpointed file as a KGO
- should we also test the hydrostatic balance routine for both compressible and incompressible forms of the Boussinesq equations?

gusto/linear_solvers.py
jshipton: I love minor points - done!

tommbendall (approved):
Looks good to me, thanks Jemma

PR #491, opened 2024-03-22
Linear boussinesq

jshipton:


gusto/equations.py
```diff
@@ -1347,24 +1370,38 @@ def __init__(self, domain, parameters, Omega=None,
...
+        if compressible:
+            cs = parameters.cs
+            linear_div_form = name_label(
+                subject(prognostic(cs**2 * phi * div(u_trial) * dx, 'p'),
+                        self.X), "divergence")
+            divergence_form = name_label(linearisation(
+                subject(prognostic(cs**2 * phi * div(u) * dx, 'p'), self.X),
+                linear_div_form), "divergence")
+        else:
+            # This enforces that div(u) = 0
```
jshipton, line 1395: the results look odd when running incompressible Boussinesq with RK4 but are fine with SIQN - this comment is about the forcing term which makes me suspicious...

gusto/equations.py
```diff
@@ -1382,3 +1419,90 @@ def __init__(self, domain, parameters, Omega=None,
...
+        Raises:
+            NotImplementedError: active tracers are not implemented.
+        """
+
+        if linearisation_map == 'default':
+            # Default linearisation is time derivatives, pressure gradient,
+            # Coriolis and transport term from depth equation
+            linearisation_map = lambda t: \
+                (any(t.has_label(time_derivative, pressure_gradient, coriolis))
+                 or t.get(name_label) in ["gravity", "divergence", "incompressibility"]
```
jshipton, line 1494: is it odd that some of our terms are named using a label (e.g. `pressure_gradient`) whereas others have a `name_label`? should we rationalise this?

gusto/equations.py
atb1995: I'd assume it should work fine for small enough `dt` - the issue I was having was that the range of stable `dt` for the IMEX RK schemes was being limited by effectively treating a `div(u)` term explicitly. You could try using IMEX RK, with the div term labeled first explicitly, then implicitly and see if you see the same issues when it is explicit (but not when it is implicit). Or just use smaller `dt`?

jshipton: no, it's not that - sorry, I should have been clearer. The results don't show instability - they're just wrong. And the reason is that this isn't really an evolution equation for the pressure. What we're trying to do is solve for the pressure that gives us a non-divergent velocity. When we use the SIQN time stepper, the incompressible term is implicit and, as the comment here says, we overwrite the pressure rather than updating it.

gusto/equations.py
jshipton: I fixed this (or at least improved it) in #493

jshipton:
I think that the issue with the incompressible term should become an issue, since that's not what this PR was for. Would you like a test for the linear compressible case @tommbendall ?

tommbendall:
There are conflicts with this branch now so I'm afraid these will need resolving before this can go on.

I think PR is fine without a test given the example that you've added, and I agree that we should just have an issue to explain why this won't work for now with explicit timesteppers.

tommbendall (approved):
Everything's passing so this is ready to merge!

PR #492, opened 2024-03-26
Improve tracer density

ta440:
Improving the tracer density diagnostic. This constructs a higher-order space for the tracer density based on the function spaces for the mixing ratio and density.
The active tracers of rain, cloud_water, water_vapour, have an added density_name argument to make them compatible with using conservative transport.


gusto/active_tracers.py
```diff
@@ -134,15 +141,21 @@ def __init__(self, name='cloud_water', space='theta',
...
+                         transport_eqn=transport_eqn,
+                         density_name=density_name,
+                         phase=Phases.liquid, chemical='H2O')
 
 
 class Rain(ActiveTracer):
     """An object encoding the details of rain as a tracer."""
     def __init__(self, name='rain', space='theta',
                  variable_type=TracerVariableType.mixing_ratio,
+                 density_name=None,
```
jshipton, line 158: should the `density_name` argument come after the `transport_eqn` argument for consistency?

gusto/diagnostics.py
```diff
@@ -1717,7 +1723,36 @@ def setup(self, domain, state_fields):
             domain (:class:`Domain`): the model's domain object.
             state_fields (:class:`StateFields`): the model's field container.
         """
+
         m_X = state_fields(self.mixing_ratio_name)
         rho_d = state_fields(self.density_name)
+
+        m_X_space = m_X.function_space()
```
tommbendall, line 1730: I think we want all of this code to be protected behind an `if self.space_defined is None` -- currently this diagnostic allows the user to pass in a `space` (and we want the user to be able to do that) but it will be overridden by this code. I think we only want to override if `space is None`

tommbendall (changes requested):

gusto/active_tracers.py
ta440: Oops, good spot, thanks!

gusto/diagnostics.py
ta440: Cool, I've changed it to enable this, although I've used `if self.space is None` instead of `self.space_defined`

gusto/diagnostics.py
```diff
@@ -1720,4 +1726,36 @@ def setup(self, domain, state_fields):
         m_X = state_fields(self.mixing_ratio_name)
         rho_d = state_fields(self.density_name)
         self.expr = m_X*rho_d
-        super().setup(domain, state_fields)
+
+        if self.space is None:
```
tommbendall, line 1730: Final thing: could you add some comments to explain why we want to have a specific space for the tracer density? Thanks!

tommbendall (changes requested):

gusto/diagnostics.py
ta440: Sure, I've added a few more comments.

tommbendall (approved):
Thanks Tim!

PR #493, opened 2024-03-26
Name terms

jshipton:
This fixes an inconsistency in the way we name the terms in our equations. I've got rid of `name_label` and created labels for the terms that we have.

This also fixes a bug in the `DynamicsLabel` and `PhysicsLabel` classes where nothing was returned if the super class was called.


tommbendall (approved):
Looks good, thanks for fixing the bug in the `DynamicsLabel` and `PhysicsLabel` too

PR #494, opened 2024-03-26
1D shallow water equations

jshipton:
This PR adds an equation class for the 1D shallow water equations and the required 1D transport and diffusion terms.

To Do:
* make example into a KGO test instead of an example
* define linearisation
* this won't work with SIQN without another bespoke linear solver - I'd rather not write that!

@tommbendall any comments / requests?


jshipton:
aha, we used to make the velocity a vector with just one component... @tommbendall would you like to keep it like this (to test the usual code for the transport forms) or shall I change this test to use the new 1D forms?

gusto/equations.py
```diff
@@ -879,6 +881,182 @@ def __init__(self, domain, parameters, fexpr=None, bexpr=None,
...
 
 
+class ShallowWaterEquations_1d(PrognosticEquationSet):
+
+    u"""
+    Class for the (rotating) 1D shallow-water equations, which describe
+    the velocity 'u', 'v' and the depth field 'D', solving some variant of:  \n
+    ∂u/∂t + u∂u/∂x - fv + g*∂D/∂x = 0,                                       \n
+    ∂v/∂t + fu = 0,                                                          \n
+    ∂D/∂t + ∂(uD)/∂x = 0,                                                     \n
```
tommbendall, line 891: Super-fussy, on github the \n doesn't look like it's aligned

gusto/equations.py
```diff
@@ -879,6 +881,182 @@ def __init__(self, domain, parameters, fexpr=None, bexpr=None,
...
+
+        super().__init__(domain, parameters,
+                         fexpr=fexpr, space_names=space_names,
+                         linearisation_map=linearisation_map,
+                         no_normal_flow_bc_ids=no_normal_flow_bc_ids,
+                         active_tracers=active_tracers)
+
+        # Use the underlying routine to do a first linearisation of the equations
+        self.linearise_equation_set()
+        for t in self.residual:
```
tommbendall, line 1056: This print statement will need removing!

tommbendall (changes requested):
It's great that we'll finally be able to do this! As well as the picky individual comments, some broad thoughts:
- as we discussed offline it would probably be better for the normal form to return the 1D form if (a) the mesh is 1D and (b) the velocity is a scalar. Hopefully this will ensure that the tests pass
- could you rename the example to `shallow_water_1d.py` or something similar, for some reason I don't like it starting with a number!

gusto/equations.py
jshipton: urgh, sorry!

tommbendall (approved):
I'm happy with this now! I'll leave it un-merged so that you can do it if you think that it's ready

PR #495, opened 2024-03-27
Compressible vorticity

Witt-D:
Added compressible vorticity diagnostics so that a vector vorticity can be calculated for the compressible euler equations for both the absolute and relative versions of vorticity


gusto/equations.py
```diff
@@ -1072,10 +1072,12 @@ def __init__(self, domain, parameters, Omega=None, sponge=None,
         # -------------------------------------------------------------------- #
         # Extra Terms (Coriolis, Sponge, Diffusion and others)
         # -------------------------------------------------------------------- #
-        if Omega is not None:
-            # TODO: add linearisation and label for this
-            residual += subject(prognostic(
-                inner(w, cross(2*Omega, u))*dx, "u"), self.X)
+
```
jshipton, line 1075: we better not lose the comment reminding us to add the linearisation here

gusto/diagnostics.py
```diff
@@ -1687,6 +1687,110 @@ def setup(self, domain, state_fields):
...
+
+
+class CompressibleAbsoluteVorticity(CompressibleVorticity):
+    u""" Diagnostic field for compressible euler absolute vorticity  """
+    name = 'CompressibleAbsoluteVorticity'
+
+    def __init__(self, parameters, space=None, method='solve'):
+        u"""
+        Args:
+            space (:class:`FunctionSpace`, optional): the function space to
```
jshipton, line 1772: please update docstring to include `parameters` argument

gusto/diagnostics.py
```diff
@@ -1687,6 +1687,110 @@ def setup(self, domain, state_fields):
...
+
+        super().setup(domain, state_fields, space=space)
+
+        if self.method == 'solve':
+            vort = TrialFunction(space)
+            w = TestFunction(space)
+            n = FacetNormal(domain.mesh)
+            a = inner(vort, w) * dx
+            L = inner(u, curl(w)) * dx - jump(cross(w, u), n) * dS_h
+            if vorticity_type != 'relative':
```
jshipton, line 1729: why not `if vorticity_type == 'absolute'` as above?

gusto/diagnostics.py
```diff
@@ -1687,6 +1687,110 @@ def setup(self, domain, state_fields):
         super().setup(domain, state_fields, vorticity_type="relative")
 
 
+class CompressibleVorticity(DiagnosticField):
+    u""" Base diagnostic Field for three dimensional Vorticity """
```
jshipton, line 1691: does this (should this?) also work in a vertical slice?

jshipton (changes requested):
Thanks @Witt-D, it'll be great to have this in! I've just left a couple of minor comments to address...

gusto/diagnostics.py
Witt-D: This is a good question, which im not sure on so will have a test around and decide if it does / should it, i think it should work but may require a bit of thinking

tommbendall:
I'm afraid that this now conflicts with changes that have just gone onto main -- could you fix these and then we can look at getting this on?

gusto/diagnostics.py
jshipton: I think it's just `CompressibleParameters` - this should also work for the Boussinesq equations so you can remove any mention of Euler... in fact, it also works for incompressible...

gusto/diagnostics.py
```diff
@@ -1689,6 +1692,121 @@ def setup(self, domain, state_fields):
...
+        else:
+            space = domain.spaces('H1')
+
+        u = state_fields('u')
+        if self.method != 'solve':
+            if vorticity_type == 'relative':
+                self.expression = curl(u)
+            elif vorticity_type == 'absolute':
+                Omega = as_vector((0, 0, self.parameters.Omega))
+                self.expression = curl(u) + 2*Omega
```
tommbendall, line 1727: Should this be `self.expr`?

gusto/diagnostics.py
```diff
@@ -1689,6 +1692,121 @@ def setup(self, domain, state_fields):
...
+                raise ValueError(f'vorticity type must be one of {vorticity_types}, not {vorticity_type}')
+        if domain.mesh.topological_dimension() == 3:
+            space = domain.spaces('HCurl')
+        else:
+            space = domain.spaces('H1')
+
+        u = state_fields('u')
+        if self.method != 'solve':
+            if vorticity_type == 'relative':
+                self.expression = curl(u)
```
tommbendall, line 1724: Should this be `self.expr`?

gusto/diagnostics.py
```diff
@@ -1689,6 +1692,121 @@ def setup(self, domain, state_fields):
...
+        Sets up the :class:`Function` for the diagnostic field.
+
+        Args:
+            domain (:class:`Domain`): the model's domain object.
+            state_fields (:class:`StateFields`): the model's field container.
+            vorticity_type (str, optional): denotes which type of vorticity to
+                be computed ('relative', 'absolute'). Defaults to
+                None.
+        """
+        # TODO Do we eventually want a potential voriticy?
```
tommbendall, line 1709: Yes I think we will -- you can remove this comment for now though

gusto/diagnostics.py
tommbendall: Since you've put in effort to make this work in a vertical slice, you can change this comment!

gusto/diagnostics.py
```diff
@@ -1689,6 +1692,121 @@ def setup(self, domain, state_fields):
...
+                if vorticity_type == 'absolute':
+                    Omega = as_vector((0, 0, self.parameters.Omega))
+                    L += inner(2*Omega, w) * dx
+            else:
+                vort = TrialFunction(space)
+                gamma = TestFunction(space)
+                n = FacetNormal(domain.mesh)
+                a = inner(vort, gamma) * dx
+                L = ( inner(domain.perp(grad(gamma)), u))*dx - jump(inner(domain.perp(n), u)*gamma)*ds_h
+                # TODO implement absolute version, unsure atm how to get corioilis in vertical slice smartly
```
tommbendall, line 1747: This isn't obvious to me, and it isn't obvious that we'll ever want the absolute vorticity in a vertical slice. I suggest that you put an if check here so that we fail with a `NotImplementedError` in this situation

gusto/equations.py
```diff
@@ -1256,10 +1256,12 @@ def __init__(self, domain, parameters, Omega=None, sponge_options=None,
         # -------------------------------------------------------------------- #
         # Extra Terms (Coriolis, Sponge, Diffusion and others)
         # -------------------------------------------------------------------- #
-        if Omega is not None:
-            # TODO: add linearisation
-            residual += coriolis(subject(prognostic(
+        if parameters.Omega is not None:
+            # TODO add linerisation and label for this
+            Omega = as_vector((0, 0, parameters.Omega))
```
tommbendall, line 1261: Maybe we should call this variable `Omega_vec` to avoid confusion

gusto/diagnostics.py
```diff
@@ -1689,6 +1692,121 @@ def setup(self, domain, state_fields):
...
+
+        Args:
+            domain (:class:`Domain`): the model's domain object.
+            state_fields (:class:`StateFields`): the model's field container.
+        """
+        super().setup(domain, state_fields, vorticity_type='relative')
+
+
+class AbsoluteVorticity(Vorticity):
+    u""" Diagnostic field for compressible euler absolute vorticity  """
```
tommbendall, line 1780: As Jemma says, you could remove reference to `compressible euler` here?

PR #503, opened 2024-04-11
Fix use of repeated limiter in MixedFSLimiter

tommbendall:
This fixes a small bug in how the `MixedFSLimiter` works when used with a single instance of a sub-limiter for different fields.

Previously we were assigning the index of the field component to the sub-limiter, so when we were using a single limiter object for multiple fields, only the last field was having the limiter applied to it. Instead we can store the field indices in the `MixedFSLimiter` itself.

Thanks @nhartney for putting together the test that revealed this!

### DIFF
--- a/integration-tests/transport/test_zero_limiter.py
+++ b/integration-tests/transport/test_zero_limiter.py
@@ -167,7 +167,7 @@ def test_without_limiter(tmpdir):
 
     cloud_without_limiter = stepper_without_limiter.fields('cloud_water')
 
-    assert cloud_without_limiter.dat.data.min() < 0, "The minimum of cloud is negative"
+    assert cloud_without_limiter.dat.data.min() < 0, "The minimum of unlimited cloud should be negative"
 
 
 def test_with_limiter(tmpdir):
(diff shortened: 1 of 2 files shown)


integration-tests/transport/test_zero_limiter.py
```diff
@@ -1,174 +1,199 @@
...
+    # Setup and run verison without limiter
+    dirname = str(tmpdir)
+
+    stepper_without_limiter, tmax = setup_zero_limiter(dirname)
+
+    stepper_without_limiter.run(t=0, tmax=tmax)
+
+    cloud_without_limiter = stepper_without_limiter.fields('cloud_water')
+
+    assert cloud_without_limiter.dat.data.min() < 0, "The minimum of cloud is negative"
```
nhartney, line 170: Should the assert be something like "The minimum of cloud is already non-negative"?

tommbendall: Thanks for pointing that out! Should be fixed now

nhartney:
Brilliant! Thanks so much for sorting this out.

nhartney (approved):

PR #504, opened 2024-04-12
Conservative transport with physics

ta440:
This resolves an incompatibility when using conservative transport and a physics scheme. The conservative transport needs a mass weighted (multiplied by density) form of the mass and transport terms, whilst a physics scheme does not use these. To resolve this, we introduce a mass_weighted label. When there is a term with both transport and mass_weighted labels, then we firstly check if there is a physics scheme, and raise an error if so (this means we are solving all the terms at once, when we need to split the transport and physics e.g. use SplitPrescribedTransport instead of PrescribedTransport). Else, we replace the terms with their mass-weighted counterparts to enable conservative transport.
A test of this functionality is added in the integration-tests/model/test_coupled_transport_with_physics.py file.


tommbendall (changes requested):

integration-tests/model/test_conservative_transport_with_physics.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # ------------------------------------------------------------------------ #
+    # Initial conditions
+    # ------------------------------------------------------------------------ #
+
+    rho0 = stepper.fields("rho_d")
+    ash0 = stepper.fields("ash")
+
+    # Set an initial constant density field
+    rho0.interpolate(Constant(0.0))
```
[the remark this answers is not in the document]
ta440: It isn't blowing up, but just setting any transport terms to zero. I've changed the test to have a source of area one in the bottom LHS. When we set the density to zero, we have that the box remains in place (left) and with a constant density of 2000, the box starts advecting with velocity (0.5,0.5).
source_sink_conservative_source_time_varyFalsezero_rho_ash_tidx_-1
source_sink_conservative_source_time_varyFalse_ash_tidx_-1

gusto/time_discretisation.py
```diff
@@ -168,6 +168,24 @@ def setup(self, equation, apply_bcs=True, *active_labels):
                     self.evaluate_source.append(t.labels[physics_name])
                     self.physics_names.append(t.labels[physics_name])
 
+        # Check if there is any conservative transport, which will
```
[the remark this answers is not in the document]
ta440: I see what you mean. However, this means we assume a transport method is always specified, which isn't always the case. If no transport methods are given then we won't correctly replace the terms in the residual.

gusto/equations/prognostic_equations.py
```diff
@@ -142,9 +142,29 @@ def generate_mass_terms(self):
...
 
+        if self.active_tracers is None:
+            tracer_names = []
+        else:
+            tracer_names = [tracer.name for tracer in self.active_tracers]
+
         for i, (test, field_name) in enumerate(zip(self.tests, self.field_names)):
             prog = split(self.X)[i]
             mass = subject(prognostic(inner(prog, test)*dx, field_name), self.X)
+            for j, ind_tracer_name in enumerate(tracer_names):
```
tommbendall, line 153: Very picky, but would this variable make more sense to be called just `tracer_name`?

tommbendall, line 153: Could we add a comment above this for loop to explain what the purpose of the loop is?

gusto/time_discretisation/time_discretisation.py
```diff
@@ -161,6 +161,28 @@ def setup(self, equation, apply_bcs=True, *active_labels):
...
+
+                # Check if the mass term for this prognostic is mass-weighted
+                if len(self.residual.label_map(lambda t: t.get(prognostic) == field and t.has_label(time_derivative) and t.has_label(mass_weighted), map_if_false=drop)) == 1:
+                    field_terms = self.residual.label_map(lambda t: t.get(prognostic) == field and not t.has_label(time_derivative), map_if_false=drop)
+
+                    # Check that the equation for this prognostic does not involve
+                    # both mass-weighted and non-mass-weighted terms; if so, a split
+                    # timestepper should be used instead.
+                    if len(field_terms.label_map(lambda t: t.has_label(mass_weighted), map_if_false=drop)) > 0:
+                        if len(field_terms.label_map(lambda t: not t.has_label(mass_weighted), map_if_false=drop)) > 0:
```
tommbendall, line 176: Could these two if statements be joined by an `and` rather than being nested?

tommbendall (changes requested):
Very nearly there, thanks Tim! Just a couple of minor suggestions

gusto/equations/prognostic_equations.py
ta440: Sure!

tommbendall (approved):
Thanks Tim!

PR #505, opened 2024-04-26
Ensemble parallelism

jshipton:
When running time parallel problems we still need to interact with the io. This PR is a start. The `Ensemble` is created first so that the spatial communicator can be passed to the mesh. The ensemble instance is then passed to the timestepper so that it can be passed as required to the io and (in future, or on other branches!) to the time discretisation classes.

Here I have sorted out the VTK output. Diagnostics also seem to work, but I have not verified that they are correct. I have not thought about netCDF output or the case where we pick up. I have implemented a test that doesn't check anything other than for hanging. I can see that it does hang if `spatial_parallelism=True` but this seems to be a pytest issue as it is creating multiple output directories, one for each process.

I know @JDBetteridge and @tommbendall have been looking at parallel tests and having issues - would be good to hear what you think!

### DIFF
--- a/gusto/configuration.py
+++ b/gusto/configuration.py
@@ -61,7 +61,7 @@ class Configuration(object):
 
         When attributes are provided as floats or integers, these are converted
         to Firedrake :class:`Constant` objects, other than a handful of special
-        integers (dumpfreq, pddumpfreq, chkptfreq and log_level).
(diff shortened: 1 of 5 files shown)


JDBetteridge:
A quick note that `Ensembe` is currently broken in Firedrake. This would be a good little project for the upcoming hackathon!

PR #506, opened 2024-04-29
Remove guess of first value in explicit multistage scheme

tommbendall:
At the start of each explicit multistage step, we initialise the expected result with our current field as a "guess".

This was added in #472 which came onto main in #469. It was an attempt to ensure that the explicit multistage schemes with the increment form developed in #413 could be used with conservative tracer transport. These equations do not have a unique solution if the density is zero, but provided a guess and using an iterative method helps to avoid issues).

However it appears that it may be causing problems in various tests! So this PR removes that line.


jshipton:
Should we revert to the previous KGO for the linear shallow water test?

jshipton (approved):

PR #511, opened 2024-07-03
Add `gusto.recovery` to packages

francesco-ballarin:
Noticed this when running through `pip install` rather than `python3 setup.py`, it's possible that `pip` is more picky than legacy setuptools when installing subpackages.

cc @jshipton


jshipton (approved):
Fantastic - thanks @francesco-ballarin !!

PR #513, opened 2024-07-20
Fix limit_midpoint kernel

tommbendall:
This fixes the `limit_midpoint` kernel by changing the DoF ordering -- to pick up on the corresponding change from Firedrake. This fixes Gusto's currently failing tests.


jshipton (approved):
thanks @tommbendall !

PR #518, opened 2024-07-22
Courant numbers for a scalar velocity field

ta440:
Checks if the velocity field is a scalar, and if so, computes the Courant number accordingly (instead of using inner products for a vector velocity field). This Courant number is used in the 1D RSWEs example.


tommbendall (approved):
This looks good, thanks Tim!

PR #523, opened 2024-07-23
Take maximum of h across all ranks

JDBetteridge:


jshipton (approved):
thanks @JDBetteridge !

PR #525, opened 2024-07-23
Add dew point and wet-bulb temperature diagnostics

tommbendall:
This adds diagnostics for dew point and wet-bulb temperatures. There is probably no application in which we will use them but they were fun to make. Physically these temperatures are defined by implicit formulae, so can't be solved analytically. I have implemented them through fixed-point methods. I could've used one of the empirical formulae but this would've been less fun.

Actually the more important contributions from this PR are:
- the iterative base diagnostic, which may be useful in the future
- the introduction of first unit-tests for diagnostics (this is based off some earlier work by @Witt-D)

I have also attached a document describing how I arrived at the implementation of these diagnostics ...

[Wet_Temperatures.pdf](https://github.com/user-attachments/files/16353109/Wet_Temperatures.pdf)


jshipton (approved):
I'm itching to find a use for this now!!

PR #530, opened 2024-07-26
set quadrature degree for forms in equation residual

jshipton:
This PR ensures that we no longer unthinkingly exceed sensible quadrature degree limits. This is achieved by adding a `label_map` to the equation residual in the `setup_equation` method of the `Timestepper` class which replaces all of the terms with new terms containing the same form and labels, the only difference being that the integrals in the form have had their metadata updated to contain a quadrature degree of 5.

To Do:
* use logging to log an info level warning when the quadrature degree is set
* add `max_quad_deg` to other equation sets
* use `max_quad_deg` in initialisation (and diagnostics?)


tommbendall:
Closing as this is now duplicated by #595 (which may itself be shortly superseded!)

PR #533, opened 2024-08-07
Reorganise prescribed transport

tommbendall:
This reorganises the prescribed transport to allow more complicated methods for defining the transporting velocity. 

The transporting velocity is now set up after the stepper, and can be done through three methods:
- either not at all, in which case the velocity is assumed to be constant in time
- through a time-varying expression as a python function (as we already done)
- through a user-defined apply method -- this is the main new capability

The main motivation is that sometimes we want more complicated methods for defining the transporting velocity, e.g.:
1. specifying the time-varying velocity through a stream function, by interpolating into the H1 space then projecting `gradperp(psi)`
2. the velocity is discontinuous in time. These cases cannot be handled with a single expression and so need something more complicated

It also has the nice benefit that time-varying winds are now set up after the timestepper, allowing the wind to be defined with the initial conditions.

### DIFF
--- a/gusto/timestepping/split_timestepper.py
+++ b/gusto/timestepping/split_timestepper.py
@@ -87,15 +87,18 @@ class SplitPrescribedTransport(Timestepper):
     velocity as a prognostic variable.
     """
 
-    def __init__(self, equation, scheme, io, spatial_methods=None,
-                 physics_schemes=None,
-                 prescribed_transporting_velocity=None):
+    def __init__(self, equation, scheme, io, prescribed_transporting_velocity,
+                 spatial_methods=None, physics_schemes=None):
         """
         Args:
             equation (:class:`PrognosticEquation`): the prognostic equation
             scheme (:class:`TimeDiscretisation`): the scheme to use to timestep
                 the prognostic equation
             io (:class:`IO`): the model's object for controlling input/output.
+            prescribed_transporting_velocity: (bool): Whether a time-varying
+                transporting velocity will be defined. If True, this will
+                require the transporting velocity to be setup by calling either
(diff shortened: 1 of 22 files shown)


gusto/timestepping/split_timestepper.py
```diff
@@ -153,10 +154,68 @@ def setup_scheme(self):
...
+        """
+        Sets up the prescribed transporting velocity, through a python function
+        which has time as an argument, and returns a `ufl.Expr`. This allows the
+        velocity to be updated with time.
+
+        Args:
+            expr_func (func): a python function with a single argument that
+                represents the model time, and returns a `ufl.Expr`.
+        """
+
```
jshipton, line 167: Do we need to check that the velocity has not already been setup (e.g. by using the other method)?

gusto/timestepping/split_timestepper.py
```diff
@@ -153,10 +154,68 @@ def setup_scheme(self):
...
+        Sets up the prescribed transporting velocity, through a python function
+        which has time as an argument. This function will perform the evaluation
+        of the transporting velocity.
+
+        Args:
+            expr_func (func): a python function with a single argument that
+                represents the model time, and performs the evaluation of the
+                transporting velocity.
+        """
+
```
jshipton, line 185: Do we need to check that the velocity has not already been setup (e.g. by using the other method)?

jshipton (changes requested):
Just one concern that the user may use both methods to set up a velocity - suggest to check `self.is_velocity_setup` at the start of each setup method.

Otherwise this is very nice.

jshipton (approved):

Issue #278, opened 2022-07-22
What are the correct initial conditions for the 1 metre mountain test cases?


tommbendall:
There may or may not be a bug in the 1 metre mountain test cases, where the pressure is set up several times. The minimum value is used but maybe it should be the maximum?

jshipton:
Fixed by #534

PR #534, opened 2024-08-08
Fix mountain initial conditions

tommbendall:
This PR fixes the initial conditions for the mountain example. I'm confident that this is now correct for giving a max bottom Exner value that is close to 1.0 (previously it was 2.0 at the bottom and 1.0 at the top!!)

I've copied some code that @jshipton sent me ages ago so I can't take credit for the method!

This addresses #278


jshipton (approved):
Great - thanks @tommbendall! My only thought was that this could go in the initialisation file rather than in each example as it'll be useful for other examples with mountains - however, we currently only have one example with a mountain, so: approved!

PR #536, opened 2024-08-14
Solver Convergence Improvements

atb1995:
Importing solver improvements we have in GungHo:

- tau values for linear solver which accelerate convergence if set to 1. This has been shown to make the moist baroclinic in channel stable for 2x2 SIQN with tau_t and tau_r = 1. Description of why these work has been attached in 
[Tau_Values.pdf](https://github.com/user-attachments/files/16623363/Tau_Values.pdf)

- Accelerate solver convergence by zeroing implicit forcing's for transport terms

These changes have been added on switches, so do not change KGOs.

### DIFF
--- a/gusto/solvers/linear_solvers.py
+++ b/gusto/solvers/linear_solvers.py
@@ -33,7 +33,7 @@ __all__ = ["BoussinesqSolver", "LinearTimesteppingSolver", "CompressibleSolver",
 class TimesteppingSolver(object, metaclass=ABCMeta):
     """Base class for timestepping linear solvers for Gusto."""
 
-    def __init__(self, equations, alpha=0.5, tau_values=None,
+    def __init__(self, equations, alpha=0.5, tau_values={},
                  solver_parameters=None, overwrite_solver_parameters=False):
         """
         Args:
@@ -41,7 +41,7 @@ class TimesteppingSolver(object, metaclass=ABCMeta):
             alpha (float, optional): the semi-implicit off-centring factor.
                 Defaults to 0.5. A value of 1 is fully-implicit.
             tau_values (dict, optional): contains the semi-implicit relaxation
-                parameters. Defaults to None.
+                parameters. Defaults to an empty dictionary.
             solver_parameters (dict, optional): contains the options to be
                 passed to the underlying :class:`LinearVariationalSolver`.
                 Defaults to None.
@@ -54,13 +54,6 @@ class TimesteppingSolver(object, metaclass=ABCMeta):
         self.dt = equations.domain.dt
         self.alpha = alpha
         self.tau_values = tau_values
-
-        # Check that we have a tau value for each field
-        if self.tau_values is not None:
-            for field in equations.field_names:
-                if field not in self.tau_values:
-                    raise KeyError(f"You must select tau values for all fields to be solved, "
-                                   f"missing value is {field}")
 
         if solver_parameters is not None:
             if not overwrite_solver_parameters:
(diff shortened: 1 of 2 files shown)


gusto/solvers/linear_solvers.py
```diff
@@ -164,24 +181,28 @@ def __init__(self, equations, alpha=0.5,
...
-        super().__init__(equations, alpha, solver_parameters,
+        super().__init__(equations, alpha, tau_u, tau_r, tau_t, solver_parameters,
                          overwrite_solver_parameters)
 
     @timed_function("Gusto:SolverSetup")
     def _setup_solver(self):
 
         equations = self.equations
         dt = self.dt
-        beta_ = dt*self.alpha
```
jshipton, line 175: It looks to me like `alpha` can be passed in but is no longer used anywhere - is that correct? Maybe it makes sense to have `alpha` default to 0.5 and the various `taus default to `None` but then set them to `alpha` if they are not specified?

atb1995: Thanks Jemma, I have updated so that if not tau values are given, they default to alpha

gusto/solvers/linear_solvers.py
```diff
@@ -52,6 +58,23 @@ def __init__(self, equations, alpha=0.5, solver_parameters=None,
         self.dt = equations.domain.dt
         self.alpha = alpha
 
+        # Set relaxation parameters. If an alternative has not been given, set
+        # to semi-implicit off-centering factor
+        if tau_u is not None:
```
tommbendall, line 63: these could actually be fit on one line nicely, e.g.
`self.tau_u = tau_u if tau_u is not None else alpha`

gusto/timestepping/semi_implicit_quasi_newton.py
```diff
@@ -477,3 +485,24 @@ def apply(self, x_in, x_nl, x_out, label):
...
+        Args:
+            equation (:class:`PrognosticEquationSet`): the prognostic
+                equation set to be solved
+            x_in (:class:`FieldCreator`): the field to be incremented.
+            x_out (:class:`FieldCreator`): the output field to be updated.
+            transported_field_names (str): list of fields names for transported fields
+        """
+        for field_name in transported_field_names:
+            if field_name != 'u':
+                logger.info(f'SIQN: Zeroing Implicit forcing for {field_name}')
```
tommbendall, line 506: I think the `SIQN` in the logged message now needs to be `Semi-Implicit Quasi Newton`

tommbendall (changes requested):

gusto/solvers/linear_solvers.py
```diff
@@ -33,15 +33,15 @@
...
+    def __init__(self, equations, alpha=0.5, tau_values=None,
                  solver_parameters=None, overwrite_solver_parameters=False):
         """
         Args:
             equations (:class:`PrognosticEquation`): the model's equation.
             alpha (float, optional): the semi-implicit off-centring factor.
                 Defaults to 0.5. A value of 1 is fully-implicit.
             tau_values (dict, optional): contains the semi-implicit relaxation
-                parameters. Defaults to an empty dictionary.
+                parameters. Defaults to None.
```
jshipton, line 44:
```suggestion
                parameters. Defaults to None, in which case the value of alpha is used.
```

gusto/solvers/linear_solvers.py
```diff
@@ -147,7 +147,7 @@ def __init__(self, equations, alpha=0.5, tau_values={},
...
```
jshipton, line 150:
```suggestion
                parameters. Defaults to None, in which case the value of alpha is used.
```

gusto/solvers/linear_solvers.py
```diff
@@ -33,13 +33,19 @@
 class TimesteppingSolver(object, metaclass=ABCMeta):
     """Base class for timestepping linear solvers for Gusto."""
 
-    def __init__(self, equations, alpha=0.5, solver_parameters=None,
-                 overwrite_solver_parameters=False):
+    def __init__(self, equations, alpha=0.5, tau_u=None, tau_r=None, tau_t=None,
```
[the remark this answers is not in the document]
atb1995: I've switched to a dictionary, with defaulting to alpha if the dictionary has no entry for a given field

jshipton (approved):
Thanks Alex!

tommbendall (approved):

PR #541, opened 2024-08-18
Revamp examples

tommbendall:
This revamps our examples, making them follow the structure now used in the case studies repository.

Examples are organised as functions, which should allow them to be easily run from the command line or through pytest. They all follow a common pattern.

I also realised that our hydrostatic examples don't use the hydrostatic equations. I've tried to turn them on but this has exposed errors which I will probably fix in a different PR.


jshipton (approved):
Thanks @tommbendall !!

PR #542, opened 2024-08-19
Allow reference profiles to be updated

tommbendall:
This PR allows the reference profiles used in the Semi-Implicit Quasi Newton time stepper to be updated with time.

This is set by an argument to the time stepper. When the reference profiles are updated, the compressible linear solver also needs to recompute some operators. I have also attempted to make this work with checkpointing.


gusto/core/io.py
```diff
@@ -531,7 +531,11 @@ def setup_dump(self, state_fields, t, pick_up=False):
 
         # dump initial fields
         if not pick_up:
-            self.dump(state_fields, t, step=1)
+            step = 1
+            last_ref_update_time = None
+            initial_steps = None
+            time_data = (t, step, initial_steps, last_ref_update_time)
```
jshipton, line 537: I feel slightly concerned about having a tuple of things that I might need to remember the order of! Could we use a [named tuple](https://docs.python.org/3/library/collections.html#collections.namedtuple)?

Something like (completely untested!):
```
TimeData = namedtuple('TimeData', ['t', 'step', 'initial_steps', 'last_ref_update_time'])
time_data = TimeData(t, step, initial_steps, last_ref_update_time)
```
Then you can later access these like: `t = time_data.t`.

gusto/timestepping/semi_implicit_quasi_newton.py
```diff
@@ -302,8 +339,7 @@ def timestep(self):
 
             for inner in range(self.num_inner):
 
-                # TODO: this is where to update the reference state
```
jshipton, line 305: Was this comment just wrong?

jshipton (changes requested):
I just have a couple of small questions... (see comments)

jshipton (approved):

PR #544, opened 2024-08-19
use correct beta_u in shallow water linear solvers

jshipton:
Sorry @Witt-D I didn't notice this in your PR.


tommbendall (commented):
This puts things in line with the rest of the code ...  but the awkward thought I've had is that maybe we don't want the `beta` values to be constant at all?

In theory we could update `dt`, but if we did then we would want the `beta` values to be updated too. At the moment they won't be, so do we actually want:
```
beta = self.dt*Constant(tau)
```
instead of
```
beta_ = self.dt*tau
beta = Constant(beta_)
```

gusto/solvers/linear_solvers.py
```diff
@@ -177,25 +177,20 @@ def __init__(self, equations, alpha=0.5, tau_values=None,
...
         dt = self.dt
         # Set relaxation parameters. If an alternative has not been given, set
         # to semi-implicit off-centering factor
-        beta_u_ = dt*self.tau_values.get("u", self.alpha)
-        beta_t_ = dt*self.tau_values.get("theta", self.alpha)
-        beta_r_ = dt*self.tau_values.get("rho", self.alpha)
+        beta_u = dt*self.tau_values.get("u", self.alpha)
+        beta_t = dt*self.tau_values.get("theta", self.alpha)
+        beta_r = dt*self.tau_values.get("rho", self.alpha)
+        beta_u_cp = Constant(beta_u * cp)
```
tommbendall, line 187: Can we remove this variable and just have beta_u*cp in the code below?

jshipton: Should've known I couldn't get away with being lazy!

tommbendall (approved):
Thanks!

PR #545, opened 2024-08-20
Conservative transport integration test

ta440:
Adds an integration test for conservative transport, which tests consistency and conservation for three configurations of rho_d and m_X: Same space (both in DG) order 1, different spaces (rho_d in DG, m_X in theta) for orders 0 and 1.
Also, fixed some bugs in the wrapper.py script with the implementation of the conservative versions of the EmbeddedDG and Recovery wrappers.

### DIFF
--- a/integration-tests/transport/test_conservative_transport.py
+++ b/integration-tests/transport/test_conservative_transport.py
@@ -1,0 +1,186 @@
+"""
+Tests the conservative transport method, for a mixing ratio and density in the
+same and different spaces. This checks that there is a conservation of
+"""
+
+from gusto import *
+from firedrake import PeriodicIntervalMesh, ExtrudedMesh, exp, cos, sin, SpatialCoordinate, \
+    assemble, dx, FunctionSpace, pi, min_value, as_vector, BrokenElement, errornorm
+import pytest
+
+
+def setup_conservative_transport(dirname, space, property):
+
+    # Domain
+    Lx = 2000.
+    Hz = 2000.
+
+    # Time parameters
+    dt = 2.
+    tmax = 2000.
+
+    nlayers = 10.  # horizontal layers
+    columns = 10.  # number of columns
+
+    # Define the spaces and order
+    if space == 'same':
+        rho_d_space = 'DG'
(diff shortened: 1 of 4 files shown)


integration-tests/transport/test_conservative_transport.py
```diff
@@ -0,0 +1,184 @@
...
+    else:
+        suboptions = {}
+
+    opts = MixedFSOptions(suboptions=suboptions)
+
+    transport_scheme = SSPRK3(domain, options=opts, increment_form=False)
+    transport_methods = [DGUpwind(eqn, "m_X"), DGUpwind(eqn, "rho_d")]
+
+    # Timestepper
+    stepper = PrescribedTransport(eqn, transport_scheme, io, transport_methods, prescribed_transporting_velocity=u_t)
```
tommbendall, line 147: Note that this will need updating when we merge the other branch up to head-of-trunk!

integration-tests/transport/test_conservative_transport.py
```diff
@@ -0,0 +1,184 @@
...
+    m_X_init = Function(V_m_X)
+    rho_d_init = Function(V_rho)
+
+    m_X_init.assign(stepper.fields("m_X"))
+    rho_d_init.assign(stepper.fields("rho_d"))
+
+    return stepper, m_X_init, rho_d_init
+
+
+@pytest.mark.parametrize("space", ["same", "diff_order_0", "diff_order_1"])
```
tommbendall, line 164: Could we call the `space` variable something like `pair_of_spaces`, and change `"same"` to `"same_order_1"`?

integration-tests/transport/test_conservative_transport.py
```diff
@@ -0,0 +1,184 @@
+"""
+Tests the conservative transport method, for a mixing ratio and density in the
+same and different spaces. This checks that there is a conservation of
```
tommbendall, line 3: It looks like this comment needs finishing!

tommbendall (changes requested):
Thanks so much for doing this Tim. The test looks great, I have a couple of very minor suggestions

PR #548, opened 2024-08-28
Fixes to some recovery related issues

tommbendall:
This fixes some issues related to the recovery operators and wrappers:
- allow the wrapper apply decorator to be used with implicit time discretisations
- tweak the boundary method code in the `RecoverySpaces` object
- enforcing the boundary conditions in the back-projection step in wrappers, which I found to be necessary to get the diffusion looking correct in the degree 0 Straka bubble


jshipton (approved):
Great - thanks @tommbendall !

PR #552, opened 2024-09-16
Add missing plotting scripts

tommbendall:
I must've forgotten to add and commit these plotting scripts in #541


jshipton (approved):
oops - thanks for adding now!

PR #553, opened 2024-09-23
Advective-then-Flux form for Runge-Kutta schemes

tommbendall:
This PR implements an "advective-then-flux" form for RK schemes with the conservative transport equation.

This involves:
- adding an enumerator for the formulation of the RK scheme -- we now have `linear`, `predictor` and `increment` forms, which follow the attached document
- I have implemented the `linear` form, which allows the final RK stage of the final subcycle to use a different operator
- there is a new `all_but_last` label, which indicates the form to use for all but the very final RK stage
- there is an `advective_then_flux` boolean argument to the `DGUpwind` operator, which signifies to use the `advective` form of the transport equation for all-but-the-last RK stage
- there is of course a unit-test which demonstrates that this formulation has the correct properties

This seems to have a similarly positive impact to the density/depth predictor for transport with large Courant numbers.

### DIFF
--- a/gusto/spatial_methods/transport_methods.py
+++ b/gusto/spatial_methods/transport_methods.py
@@ -8,8 +8,10 @@ from firedrake import (
 )
 from firedrake.fml import Term, keep, drop
 from gusto.core.configuration import IntegrateByParts, TransportEquationType
-from gusto.core.labels import (prognostic, transport, transporting_velocity, ibp_label,
-                               mass_weighted)
+from gusto.core.labels import (
+    prognostic, transport, transporting_velocity, ibp_label, mass_weighted,
+    all_but_last
+)
 from gusto.core.logging import logger
 from gusto.spatial_methods.spatial_methods import SpatialMethod
 
@@ -82,6 +84,10 @@ class TransportMethod(SpatialMethod):
 
             # Create new term
             new_term = Term(self.form.form, original_term.labels)
+
+            # Add all_but_last form
+            if hasattr(self, "all_but_last_form"):
+                new_term = all_but_last(new_term, self.all_but_last_form)
 
             # Check if this is a conservative transport
             if original_term.has_label(mass_weighted):
@@ -230,13 +236,13 @@ class DGUpwind(TransportMethod):
                     )
 
                 if advective_then_flux and vector_manifold_correction:
-                    self.form_for_early_stages = vector_manifold_advection_form(
+                    self.all_but_last_form = vector_manifold_advection_form(
                         self.domain, self.test, self.field, ibp=ibp,
                         outflow=outflow
                     )
 
                 elif advective_then_flux:
-                    self.form_for_early_stages = upwind_advection_form(
+                    self.all_but_last_form = upwind_advection_form(
                         self.domain, self.test, self.field, ibp=ibp,
                         outflow=outflow
                     )
--- a/gusto/spatial_methods/transport_methods.py
+++ b/gusto/spatial_methods/transport_methods.py
@@ -181,7 +181,7 @@ class DGUpwind(TransportMethod):
         self.outflow = outflow
 
         if (advective_then_flux
-            and self.transport_equation_type != TransportEquationType.conservative):
+                and self.transport_equation_type != TransportEquationType.conservative):
             raise ValueError(
                 'DG Upwind: advective_then_flux form can only be used with '
                 + 'the conservative form of the transport equation'
--- a/gusto/spatial_methods/transport_methods.py
+++ b/gusto/spatial_methods/transport_methods.py
@@ -175,7 +175,7 @@ class DGUpwind(TransportMethod):
                 transport equation for all but the last steps of some
                 (potentially subcycled) Runge-Kutta scheme, before using the
                 conservative form for the final step to deliver a mass-
-                conserving increment. This optiona only makes sense to use with
+                conserving increment. This option only makes sense to use with
                 Runge-Kutta, and should be used with the "linear" Runge-Kutta
                 formulation. Defaults to False, in which case the conservative
                 form is used for every step.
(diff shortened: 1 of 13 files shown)


gusto/spatial_methods/transport_methods.py
```diff
@@ -163,50 +170,82 @@ def __init__(self, equation, variable, ibp=IntegrateByParts.ONCE,
                 vector manifold correction term. Defaults to False.
             outflow (bool, optional): whether to include outflow at the domain
                 boundaries, through exterior facet terms. Defaults to False.
+            advective_then_flux (bool, optional): whether to use the advective-
+                then-flux formulation. This uses the advective form of the
+                transport equation for all but the last steps of some
+                (potentially subcycled) Runge-Kutta scheme, before using the
+                conservative form for the final step to deliver a mass-
+                conserving increment. This optiona only makes sense to use with
```
atb1995, line 178:
```suggestion
                conserving increment. This option only makes sense to use with
```

gusto/time_discretisation/explicit_runge_kutta.py
```diff
@@ -2,17 +2,46 @@
...
+    Enumerator to describe the formulation of a Runge-Kutta scheme.
+
+    The following Runge-Kutta methods for solving dy/dt = F(y) are encoded here:
+    - `increment`:                                                            \n
+        k_0 = F[y^n]                                                          \n
+        k_m = F[y^n - dt*Sum_{i=0}^{m-1} a_{m,i} * k_i], for m = 1 to M - 1   \n
+        y^{n+1} = y^n - dt*Sum_{i=0}^{M-1} b_i*k_i                            \n
+    - `predictor`:
+        y^0 = y^n                                                             \n
+        y^m = q^0 - dt*Sum_{i=0}^{m-1} a_{m,i} * F[y^i], for m = 1 to M - 1   \n
```
atb1995, line 34: Do you mean y_0 here? what is q_0?

gusto/time_discretisation/explicit_runge_kutta.py
```diff
@@ -2,17 +2,46 @@
...
+        y^{n+1} = y^0 - dt*Sum_{i=0}^{m-1} b_i * F[y^i]                       \n
+    - `linear`:
+        y^0 = y^n                                                             \n
+        y^m = q^0 - dt*F[Sum_{i=0}^{m-1} a_{m,i} * y^i], for m = 1 to M - 1   \n
```
atb1995, line 38: Same as above

gusto/time_discretisation/explicit_runge_kutta.py
```diff
@@ -114,45 +143,85 @@ def setup(self, equation, apply_bcs=True, *active_labels):
...
-        if self.increment_form:
+        if self.rk_formulation == RungeKuttaFormulation.increment:
             return super().solver
-        else:
+
+        elif self.rk_formulation == RungeKuttaFormulation.predictor:
             # In this case, don't set snes_type to ksp only, as we do want the
-            # outer Newton iteration
+            # outer Newton iteration. This is achieved by not calling the
+            # "super" method, in which the default snes_type is set to ksp_only
```
atb1995, line 165: I'm not sure we need a nonlinear solve here? ksp only should be sufficient.

gusto/time_discretisation/explicit_runge_kutta.py
```diff
@@ -114,45 +143,85 @@ def setup(self, equation, apply_bcs=True, *active_labels):
...
                 solver = NonlinearVariationalSolver(
                     problem, solver_parameters=self.solver_parameters,
-                    options_prefix=solver_name)
+                    options_prefix=solver_name
+                )
                 solver_list.append(solver)
             return solver_list
 
+        elif self.rk_formulation == RungeKuttaFormulation.linear:
+            # In this case, don't set snes_type to ksp only, as we do want the
```
atb1995, line 183: Same as above

gusto/time_discretisation/explicit_runge_kutta.py
```diff
@@ -267,6 +397,67 @@ def solve_stage(self, x0, stage):
...
+                    self.field_rhs.assign(
+                        self.field_rhs
+                        + self.butcher_matrix[stage, i]*self.field_lhs[i_cycle_stage]
+                    )
+
+                # Evaluate physics and apply limiter, if necessary
+                for evaluate in self.evaluate_source:
+                    evaluate(self.field_rhs, self.dt)
+                if self.limiter is not None:
+                    self.limiter.apply(self.field_rhs)
```
atb1995, line 424: Do we want to apply the limiter here, or after the solve?

atb1995 (changes requested):
These changes largely look good to me, the use of the enumerated type is sensible and the changes successfully implement an advective-then-flux version of conservative transport. I'm happy that the change has sufficient testing, I only have a few minor comments/requests.

gusto/time_discretisation/explicit_runge_kutta.py
tommbendall, line 34:
```suggestion
        y^m = y^0 - dt*Sum_{i=0}^{m-1} a_{m,i} * F[y^i], for m = 1 to M - 1   \n
```

gusto/time_discretisation/explicit_runge_kutta.py
tommbendall:
```suggestion
        y^m = y^0 - dt*F[Sum_{i=0}^{m-1} a_{m,i} * y^i], for m = 1 to M - 1   \n
```

gusto/time_discretisation/explicit_runge_kutta.py
tommbendall: Probably a very good point. I think we've decided to resolve this in #560

gusto/time_discretisation/explicit_runge_kutta.py
tommbendall: Probably a very good point. I think we've decided to resolve this in #560

gusto/time_discretisation/explicit_runge_kutta.py
tommbendall: It should be here, as it is applied on the right-hand side field that is evaluated. There is a final application of the limiter at the end after the final solve

atb1995 (approved):

Issue #554, opened 2024-09-25
Environment variables for logging require Python3.11


JHopeCollins:
The logging functionality allows setting the logging level with environment variables (`GUSTO_LOG_LEVEL` etc). However, these environment variables are processed using `logging.getLevelNamesMapping`
https://github.com/firedrakeproject/gusto/blob/bd9e2fe0be7d7bc872be654a9d6ac701013a81c7/gusto/core/logging.py#L73

This function was added in Python3.11 so that's a hard requirement if you want to use the environment variables.
https://docs.python.org/3/library/logging.html#logging.getLevelNamesMapping

Whereas `logging.getLevelNamesMapping` returns a dictionary with all the name mappings (e.g. `"DEBUG"` returns the `logging.DEBUG` enum), `logging.getLevelName` just returns the result for a single level name, and is available in all Python versions supported by Firedrake.
https://docs.python.org/3/library/logging.html#logging.getLevelName

Changing to this function would remove the Python3.11 requirement.

JHopeCollins:
Closed by #555

PR #555, opened 2024-09-25
Parse logging environment variables with pre Python3.11 function

JHopeCollins:
This addresses #554, see issue for details.


tommbendall (approved):
Thanks Josh, this looks like a simple fix

PR #560, opened 2024-10-22
Time Discretisation Improvements

atb1995:
This pull requests makes some time discretisation improvements, namely:

1. Limiters correctly applied for IMEX schemes (SDC and Runge-Kutta)
2. Solver initial guesses added to speed up model performance (also needed for implicit and IMEX schemes for compressible Euler equations)
3. For IMEX Runge-Kutta both nonlinear & linear solver options are now avaliable, for the nonlinear solve + final linear solve

### DIFF
--- a/gusto/time_discretisation/imex_runge_kutta.py
+++ b/gusto/time_discretisation/imex_runge_kutta.py
@@ -122,7 +122,7 @@ class IMEXRungeKutta(TimeDiscretisation):
         # Check all terms are labeled implicit, exlicit
         for t in self.residual:
             if ((not t.has_label(implicit)) and (not t.has_label(explicit))
-               and (not t.has_label(time_derivative))  and (not t.has_label(physics_label))):
+               and (not t.has_label(time_derivative)) and (not t.has_label(physics_label))):
                 raise NotImplementedError("Non time-derivative terms must be labeled as implicit or explicit")
 
         self.xs = [Function(self.fs) for i in range(self.nStages)]
@@ -253,7 +253,7 @@ class IMEXRungeKutta(TimeDiscretisation):
 
         # Apply limiter
         if self.limiter is not None:
-                self.limiter.apply(self.x_out)
+            self.limiter.apply(self.x_out)
         x_out.assign(self.x_out)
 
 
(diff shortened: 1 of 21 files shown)


gusto/time_discretisation/imex_runge_kutta.py
```diff
@@ -209,19 +229,31 @@ def final_solver(self):
...
     @wrapper_apply
     def apply(self, x_out, x_in):
         self.x1.assign(x_in)
+        self.x_out.assign(x_in)
         solver_list = self.solvers
 
         for stage in range(self.nStages):
             self.solver = solver_list[stage]
+            # Set initial solver guess
+            if (stage > 0):
```
tommbendall, line 243:
```suggestion
            if stage > 0:
```

tommbendall (changes requested):
Thanks Alex, these look like good general improvements. I can't get my head around exactly what a limiter does for an implicit scheme, but it looks like it is being applied at sensible places.

The thing that makes me most nervous is the application of guesses for the RK scheme! But I think what you've done looks right

gusto/time_discretisation/imex_runge_kutta.py
```diff
@@ -102,7 +122,7 @@ def setup(self, equation, apply_bcs=True, *active_labels):
         # Check all terms are labeled implicit, exlicit
         for t in self.residual:
             if ((not t.has_label(implicit)) and (not t.has_label(explicit))
-               and (not t.has_label(time_derivative))):
+               and (not t.has_label(time_derivative)) and (not t.has_label(physics_label))):
```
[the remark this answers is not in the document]
atb1995: I have removed this after clarifying that it isn't needed with `SplitPhysicsTimestepper`

gusto/time_discretisation/time_discretisation.py
```diff
@@ -374,6 +382,13 @@ def setup(self, equation, apply_bcs=True, *active_labels):
             self.ncycles = 1
         self.x0 = Function(self.fs)
         self.x1 = Function(self.fs)
+        # If the time_derivative term is nonlinear, we must use a nonlinear solver
+        if (len(self.residual.label_map(lambda t: t.has_label(nonlinear_time_derivative), map_if_false=drop)) > 0
```
tommbendall, line 386: Could we separate this over multiple lines?

tommbendall (changes requested):

gusto/time_discretisation/time_discretisation.py
atb1995: Done!

tommbendall (approved):

PR #567, opened 2024-10-24
Improve clarity of SWSaturationAdjustment class

nhartney:
This pull requests makes some improvements to the layout of the SWSaturationAdjustment class. It clarifies the order of the source terms and the test functions as `water_vapour`, `cloud_water`, (`D`), (`b`), where `D` is added for convective feedback and `b` for thermal feedback. This alignment between the order of source terms and test functions ensures that the correct terms are added to the residual. This pull request does not change what happens in the code but changes the layout slightly and adds some comments with the aim of making this clear.


tommbendall (approved):
This all looks good to me

PR #570, opened 2024-11-04
Update workflow to use node20 instead of node16

ta440:
Github actions now uses node20 instead of node16. 
In line with this:

- `actions/checkout@v3` is updated to `actions/checkout@v4`
-  `actions/setup-python@v4` is updated to `actions/setup-python@v5`
- `actions/configure-pages@v3` is updated to `actions/configure-pages@v5`
- `actions/deploy-pages@v2` is updated to `actions/deploy-pages@v4`

The current flake8 notations action rbialon/flake8-annotations@v1 still uses node16 and there doesn't appear to be an updated version of this that uses node20.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -29,7 +29,7 @@ jobs:
     # Steps represent a sequence of tasks that will be executed as
     # part of the jobs
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


tommbendall (approved):
Nothing controversial here, thank you Tim for being on-the-ball and sorting this!

PR #573, opened 2024-11-14
Thermal sw

jshipton:
Separate out thermal shallow water and include equivalent buoyancy formulation.


gusto/core/configuration.py
```diff
@@ -147,6 +147,15 @@ class ShallowWaterParameters(Configuration):
     g = 9.80616
     Omega = 7.292e-5  # rotation rate
     H = None  # mean depth
+    # Factor that multiplies the vapour in the equivalent buoyancy
```
tommbendall, line 150: I think these parameters (`beta2`, `nu` and `q0`) are all features of having moisture rather than the equivalent buoyancy, so could we omit the part about the equivalent buoyancy formulation in these comments?

gusto/equations/shallow_water_equations.py
```diff
@@ -13,24 +13,26 @@
...
     Class for the (rotating) shallow-water equations, which evolve the velocity
     'u' and the depth field 'D', via some variant of:                         \n
-    ∂u/∂t + (u.∇)u + f×u + g*∇(D+b) = 0,                                      \n
+    ∂u/∂t + (u.∇)u + f×u + g*∇(D+B) = 0,                                      \n
     ∂D/∂t + ∇.(D*u) = 0,                                                      \n
-    for Coriolis parameter 'f' and bottom surface 'b'.
+    for Coriolis parameter 'f' and bottom surface 'B'.
     """
 
     def __init__(self, domain, parameters, fexpr=None, bexpr=None,
```
tommbendall, line 32: Should we change `bexpr` to `Bexpr` or something like `topog_expr` to help avoid confusion?

gusto/equations/shallow_water_equations.py
```diff
@@ -456,3 +654,11 @@ def __init__(self, domain, parameters, fexpr=None,
 
         # Use the underlying routine to do a first linearisation of the equations
         self.linearise_equation_set()
+
+
+def compute_saturation(q0, nu, H, g, D, b, B=None):
```
tommbendall, line 659: Should we make this a method of the `ThermalShallowWaterEquations`? It could return `None` is there are no active tracers. I don't know if it would be better to be have a mixed field `X` as the argument or the individual fields, but this would also give access to the parameters and `B`

gusto/equations/shallow_water_equations.py
```diff
@@ -272,7 +232,245 @@ def __init__(self, domain, parameters, fexpr=None, bexpr=None,
...
+    'u' and the depth field 'D', via some variant of either:                  \n
+    ∂u/∂t + (u.∇)u + f×u + b*∇(D+B) + 0.5*D*∇b= 0,                            \n
+    ∂D/∂t + ∇.(D*u) = 0,                                                      \n
+    ∂b/∂t + u.∇(b) = 0,                                                       \n
+    for Coriolis parameter 'f', bottom surface 'B' and buoyancy field b,
+    or, if equivalent_buoyancy=True:
+    ∂u/∂t + (u.∇)u + f×u + b_e*∇(D+B) + 0.5*D*∇(b_e + beta_2 q_v)= 0,         \n
+    ∂D/∂t + ∇.(D*u) = 0,                                                      \n
+    ∂b_e/∂t + u.∇(b_e) = 0,                                                   \n
+    ∂q_t/∂t + u.∇(q_t) = 0,                                                   \n
```
tommbendall, line 253: This matches the code below which requires us to have the `q_t` moisture variable when using the equivalent buoyancy formulation. I had imagined we would do it with an `ActiveTracer` object but this makes sense to me

gusto/equations/shallow_water_equations.py
```diff
@@ -272,7 +232,245 @@ def __init__(self, domain, parameters, fexpr=None, bexpr=None,
...
+            residual += topography_form
+
+        # -------------------------------------------------------------------- #
+        # Linearise equations
+        # -------------------------------------------------------------------- #
+        # Add linearisations to equations
+        self.residual = self.generate_linear_terms(residual, self.linearisation_map)
+
+    def setup(self):
+        self.field_names = ['u', 'D']
```
tommbendall, line 394: Do we need this `setup` function? Could we just set `self.field_names` and `self.space_names` in the `__init__` method?

gusto/equations/shallow_water_equations.py
```diff
@@ -177,53 +155,38 @@ def __init__(self, domain, parameters, fexpr=None, bexpr=None,
...
-                                                    'u'), self.X))
-            residual += source_form
-
         # -------------------------------------------------------------------- #
         # Linearise equations
         # -------------------------------------------------------------------- #
         # Add linearisations to equations
         self.residual = self.generate_linear_terms(residual, self.linearisation_map)
 
+    def setup(self):
```
tommbendall, line 169: Do we need this `setup` function? Could we just set `self.field_names` and `self.space_names` in the `__init__` method?

tommbendall (changes requested):
I'm super excited about this. My main questions are whether we can move the `compute_saturation` routine to be a method of the `ThermalShallowWaterEquations` , and about the `setup` method

gusto/core/configuration.py
jshipton: hmm the only time we get the parameters from here is when we use the equivalent buoyancy formulation... @nhartney does it make sense to get the moisture parameters from here in other cases? Are there more parameters that could go here?

gusto/equations/shallow_water_equations.py
jshipton: yes

gusto/equations/shallow_water_equations.py
jshipton: @nhartney in your branch this is in its own special file all by itself - I was wondering why that was? Is it used anywhere else?

gusto/equations/shallow_water_equations.py
jshipton: Yes, I wondered about how to do this... this way requires less work for the user and does the equivalent buoyancy formulation even mean anything without the moisture??

gusto/equations/shallow_water_equations.py
jshipton: ugh, yeah, I don't like this either and there might be a better way of doing it... The problem is that the mixed function space is set up in the `PrognosticEquationSet` `__init__` method which gets  passed `field_names` and `spaces` from the `ShallowWaterEquations` `__init__` method... so if we set those things in the `__init__` method of `ThermalShallowWaterEquations` then they get overwritten in `ShallowWaterEquations`.

gusto/equations/shallow_water_equations.py
jshipton: same comment as above - happy to talk about alternatives because I don't like this either!

tommbendall: Ah I understand, this is because the `ThermalShallowWaterEquations` inherit from the `ShallowWaterEquations`, and we go through the whole normal process of setting up the shallow water equations and then modify them...

It does feel like unnecessary effort to avoid calling the parent `__init__` method and duplicate all of the momentum equation in the `ThermalShallowWaterEquations` class.

If we required the list of field names and the dictionary of function spaces to be `@abstractproperty`s of all `PrognosticEquationSet`s would that work?

examples/shallow_water/linear_thermal_galewsky_jet.py
```diff
@@ -0,0 +1,164 @@
+from gusto import *
```
tommbendall, line 1: I realise this example has just been added here for now and maybe you're still planning to do this, but we will need to make it match the format of the other examples to ensure that it gets run through the testing -- i.e. it will need a `__main__` method and use arg-parse so that it can be called from the command line with different arguments.

To be consistent with everything else, we'd also need a plotting script and a figure, although I think we could set up an issue and do that in a different PR

examples/shallow_water/linear_thermal_galewsky_jet.py
```diff
@@ -0,0 +1,164 @@
...
+    # Initial D field is calculated by integrating D_integrand w.r.t. phi
+    # Assumes the input is between phi0 and phi1.
+    # Note that this function operates on vectorized input.
+    from numpy import exp, sin, tan
+    f = 2.0*parameters.Omega*sin(th)
+    u_zon = (80.0/en)*exp(1.0/((th - phi0)*(th - phi1)))
+    return u_zon*(f + tan(th)*u_zon/R)
+
+
+def Dval(X):
```
tommbendall, line 98: We could use our `NumericalIntegral` code here, e.g. like this galewsky jet case study: https://github.com/firedrakeproject/gusto_case_studies/blob/main/case_studies/shallow_water/galewsky_jet.py#L128

gusto/core/configuration.py
nhartney: I think it doesn't make sense to get the moisture parameters from here if we aren't using the equivalent buoyancy formulation, and I don't think there are more parameters that could go here. We need no moisture parameters for thermal shallow water (usual formulation), and if we are doing moist thermal shallow water then the moisture parameters will be passed through with the physics scheme.

gusto/equations/shallow_water_equations.py
nhartney: There wasn't a good reason to have it in a file on its own - I did it only because I didn't know where it should belong. I also use the function elsewhere to set up some of the examples.

examples/shallow_water/linear_thermal_galewsky_jet.py
nhartney: I did just put this here for now but I can make those changes if we decide we want this example to stay.

tommbendall: I think it would be a nice example to have -- I think we should have something with the linear thermal equations

gusto/core/configuration.py
tommbendall: I actually am happy about keeping the parameters in here -- and maybe we should put other moist shallow water parameters in here too. We currently keep other moisture parameters in the parameters object for the compressible equations

gusto/equations/shallow_water_equations.py
tommbendall: Yes I agree with you both!

gusto/core/configuration.py
nhartney: Oh yes, that's true. And I was wrong when I said that there are no other parameters we could go here - if we wanted to cover all our bases for moist shallow water options we'd need to add a `beta1` parameter for convective feedback from moisture (to allow us to use the moist convective thermal and the moist convective pseudo-thermal equations). Then there's also the thresholds for phase changes and the proportions converted, but I feel that they don't belong with the equations though they are moist shallow water parameters.

nhartney:
We are also going to need to either adapt the existing `ThermalSWSolver` in `linear_solvers.py` to work with the equivalent buoyancy formulation or add a separate linear solver to do this, or else use something other than the SIQN scheme as a timestepper with the equivalent buoyancy formulation.

gusto/equations/shallow_water_equations.py
```diff
@@ -272,7 +232,245 @@ def __init__(self, domain, parameters, fexpr=None, bexpr=None,
...
+            lambda t: t.has_label(pressure_gradient),
+            drop)
+
+        # add (moist) thermal source terms not involving topography -
+        # label these as the equivalent pressure gradient term
+        if equivalent_buoyancy:
+            q0 = parameters.q0
+            nu = parameters.nu
+            beta2 = parameters.beta2
+            qsat_expr = compute_saturation(q0, nu, H, g, D, b_name, topog)
```
nhartney, line 341: I think we want `b` here as the argument rather than `b_name`.

gusto/equations/shallow_water_equations.py
```diff
@@ -272,7 +232,245 @@ def __init__(self, domain, parameters, fexpr=None, bexpr=None,
...
+        # -------------------------------------------------------------------- #
+        # drop usual pressure gradient term
+        residual = self.residual.label_map(
+            lambda t: t.has_label(pressure_gradient),
+            drop)
+
+        # add (moist) thermal source terms not involving topography -
+        # label these as the equivalent pressure gradient term
+        if equivalent_buoyancy:
+            q0 = parameters.q0
```
nhartney, line 338: I think `q0`, `nu` and `beta2` need to all be a Firedrake `Constant`.

nhartney (changes requested):

gusto/equations/shallow_water_equations.py
jshipton: because they are now coming from the parameters class, they are already `Constant`s (see line 81 of configuration.py)

gusto/equations/shallow_water_equations.py
jshipton: yes, good spot!

gusto/equations/shallow_water_equations.py
```diff
@@ -272,7 +240,279 @@ def __init__(self, domain, parameters, fexpr=None, bexpr=None,
...
+        q0 = self.parameters.q0
+        nu = self.parameters.nu
+        g = self.parameters.g
+        H = self.parameters.H
+        D, b = split(X)[1:3]
+        if topog is None:
+            sat_expr = q0*H/(D) * exp(nu*(1-b/g))
+        else:
+            sat_expr = q0*H/(D+topog) * exp(nu*(1-b/g))
+            return sat_expr
```
nhartney, line 444: Is this an indentation mistake?

gusto/equations/shallow_water_equations.py
```diff
@@ -272,7 +240,278 @@ def __init__(self, domain, parameters, fexpr=None, bexpr=None,
...
+            # Default linearisation is time derivatives, pressure gradient,
+            # Coriolis and transport term from depth and buoyancy equation
+            linearisation_map = lambda t: \
+                (any(t.has_label(time_derivative, pressure_gradient, coriolis))
+                 or (t.get(prognostic) in ['D', self.b_name]
+                     and t.has_label(transport)))
+
+        super().__init__(domain, parameters,
+                         equivalent_buoyancy=equivalent_buoyancy,
+                         fexpr=fexpr, bexpr=bexpr, space_names=space_names,
```
nhartney, line 510: `bexpr` is now `topog_expr` in the `init` method of the super class.

tommbendall (approved):
This is so exciting! I think it's ready now, well done @nhartney and @jshipton

PR #575, opened 2024-11-18
Fix Gusto following changes to value_size and value_shape

tommbendall:
Updates parts of Gusto getting `value_shape` or `value_size` from the `ufl_element`, to now get it from the function space directly. This fixes the broken main branch following the Firedrake PR: https://github.com/firedrakeproject/firedrake/pull/3862


jshipton (approved):
Thanks for the quick fix @tommbendall !!

PR #579, opened 2024-11-26
Allow use of SUPG for mixed function spaces

atb1995:
Currently SUPG does not work for mixed function spaces. Here I have allowed for the altering of multiple test functions. 

Note that MixedFSWrapper is not used, as altering a single test function in the tuple of test functions for a MixedFunctionSpace was difficult to do.

There is scope in future to not use SUPG as a wrapper, but this is not addressed in this pull request.
moist_bf_theta_e_-1
moist_bf_theta_e_supg3_-1


gusto/time_discretisation/time_discretisation.py
```diff
@@ -219,31 +226,33 @@ def setup(self, equation, apply_bcs=True, *active_labels):
 
             else:
                 if self.wrapper_name == "supg":
-                    self.wrapper.setup()
+                    for field_name in self.wrapper_field_names:
+                        self.wrapper.setup(field_name)
+                        new_test = self.wrapper.test
+                        self.residual = self.residual.label_map(
+                            lambda t: t.get(prognostic) == field_name and t.has_label(transport),
```
tommbendall, line 233: can we change this so that we are testing whether the label is not `time_derivative`? I think it's better if the time discretisation to not know what terms it contains

atb1995: If we apply it to the momentum eq, then it will try to replace the test function for all terms with u in, which will be problematic..

PR #580, opened 2024-11-28
Get SUPG working for MixedFunctionSpace

atb1995:
Currently SUPG does not work for mixed function spaces. Here I have allowed for the altering of multiple test functions. 

Note that MixedFSWrapper is not used, as altering a single test function in the tuple of test functions for a MixedFunctionSpace was difficult to do.

There is scope in future to not use SUPG as a wrapper, but this is not addressed in this pull request.
moist_bf_theta_e_-1
moist_bf_theta_e_supg3_-1


gusto/core/configuration.py
```diff
@@ -200,8 +200,9 @@ class SUPGOptions(WrapperOptions):
 
     name = "supg"
     tau = None
-    default = 1/sqrt(15)
+    default = 1/(sqrt(15))
```
tommbendall, line 203:
```suggestion
    default = 1/sqrt(15)
```

gusto/time_discretisation/sdc.py
```diff
@@ -163,9 +163,9 @@ def __init__(self, base_scheme, domain, M, maxk, quad_type, node_type, qdelta_im
 
         # Get Q_delta matrices
         self.Qdelta_imp = genQDeltaCoeffs(qdelta_imp, form=formulation,
-                                          nodes=self.nodes, Q=self.Q)
+                                          nodes=self.nodes, Q=self.Q, nNodes=M, nodeType=node_type, quadType=quad_type)
```
tommbendall, line 166: Just to check, is this a change that should be part of this PR?

gusto/time_discretisation/time_discretisation.py
```diff
@@ -219,31 +223,42 @@ def setup(self, equation, apply_bcs=True, *active_labels):
 
             else:
                 if self.wrapper_name == "supg":
-                    self.wrapper.setup()
+                    if self.wrapper_field_names is not None:
+                        for field_name in self.wrapper_field_names:
+                            self.wrapper.setup(field_name)
+                            new_test = self.wrapper.test
+                            self.residual = self.residual.label_map(
+                                lambda t: t.get(prognostic) == field_name and t.has_label(transport),
```
tommbendall, line 231: I generally don't like specific terms appearing in this file because I think the `time_discretisation` should be agnostic to the type of term being solved...

... if we need to specifically only replace the test function in the transport terms, could the `SUPGOptions`  object contain a list of labels for the terms that need replacing? Then the user can specify which terms are going to be adjusted

gusto/time_discretisation/time_discretisation.py
```diff
@@ -219,31 +223,42 @@ def setup(self, equation, apply_bcs=True, *active_labels):
...
-                self.residual = self.wrapper.label_terms(self.residual)
 
         # -------------------------------------------------------------------- #
         # Make boundary conditions
         # -------------------------------------------------------------------- #
 
         if not apply_bcs:
             self.bcs = None
-        elif self.wrapper is not None:
+        elif self.wrapper is not None and self.wrapper_name != "supg":
```
tommbendall, line 261: Is this amendment to the if statement definitely still needed? (I'm happy to keep it if it is)

gusto/time_discretisation/wrappers.py
```diff
@@ -381,9 +387,13 @@ def setup(self):
         # -------------------------------------------------------------------- #
         # Set up test function
         # -------------------------------------------------------------------- #
+        if self.time_discretisation.wrapper_field_names is not None:
+            uadv = Function(domain.spaces('HDiv'))
```
tommbendall, line 391: could this line be above the if statement?

gusto/time_discretisation/time_discretisation.py
```diff
@@ -101,6 +101,10 @@ def __init__(self, domain, field_name=None, solver_parameters=None,
             elif self.wrapper_name == "recovered":
                 self.wrapper = RecoveryWrapper(self, options)
             elif self.wrapper_name == "supg":
+                if options.field_names:
```
tommbendall, line 104: Could this be `if options.field_names is not None`:

gusto/core/configuration.py
```diff
@@ -200,8 +200,9 @@ class SUPGOptions(WrapperOptions):
...
     ibp = IntegrateByParts.TWICE
+    field_names = []
```
tommbendall, line 205: It might be more natural to have `None` as the default option here

e.g. : https://nikos7am.com/posts/mutable-default-arguments/#:~:text=The%20solution%20and%20the%20standard%20way%20of%20doing,list_to_append.%20append%28element%29%20return%20list_to_append%20%3E%3E%3E%20b%20%3D%20append_to_list%2820%29

gusto/time_discretisation/sdc.py
atb1995: Nothing gets past you.. they are required for the parallel preconditioners, so thought I would just get it onto main. Can remove and put in a separate PR?

gusto/time_discretisation/time_discretisation.py
atb1995: For mixed FS yes

gusto/core/configuration.py
atb1995: Done!

gusto/time_discretisation/time_discretisation.py
atb1995: I've instead just done it as `self.wrapper_field_names = options.field_names` and the None check is now further down

gusto/time_discretisation/wrappers.py
atb1995: Done, put it below

gusto/time_discretisation/time_discretisation.py
atb1995: Okay agreed, I have added a list of term labels to cycle through. I need to add a test for this so hang fire for that

gusto/core/configuration.py
```diff
@@ -202,6 +202,7 @@ class SUPGOptions(WrapperOptions):
     tau = None
     default = 1/sqrt(15)
     ibp = IntegrateByParts.TWICE
+    suboptions = None # Dictionary containing field_name and term_labels
```
tommbendall, line 205: As well as fixing the lint, can you expand on this comment just to make it a bit clearer what `suboptions` should be, i.e. the keys of the dictionary should be strings for the names of the fields, and the values should be a list of terms to apply the SUPG test function to

tommbendall (approved):
Thanks Alex!

PR #587, opened 2024-12-14
Add a script to automatically update the KGOs (and update KGOs)

tommbendall:
Adds a script which automatically updates the KGO values


jshipton (approved):
I have confirmed that this works on my laptop! Thanks @tommbendall !!

PR #588, opened 2024-12-16
Possibly a small typo in readme url?

sit23:
There was a slash in the install URL where there should be a `.` Just checking if this is deliberate or not?

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ These compatible methods underpin the Met Office's next-generation model, [LFRic
 
 The best way to install Gusto is as an additional package when installing [Firedrake](http://firedrakeproject.org). Usually, for a Mac with Homebrew or an Ubuntu installation this is done by downloading the Firedrake install script and executing it:
 ```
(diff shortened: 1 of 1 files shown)


jshipton (approved):
thanks @sit23 !

README.md
```diff
@@ -16,7 +16,7 @@ These compatible methods underpin the Met Office's next-generation model, [LFRic
 
 The best way to install Gusto is as an additional package when installing [Firedrake](http://firedrakeproject.org). Usually, for a Mac with Homebrew or an Ubuntu installation this is done by downloading the Firedrake install script and executing it:
 ```
-curl -0 https://raw.githubusercontent/com/firedrakeproject/firedrake/master/scripts/firedrake-install
+curl -0 https://raw.githubusercontent.com/firedrakeproject/firedrake/master/scripts/firedrake-install
```
JHopeCollins, line 19: Adding a comment here because I may not remember by the time I get down to Exeter tomorrow.

It might be better not to duplicate this command here but instead just have the pointer to the Firedrake install page below, and only the gusto flag here. That way if the Firedrake installation method changes (which it might do soon!) you don't have to remember to update it here.

PR #590, opened 2024-12-16
Add a pyproject.toml

connorjward:
Note that the `requirements.txt` should probably stay for `firedrake-install` to remain supported (which will go away soon when we have `pip install firedrake`).


tommbendall (approved):
Thanks Connor!

PR #595, opened 2024-12-16
Restrict quadrature degree to some maximum value

tommbendall:
A newer version of #530 to restrict the quadrature degree for complicated expressions


tommbendall:
I think this is actually quite a hard problem!

I don't think we do this as is, as this will very often increase the quadrature degree, as for most forms it will be lower than 5!

To make things more confusing, we also currently already restrict the quadrature degree for the compressible linear solver, which does make me wonder how much benefit we'll get..

Possible thoughts:
- we could move the `max_quad_deg` into domain, and make it a function of the polynomial degrees
- we could make the measure replacement a utility routine, not part of the time stepper, and it could be called selectively (e.g. if we have a domain with orography)
- our worst quadrature degrees are when the Exner function appears -- we could just hard-code the quadrature degree for those terms

tommbendall:
Closing as this is now superseded by #599

PR #599, opened 2024-12-17
Limit quadrature degree for terms involving the Exner pressure

tommbendall:
This picks up on the changes from #530 and #595, and introduces a `max_quad_degree` belonging to the domain. This is used to fix the quadrature degree for terms involving some particularly non-polynomial expression -- the Exner pressure is a particular culprit here.

This is hopefully a small optimisation for the forcing terms and hydrostatic balance (we already did this for the linear solver).


jshipton (dismissed):

jshipton (approved):
I still approve

PR #602, opened 2024-12-18
SDC: Fix wrapper and augmentation + Add in MIN-SR-FLEX

atb1995:
Small quality of life fixes. Inherits wrapper + augmentation from base time discretisation. Add in an option to update implicit Qdelta for a specific type of preconditioner


tommbendall (approved):
The best PRs remove more lines than they add!

PR #606, opened 2024-12-18
Add SSPRK schemes with higher stage count

JHopeCollins:
These schemes have more stages than formal order, but in return have higher CFL limits. The CFL limit increases faster, so the CFL/stage ratio is improved for the higher stage count schemes.

e.g. the 3 stage 3rd order has CFL=1, i.e. effective CFL 1/3 per stage
4 stage 3rd order has CFL=2, so effective CFL 1/2 per stage.


colinjcotter:
Aren’t the critical Courant numbers dependent on polynomial degree?

gusto/time_discretisation/explicit_runge_kutta.py
```diff
@@ -532,19 +532,37 @@ def __init__(
...
+    k0 = F[y^n]                                                               \n
+    k1 = F[y^n + (1/2)*dt*k1]                                                 \n
+    k2 = F[y^n + (1/2)*dt*(k0+k1)]                                            \n
+    k3 = F[y^n + (1/6)*dt*(k0+k1+k2)]                                         \n
+    y^(n+1) = y^n + (1/6)*dt*(k0 + k1 + k2 + 3*k3)                            \n
+
+    The 5-stage method can be written as:                                     \n
+
+    k0 = F[y^n]                                                               \n
+    k1 = F[y^n + (1/2)*dt*k1]                                                 \n
```
tommbendall, line 556: I guess there is no neat way of writing the 4-stage scheme with rational numbers, so we could just omit this?

JHopeCollins:
> Aren’t the critical Courant numbers dependent on polynomial degree?

These are from this paper:
10.1137/S0036142901389025

The extra dofs from the additional stages can be used to increase the cfl limit

colinjcotter:
Those critical Courant numbers are for first order upwind I think.

There is a much more recent paper by Ern with bigger stable timesteps for DG.

JHopeCollins:
SSP CFL limits are always with the caveat of "only for spatial schemes which are stable with forward Euler" aren't they?

I will plot out the stability regions for the different schemes in the morning and we can see what they look like.

Do you have a link to that paper?

tommbendall (approved):
Thanks Josh, this could be really helpful for pushing the time step and doing subcycling

colinjcotter:
Ern, Alexandre, and Jean-Luc Guermond. "Invariant-domain-preserving high-order time stepping: I. Explicit Runge--Kutta schemes." SIAM Journal on Scientific Computing 44, no. 5 (2022): A3366-A3392.

JHopeCollins:
Looking at the stability plots, the higher stage count schemes do have larger stability regions along the imaginary axis as well as elsewhere, so they have higher CFL limits even for spatial schemes higher than 1st order (i.e. eigenvalues closer to the imaginary axis), even if they're no longer SSP. But the stage=order methods also lose SSP property if the spatial scheme isn't TVD.
So yes, exactly how much the efficiency would change for non-TVD spatial schemes would need to be either worked out for the specific scheme or found empirically, it wouldn't necessarily be the same as the effective SSP CFL.

Third order methods:
SSPRK3_stability
And along the imaginary axis:
SSPRK3_stability_imag

Fourth order, 5 stage method compared to the classic RK4 method:
SSPRK4_stability
Along the imaginary axis:
SSPRK4_stability_imag

PR #609, opened 2025-01-13
Test CI fix

connorjward:
I think this requires a fix to be merged upstream in loopy.


tommbendall:
Thanks very much for dealing with that @connorjward

PR #610, opened 2025-01-20
Fix Courant number calculation

tommbendall:
Fixes the Courant number calculation which is breaking tests following this Firedrake PR: https://github.com/firedrakeproject/firedrake/pull/3947


tommbendall:
Unfortunately there seem to be other failures here...

1) `boundary_layer_mixing` tests are failing with `ufl.algorithms.check_arities.ArityMismatch`

2) `test_zero_limiter.py` and some thermal shallow water tests, failing with a petsc error in the linear solve

3) `test_moist_compressible`, `test_simultaneous_SIQN` (order 1), `test_boussinesq.py` (compressible) all failing with changed KGOs

jshipton:
Rerunning because the boundary layer test works fine on my laptop after a Firedrake update...

jshipton:
> Rerunning because the boundary layer test works fine on my laptop after a Firedrake update...

Ah well.

tommbendall:
I think we're confident that we need these fixes, and I'd like to get a clearer picture of our remaining errors so I'm going to just merge this as is

PR #613, opened 2025-02-07
Reduce pyproject.toml boilerplate

connorjward:
This fix gets automatic package discovery working so it isn't necessary to specify all subpackages.


jshipton (approved):
Nice - thanks @connorjward !

PR #614, opened 2025-02-12
Copy active tracers in SIQN

nhartney:
This changes how the list of tracers to copy (tracers that are not included in the linear solver) is formed in the SIQN scheme. Active tracers are identified as `active_tracer` in `equation_set`, instead of by whether or not their time derivative term has a `linearisation` label.


jshipton (approved):
Good spot - thanks @nhartney

PR #616, opened 2025-02-13
Horizontal & vertical splitting of advective transport terms, and a DG_Upwind version

atb1995:


gusto/equations/common_forms.py
```diff
@@ -346,3 +346,69 @@ def tracer_conservative_form(test, q, rho, ubar):
     form = transporting_velocity(L, ubar)
 
     return transport(form, TransportEquationType.tracer_conservative)
+
+
+def split_hv_advective_form(equation, field_name):
+    u"""
+    Splits advective term into horizontal and vertical terms.
+    This describes splitting u.∇(q) terms into u.(∇_h)q and w dq/dz,
+    for transporting velocity u and transported q.
```
tommbendall, line 355: I think this all looks correct. It's very cumbersome having to go through the whole of this process -- do you think there are any shortcuts we can take, e.g. in adding the linearisations?

gusto/spatial_methods/transport_methods.py
```diff
@@ -111,8 +111,41 @@ def replace_form(self, equation):
                 map_if_true=lambda t: new_term)
 
         else:
-            raise RuntimeError('Found multiple transport terms for '
```
tommbendall, line 114: Again I'm a bit worried about just removing this! Can we have a special check for the horizontal/vertical case?

gusto/spatial_methods/transport_methods.py
```diff
@@ -111,8 +111,41 @@ def replace_form(self, equation):
...
+            horizontal_term = transporting_velocity.update_value(horizontal_term, new_horizontal_transporting_velocity)
+            vertical_term = transporting_velocity.update_value(vertical_term, new_vertical_transporting_velocity)
+
+            # Create new term
+            new_horizontal_term = Term(self.form_h.form, horizontal_term.labels)
+            new_vertical_term = Term(self.form_v.form, vertical_term.labels)
+
+            # Check if this is a conservative transport
+            if horizontal_term.has_label(mass_weighted) or vertical_term.has_label(mass_weighted):
+                raise RuntimeError('Mass weighted transport terms not yet supported for multiple terms')
```
tommbendall, line 139: Should this be a `NotImplementedError`? Do you think it would be worth raising an issue to capture this debt once implemented?

gusto/spatial_methods/transport_methods.py
```diff
@@ -277,9 +310,133 @@ def __init__(self, equation, variable, ibp=IntegrateByParts.ONCE,
         self.form = form
 
 
+class Split_DGUpwind(TransportMethod):
```
tommbendall, line 313:
```suggestion
class SplitDGUpwind(TransportMethod):
```

tommbendall (changes requested):
Looks good -- unfortunately I don't instantly spot what the issue might be on the sphere. Don't feel you have to respond to every comment

gusto/core/labels.py
```diff
@@ -92,6 +92,8 @@ def __call__(self, target, value=None):
 # ---------------------------------------------------------------------------- #
 implicit = Label("implicit")
 explicit = Label("explicit")
+horizontal = Label("horizontal")
```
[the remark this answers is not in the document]
atb1995: Done

gusto/spatial_methods/transport_methods.py
atb1995: Done!

gusto/spatial_methods/transport_methods.py
atb1995: Done and issue raised #627

gusto/spatial_methods/transport_methods.py
atb1995: Done!

gusto/spatial_methods/spatial_methods.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, equation, variable, term_label):
             map_if_true=keep, map_if_false=drop)
 
         num_terms = len(self.original_form.terms)
-        assert num_terms == 1, f'Unable to find {term_label.label} term ' \
+        assert num_terms >= 1, f'Unable to find {term_label.label} term ' \
```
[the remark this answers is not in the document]
atb1995: I've changed this to loop through a list of term_labels, but I'm not sure if this is the best solution!

gusto/equations/common_forms.py
atb1995: I've tried to tidy it a bit by splitting out the advection forms, hopefully this helps!

gusto/spatial_methods/spatial_methods.py
```diff
@@ -15,19 +15,20 @@ class SpatialMethod(object):
     The base object for describing a spatial discretisation of some term.
     """
 
-    def __init__(self, equation, variable, term_label):
+    def __init__(self, equation, variable, term_labels):
```
tommbendall, line 18: We discussed this offline but putting it here for posterity: could we avoid some of the complicated logic below by using `*term_labels` here, or similar?

Can we still enforce that the original labelled form still only has one valid term to be replaced?

integration-tests/transport/test_split_dg_transport.py
```diff
@@ -0,0 +1,67 @@
...
+from gusto import *
+import pytest
+
+
+def run(timestepper, tmax, f_end):
+    timestepper.run(0, tmax)
+    return norm(timestepper.fields("f") - f_end) / norm(f_end)
+
+
+@pytest.mark.parametrize("geometry", ["slice", "sphere"])
```
tommbendall, line 17: Actually this sphere is only 2D so not really testing the splitting, so shall we remove this?

tommbendall (changes requested):
Two remaining things!

integration-tests/transport/test_split_dg_transport.py
atb1995: All done!

tommbendall (approved):
Thanks Alex!

PR #617, opened 2025-02-17
Fix package discovery

connorjward:
Sorry I got it wrong last time! This works locally for me and will be tested in https://github.com/firedrakeproject/firedrake/pull/4011.


tommbendall (approved):
This obviously looks fine to me, always fun to have single character changes!

The CI failed (obviously not related to this change) but possibly from something parallel-unsafe that has already been fixed in Firedrake? So I'll just see if the CI passes next time...

PR #618, opened 2025-02-19
Test using pip install Firedrake

connorjward:
In https://github.com/firedrakeproject/firedrake/pull/4011 there are a number of breaking changes to our Docker containers that affect downstream users. In particular:
* The container no longer contains a virtual environment
* The container ships OpenMPI instead of MPICH, so parallel tests need to be run in separate steps.

What I have done here is very similar to what [we already have in Firedrake](https://github.com/firedrakeproject/firedrake/blob/master/.github/workflows/build.yml).

This is a draft until the pip stuff is merged, but I want to have an example to show other users what they will need to do for their CI.


.github/workflows/build.yml
```diff
@@ -25,60 +25,131 @@ jobs:
...
-          python -m pip install -e .
-          python -m pip install  \
-            pytest-timeout pytest-xdist
-      - name: Gusto tests
+          . venv-gusto/bin/activate
+          pip install ./gusto-repo
+          : # Extra test dependencies
+          sudo apt-get update
+          sudo apt-get install -y parallel
+          pip install pytest-split pytest-timeout pytest-xdist
```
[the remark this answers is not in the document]
connorjward: I suppose. My thinking was that these dependencies are not strictly needed for Firedrake to be able to run its tests. Maybe we could have a `firedrake[ci]` set of optional dependencies?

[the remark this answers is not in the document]
JHopeCollins: I think there's a distinction between:

1. Absolutely bare minimum requirements to run the tests in some way or other.
2. Requirements to run the test in a "normal/common/standard" way (whatever that means)
3. Requirements to run the tests in the "firedrake recommended" way (if different to 2.)

Yes these are not strictly needed for (1). But `timeout` and `xdist` are probably needed for (2), and they are all clearly needed for (3), the "firedrake-approved" way of running CI.

connorjward, line 76:
```suggestion
```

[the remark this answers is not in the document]
connorjward: These are all now part of the `firedrake-vanilla` container.

connorjward:
This is currently failing because the logs are not getting created/uploaded correctly. I don't understand what is happening there. @JDBetteridge do you know what needs fixing? What do these logs contain?

.github/workflows/build.yml
```diff
@@ -25,60 +25,116 @@ jobs:
...
+      - name: Run tests (nprocs = 4)
+        if: success() || steps.install-two.conclusion == 'success'
+        run: |
+          . venv-gusto/bin/activate
+          firedrake-run-split-tests 4 3 "$EXTRA_PYTEST_ARGS"
+        timeout-minutes: 10
+
       - name: Prepare logs
-        if: always()
+        if: success() || steps.install-two.conclusion == 'success'
```
JDBetteridge, line 121: I don't think this step should be changed from `always()`, if an above step fails it doesn't matter if this step fails (likewise below for the upload)

.github/workflows/build.yml
```diff
@@ -25,60 +25,116 @@ jobs:
...
+        id: prepare-logs
         run: |
           mkdir logs
           cd /tmp/pytest-of-firedrake/pytest-0/
```
JDBetteridge, line 125: It looks like the location of the files has changed, maybe due to not being in a venv now? Where does pytest dump logs when running the docker container locally?

JDBetteridge (dismissed):

.github/workflows/build.yml
connorjward: The logic here is saying to only try and upload the pytest logs if we successfully install Gusto. That seems like the right thing to do?

.github/workflows/build.yml
connorjward: But are they just the pytest logs? If so we have a better solution in Firedrake already. `firedrake-run-split-tests` outputs files called things like `pytest_nprocs1_job1.log` that we upload as artifacts [here](https://github.com/firedrakeproject/firedrake/blob/master/.github/workflows/build.yml#L182). Should we just do that here as well?

.github/workflows/build.yml
JDBetteridge: You are running tests before "install 2" though

connorjward: It's more complicated now that we call pytest lots of different times. I guess we could upload the logs for the initial tests but there are so few tests being run and the same information is available in stdout. The logs are mostly helpful for when stdout is too polluted to understand.

JDBetteridge: Gusto logs contain information that is not printed to stdout and is useful for debugging

connorjward: And were those previously found in `/tmp/pytest-of-firedrake/...`? I am confused about whether these log files we are uploading are the output from pytest or the output of something else.

.github/workflows/build.yml
JDBetteridge: I thought the gusto logs were also ending up there, but honestly can't remember off the top of my head. Something like if you call with xdist they end up there, maybe? I think it's best to check

connorjward: I've just looked at the logs from the latest run and the output looks quite confusing as each runner seems to be combining their logs without noting which test is being run. So yes you're right that the uploaded logs are not stdout but they're also very hard to understand.

This looks like it would be quite hard to recover this behaviour as we now call pytest about 20 times instead of just once so we have a lot more log files that can potentially conflict.

connorjward: I have a suggestion: if Gusto need to see the log information would it be acceptable to require them to set `GUSTO_PARALLEL_LOG=CONSOLE` in the `build.yml`? The output is then still available if needed.

I can leave a comment in `build.yml` to that effect.

tommbendall: That definitely sounds acceptable to me. Does that mean that logged calls will not be uploaded as artifacts, but will be appear in the terminal-style output for the github action?

connorjward: Yes I think so. I will trigger a run now so we can see what that looks like.

To make things clear. The workflow I have *will* upload log artifacts at the end of a run. Those artifacts are just the output of *pytest*, not Gusto.

connorjward: Actually I think I may have a better solution. Give me a few minutes.

.github/workflows/build.yml
connorjward: Resolving as this is the same discussion as another thread.

.github/workflows/build.yml
connorjward: Yep it works! Now there are 2 artifacts that get uploaded:
* pytest-logs: the pytest output from each job for each set of tests
* gusto-logs: the equivalent Gusto logging output

connorjward:
@tommbendall never mind. I need to fix the docs build too.

connorjward:
This is ready to go

jshipton (approved):
Thanks so much @connorjward !!

tommbendall (approved):
I'm definitely happy with this. Thanks very much for preparing this for us!

jshipton:
@JDBetteridge are you happy to approve the changes? I'm reluctant to override as I wasn't sure what the issue was!

JDBetteridge (approved):

PR #628, opened 2025-04-09
Diagnostics bugfix

Witt-D:
Just a quick bugfix of the Pressure diagnositc in which the lower boundary condition was being incorrectly applied in the (1, 0) order case.


tommbendall:
Closing as a duplicate of #629

PR #629, opened 2025-04-09
bugfixes for boundary conditions

Witt-D:
FIxing boundary condition triggers so that they are correctly applied to the (1, 0) case.


tommbendall (approved):
All good, thanks Dan!

PR #630, opened 2025-04-11
Fix failing boundary layer mixing test

tommbendall:
Some changes to UFL have caused the boundary layer mixing physics scheme to fail, because it was detecting different shapes of variables between two different outcomes of a `conditional` statement.

The hack to fix this is to ensure that they agree on the shape by adding a field multiplied by zero to the expression.


atb1995 (approved):
Looks good to me, leaving the comment to explain it is helpful! Approved

PR #631, opened 2025-04-11
Update Boussinesq KGO

tommbendall:
A change in an underlying package in the last couple of days (I can't see an obvious culprit) has caused one of our KGO tests to fail. The tolerance has only slightly been exceeded and things still look OK, so I think this is a case where the KGO just needs updating.


atb1995 (approved):
Slightly concerning that the KGOs have changed, but it is only slightly below a tight tolerance.

PR #633, opened 2025-04-23
Stop using docdeps container in docs workflow

connorjward:
The container is soon to be deprecated in Firedrake.


tommbendall (approved):
Thanks Connor, this makes sense

PR #636, opened 2025-06-09
Use firedrake-vanilla-default container for docs build

connorjward:
The docdeps container was accidentally reintroduced in https://github.com/firedrakeproject/gusto/commit/0992440f051dcce6497677e289b1fe15bf3cbf23.


connorjward:
Great thanks!

PR #637, opened 2025-06-25
Fixes for new interpolate behaviour

leo-collins:
Firedrake is in the process of moving interpolation behaviour from `firedrake.__future__` into Firedrake main (pull request [#4346](https://github.com/firedrakeproject/firedrake/pull/4346)). There were still some lingering uses of the old `Interpolator` in gusto which were flagged in our smoke tests. This PR fixes that so all interpolation is done with the new symbolic operator. 

Once [#4346](https://github.com/firedrakeproject/firedrake/pull/4346) is merged, all `from firedrake.__future__ import interpolate` can be removed.


gusto/physics/boundary_and_turbulence.py
```diff
@@ -141,7 +142,9 @@ def __init__(self, equation, T_surface_expr, vapour_name=None,
                 dmv_expr = surface_expr * (mv_np1_expr - m_v) / self.dt
                 source_mv_expr = test_m_v * self.source_mv * dx
 
-                self.source_interpolators.append(Interpolator(dmv_expr, self.source_mv_int))
+                self.source_interpolators.append(
+                    lambda expr=dmv_expr: assemble(interpolate(dmv_expr, self.source_theta_vd_int), tensor=self.source_mv_int)
```
connorjward, line 146:
```suggestion
                    lambda: assemble(interpolate(dmv_expr, self.source_theta_vd_int), tensor=self.source_mv_int)
```
?

gusto/recovery/recovery.py
```diff
@@ -272,7 +272,9 @@ def __init__(self, x_in, x_out, method='interpolate', boundary_method=None):
                     # so need to extract component and restore it after the boundary recovery is done
                     for i in range(V_out.value_size):
                         x_out_scalars.append(Function(CG1))
-                        self.interpolate_to_scalars.append(Interpolator(self.x_out[i], x_out_scalars[i]))
+                        self.interpolate_to_scalars.append(
+                            lambda i=i: assemble(interpolate(self.x_out[i], CG1), tensor=x_out_scalars[i])
```
connorjward, line 276: Why the `i=i`?

connorjward (commented):
Thanks Leo! Just a couple of queries about `lambda`

gusto/recovery/recovery.py
leo-collins: Something to do with scope: https://stackoverflow.com/questions/938429/scope-of-lambda-functions-and-their-parameters

Only the last value of `i` was being passed

gusto/physics/boundary_and_turbulence.py
connorjward: This is the same lambda query. I can believe that I was wrong in this suggestion.

jshipton (approved):
This looks good to me - thanks @leo-collins and @connorjward !!

gusto/recovery/recovery.py
```diff
@@ -221,7 +221,7 @@ def __init__(self, x_in, x_out, method='interpolate', boundary_method=None):
 
         self.method = method
         if method == 'interpolate':
-            self.broken_op = Interpolator(x_in, x_brok)
+            self.broken_op = lambda: assemble(interpolate(x_in, V_brok), tensor=x_brok)
```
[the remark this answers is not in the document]
leo-collins: They're equivalent since if we pass a `Function` to `interpolate` we just grab the `FunctionSpace` it is defined on.

PR #639, opened 2025-07-01
Small updates to Schaer mountain example

ta440:
Mesh was missing as an argument from `CompressibleParameters()` and `SpongeLayerParameters()`.


tommbendall (approved):
Thanks Tim!

PR #641, opened 2025-07-17
`interpolate` fixes

leo-collins:
After the next release `interpolate` won't accept a `Function` in the second slot


tommbendall (approved):
Thanks very much for doing this @leo-collins, and sorry for the delay in reviewing

PR #646, opened 2025-08-07
Potential issue in equivalent buoyancy formulation of the ThermalSWSolver class

nhartney:
This flags a possible issue in the formulation of the `ThermalSWSolver` when using the equivalent buoyancy formulation. The notation uses the variable `b` to represent both buoyancy and equivalent buoyancy. In either formulation (equivalent buoyancy or not) the field actually used in the solver should always be `b`. Then when using the equivalent buoyancy formulation the `b_e` field recovered from the equation set probably should be converted to `b` before being used by the solver. This currently happens with the reference profiles but not the fields.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,11 @@ Gusto's default development happens on the `main` branch, which is fixed to the 
 We also maintain a `future` branch which builds from the head of Firedrake's `main` branch.
 Contributions to Gusto's `main` branch must pass our Continuous Integration tests.
 
+The relationship between `future` and `main` is as follows:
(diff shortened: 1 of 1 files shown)


nhartney:
Other things that might have changed, so that we have a record:

- The original `MoistDynamicsSWSolver` used the input fields `xn` rather than the reference fields `X_ref` in calculating the saturation function and the water vapour - we think that the current approach of using `X_ref` should be better.
- Previously the reference profiles were not updated at each timestep and currently they are.

PR #659, opened 2025-09-25
Add pressure transport term to Boussinesq eq description

ta440:
Add an advection form pressure transport term to the compressible Boussinesq equation description. A very minor change, but helpful for looking at the code!


jshipton (approved):
another good spot @ta440 !!

PR #664, opened 2025-11-06
First run I/O

nhartney:
This PR changes the set-up of characteristics of the I/O. It adds a `first_run` flag to the time stepper's `run` method, which is `True` until the components of the I/O have been set up, after which it becomes `False`. This is to facilitate solving a PDE in Gusto multiple times as part of a hybrid ML model.

### DIFF
--- a/gusto/core/io.py
+++ b/gusto/core/io.py
@@ -390,9 +390,10 @@ class IO(object):
         raise_parallel_exception = 0
         error = None
 
-        if any([self.output.dump_vtus, self.output.dump_nc,
+        setup_dir = any([self.output.dump_vtus, self.output.dump_nc,
                 self.output.dumplist_latlon, self.output.dump_diagnostics,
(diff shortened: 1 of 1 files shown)


gusto/timestepping/timestepper.py
```diff
@@ -180,7 +180,7 @@ def log_field_stats(self):
                 max_val = field_data.max()
                 logger.debug(f'{field_name}, min: {min_val}, max: {max_val}')
 
-    def run(self, t, tmax, pick_up=False):
+    def run(self, t, tmax, pick_up=False, first_run=True):
```
connorjward: I know it's not really my place to review Gusto code but I feel this might be more elegant if this were called `init_io` or similar. Or you could even set this attribute inside `Timestepper.__init__`.

jshipton: yes, this is a nice idea @connorjward  - thanks! your comments are always appreciated!

PR #665, opened 2025-11-11
Refactor CI to work with Firedrake release and main

connorjward:


.github/workflows/test.yml
```diff
@@ -1,31 +1,28 @@
-name: Build Gusto
-
 on:
-  # Push to main or PR
-  push:
-    branches:
-      - main
-  pull_request:
-  schedule:
```
tommbendall: Can we keep this schedule so that we test `main` and `future` once per week?

tommbendall (changes requested):
Thanks very much for doing this Connor. My only comment is whether we can keep our scheduled weekly runs?

tommbendall (approved):

PR #675, opened 2025-11-24
fix for courant number calculation

jshipton:


JHopeCollins:
Are there any Courant number tests? If not then might be useful adding some to catch this in the future.
It's hard to test exhaustively but there are some tests for simple cases here that you could adapt to gusto: https://github.com/firedrakeproject/asQ/blob/master/tests/utils/test_diagnostics.py

PR #679, opened 2025-11-25
Separate out two different weekly tests

tommbendall:
This might be a very clumsy way to do this, but I want to avoid one set of tests being cancelled by the submission of the other set.

This fixes weekly testing.


tommbendall:
> It's probably easiest to duplicate the `schedule.yml` file into `schedule-main.yml` and `schedule-future.yml`. I didn't realise that one would cancel the other.

That's a more sensible idea, thanks!

PR #681, opened 2025-12-04
Inner loop physics

nhartney:
This PR adds the capability to evaluate physics schemes in the inner loop of the semi-implicit quasi-Newton time stepping scheme. Physics is evaluated both explicitly and implicitly, in a similar way to forcing. The parameter `physics_beta` controls the semi-implicit off-centring, where a `physics_beta` of 1 is fully implicit and a `physics_beta` of 0 is fully explicit.

The PR also changes the names of the physics schemes lists, replacing the pre-existing `physics_schemes` with `final_physics_schemes`. 

Finally, the pull request introduces a moist shallow water linear solver, `MoistThermalSWSolver`. This includes moisture in the linear solve stage and is necessary when doing physics in the inner loop. The solver is only suitable for use in a saturated atmosphere where the saturation function has the usual dependence on depth and buoyancy.

These changes are tested with a new inner loop physics test that uses the moist solver.

### DIFF
--- a/gusto/solvers/linear_solvers.py
+++ b/gusto/solvers/linear_solvers.py
@@ -824,7 +824,7 @@ class MoistThermalSWSolver(TimesteppingSolver):
 
         # Check that the third field is buoyancy
         if not equation.field_names[2] == 'b':
-            raise NotImplementedError("Field 'b'must exist to use the moist thermal linear solver in the SIQN scheme")
+            raise NotImplementedError("Field 'b' must exist to use the moist thermal linear solver in the SIQN scheme")
         if not equation.field_names[3] == 'water_vapour':
             raise NotImplementedError("Field 'water_vapour' must exist to use the moist thermal linear solver in the SIQN scheme")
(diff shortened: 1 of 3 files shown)


gusto/solvers/linear_solvers.py
```diff
@@ -771,6 +771,186 @@ def solve(self, xrhs, dy):
...
+        P_expr = (
+            qv - sat_expr*(-D/Dbar - b*nu/g + qv*nu*beta2/g)
+        )
+
+        n = FacetNormal(equation.domain.mesh)
+
+        eqn = (
+            # u equation
+            inner(w, (u - u_in)) * dx
+            - beta_u * (D - Dbar) * div(w*bbar) * dx
```
tommbendall: Alex pointed out to me that terms only involve reference values and no trial functions shouldn't be here
```suggestion
            - beta_u * (D) * div(w*bbar) * dx
```

tommbendall (approved):

PR #683, opened 2025-12-08
Use the right comm attribute for minmax kernels

connorjward:
Small fixup to go with https://github.com/firedrakeproject/firedrake/pull/4766.

The change here is simply that Firedrake objects no longer have a private `_comm` attribute. The handling of user-facing vs internal comms is handled completely internally.


tommbendall (approved):
Very straightforward -- thanks for doing this

PR #684, opened 2025-12-14
Be more sensible about tolerances

tommbendall:


jshipton (approved):
thanks @tommbendall !

PR #685, opened 2025-12-16
Use a different approach for detecting whether a mesh is extruded

connorjward:
In Firedrake `mesh._base_mesh` is now set to `None`, so using `hasattr()` is no longer valid.

~Follows https://github.com/firedrakeproject/firedrake/pull/4780~ (merged)


tommbendall (approved):
Thanks as always Connor!

PR #688, opened 2025-12-18
Fix broken examples

tommbendall:
Several examples seemed to be giving strange results:
- Williamson 2 is not preserving a solid body rotation
williamson_2_main_final

- the dry Bryan and Fritsch rising bubble is not rising symmetrically
dry_bryan_fritsch_tau

This PR fixes those by:
1. removing the `rotated_pole` from the Williamson 2 test, and from the rest of the infrastructure. I think this test broke when we reorganised the Coriolis force, as the Coriolis force no longer knew about the rotated pole. But I think the rotated pole infrastructure was always unnecessary -- a better way of rotating the pole is to rotate the mesh after it is originally generated. The best way to tackle this is to remove all of the `rotated_pole` infrastructure.
2. Turning off boundary recovery for the wind field in the lowest-order bubble tests. The boundary recovery process for vector function spaces involves separating the vector-valued field into scalar functions, performing the boundary recovery for each component, and then returning to the vector-valued field. Somehow this process is giving odd results. I will investigate in slower time, but until then we can safely turn this off.

I have added plots for the changed tests which demonstrate that they are working correctly.

### DIFF
--- a/gusto/recovery/recovery.py
+++ b/gusto/recovery/recovery.py
@@ -294,13 +294,15 @@ class Recoverer(object):
         if self.boundary_method is not None:
             # For vector elements, treat each component separately
             if self.vector_function_space:
+                # TODO: something about interpolating to scalars and then back
+                # is making the lowest-order bubble tests assymetric!
                 for i, boundary_recoverer in enumerate(self.boundary_recoverers):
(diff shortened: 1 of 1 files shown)


jshipton (approved):
Thanks @tommbendall  - sorry, I completely forgot about the rotated pole stuff!!

PR #707, opened 2026-02-23
Gamma plane

jshipton:
Create gamma plane option for shallow water and enable specification of a "trap" in the Coriolis parameter via the `ShallowWaterEquation` class.


gusto/equations/shallow_water_equations.py
```diff
@@ -201,7 +211,16 @@ def _setup_residual(self, u_transport_option):
                 fexpr = self.parameters.f0
             elif rotation is CoriolisOptions.betaplane:
                 xyz = SpatialCoordinate(self.domain.mesh)
-                fexpr = self.parameters.f0 + self.parameters.beta * xyz[1]
+                fexpr = self.parameters.f0 + self.parameters.beta * (
+                    xyz[1] - self.parameters.y0)
+            elif rotation is CoriolisOptions.gammaplane:
+                x, y = SpatialCoordinate(self.domain.mesh)
+                r, _ = rtheta_from_xy(x, y)
```
12shughes: It's also used in here, so currently my gamma plane and trap is centred on a corner of the domain. This will probably be different for different people using it, but I imagine you'd typically want a gamma plane centred in the centre of the mesh?

PR #722, opened 2026-04-09
Stop using deprecated constructor for WithGeometry

connorjward:


jshipton (approved):
thanks @connorjward !

PR #725, opened 2026-04-22
Merge in future

tommbendall:


connorjward:
It's recommend to avoid squashing commits when you do this. Otherwise the relation to the PRs introducing the changes is lost.

tommbendall:
> It's recommend to avoid squashing commits when you do this. Otherwise the relation to the PRs introducing the changes is lost.

OK, thanks. future doesn't merge cleanly into main (which is partly my fault for accidentally merging some commits into main last month that I had to revert) so there's not an easy option

connorjward:
I just tried and the only conflicts were some `.h5` files
```
On branch connorjward/merge-future
You have unmerged paths.
  (fix conflicts and run "git commit")
  (use "git merge --abort" to abort the merge)

Changes to be committed:
	modified:   gusto/core/io.py
	modified:   gusto/model/model.py
	modified:   gusto/rexi/rexi.py
	modified:   gusto/rexi/rexi_coefficients.py
	new file:   integration-tests/data/linear_sw_wave_rexi_caliari_chkpt.h5
	deleted:    integration-tests/data/linear_sw_wave_rexi_chkpt.h5
	new file:   integration-tests/data/linear_sw_wave_rexi_haut_chkpt.h5
	modified:   integration-tests/data/update_kgos.py
	modified:   integration-tests/rexi/test_linear_compressible_boussinesq.py
	modified:   integration-tests/rexi/test_linear_sw.py
	modified:   integration-tests/rexi/test_rexi_approximations.py
	modified:   unit-tests/recovery_tests/test_hcurl_boundary_kernel.py

Unmerged paths:
  (use "git add <file>..." to mark resolution)
	both modified:   integration-tests/data/linear_sw_wave_chkpt.h5
	both modified:   integration-tests/data/sw_fplane_chkpt.h5
```
If I knew which of `main` or `future` was correct for those files then things would work. I can open another PR if it would be helpful.

tommbendall:
The future ones should be the correct ones, but if they aren't right then we can have a follow-up PR to fix them. My other concern was that the changes from #706 aren't showing up in the diff for `future`, but do with this new branch

connorjward:
> The future ones should be the correct ones, but if they aren't right then we can have a follow-up PR to fix them. My other concern was that the changes from #706 aren't showing up in the diff for `future`, but do with this new branch

You're right. That's weird. Something bespoke is likely needed then.

tommbendall:
The only test not passing is a file not being in the right place in the checkpointing test. I don't think that's an issue on `future` and so would need fixing separately anyway, so I'm going to merge this in

PR #727, opened 2026-04-23
Fix failing multi-checkpoint test

tommbendall:
The behaviour of the 'a' permission for `CheckpointFile` seems to have recently changed, and now doesn't create a file. This was causing our multi-checkpointing test to fail.

This PR fixes that, by ensuring that the checkpoint file has been created before proceeding.


jshipton (approved):
Thanks @tommbendall !

PR #728, opened 2026-04-24
Minor updates to complex proxy

JHopeCollins:
1. use `mesh.unique()` for both vector and mixed implementations of FunctionSpace
2. stricter FunctionSpace test to catch more `mesh.unique` errors when constructing the proxy space
3. In the solve tests set boundary conditions on a component space that actually has boundary values.
4. Small rearrangement of BC construction to match asQ so keep it easy to compare diffs


jshipton (approved):
Thanks @JHopeCollins !!

PR #729, opened 2026-04-27
solve hybridised linear system using Fieldsplit+ILU to do the elimination rather than SCPC

jshipton:
…so we get to reuse the factorisation

### DIFF
--- a/gusto/solvers/preconditioners.py
+++ b/gusto/solvers/preconditioners.py
@@ -2,7 +2,7 @@
 
 from firedrake import (dot, jump, dS_h, ds_b, ds_t, ds,
                        FacetNormal, Tensor, AssembledVector,
-                       AuxiliaryOperatorPC, PETSc, RieszMap)
+                       AuxiliaryOperatorPC, PETSc, AssembledPC, RieszMap)
 
 from firedrake.preconditioners import PCBase
 from firedrake.matrix_free.operators import ImplicitMatrixContext
@@ -12,7 +12,7 @@ from pyop2.profiling import timed_region, timed_function
 from functools import partial
 
 
-__all__ = ["VerticalHybridizationPC", "AuxiliaryPC", "CompressibleHybridisedSCPC"]
+__all__ = ["VerticalHybridizationPC", "AuxiliaryPC", "CompressibleHybridisedSCPC", "SlateSchurPC"]
 
 
 class AuxiliaryPC(AuxiliaryOperatorPC):
@@ -665,6 +665,7 @@ class CompressibleHybridisedSCPC(PCBase):
         hybridized_prb = LinearVariationalProblem(
             aeqn, Leqn, self.y_hybrid, constant_jacobian=True
         )
+
         self.hybridized_solver = LinearVariationalSolver(
             hybridized_prb,
             options_prefix=pc.getOptionsPrefix()+self._prefix,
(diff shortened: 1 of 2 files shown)


gusto/solvers/preconditioners.py
```diff
@@ -2,7 +2,7 @@
 
 from firedrake import (dot, jump, dS_h, ds_b, ds_t, ds,
                        FacetNormal, Tensor, AssembledVector,
-                       AuxiliaryOperatorPC, PETSc, RieszMap)
+                       AuxiliaryOperatorPC, PETSc, AssembledPC, RieszMap)
```
JHopeCollins:
```suggestion
                       AuxiliaryOperatorPC, PETSc, RieszMap)
```

gusto/solvers/preconditioners.py
```diff
@@ -789,3 +790,32 @@ def applyTranspose(self, pc, x, y):
...
+        a10 = a.blocks[nf0, :nf0]
+        a01 = a.blocks[:nf0, nf0]
+        a11 = a.blocks[nf0, nf0]
+
+        schur_comp = a11 - a10*a00.inv*a01
+
+        return (schur_comp, None)
+
+    def view(self, pc, viewer=None):
+        super(AssembledPC, self).view(pc, viewer)
```
JHopeCollins: This is quite old python style. It should be fine without any arguments to `super`.
```suggestion
        super().view(pc, viewer)
```

gusto/solvers/solver_presets.py
```diff
@@ -97,28 +97,73 @@ def trace_nullsp(T):
...
+            'pc_type': 'fieldsplit',
+            'pc_fieldsplit_type': 'schur',
+            'pc_fieldsplit_schur_fact_type': 'full',
+            'pc_fieldsplit_0_fields': '0,1',
+            'pc_fieldsplit_1_fields': '2',
+            'fieldsplit_0': {
+                'ksp_type': 'preonly',
+                'pc_type': 'python',
+                'pc_python_type': 'firedrake.AssembledPC',
+                'assembled_pc_type': 'ilu',
```
JHopeCollins: This will work in both parallel and serial.
```suggestion
                'assembled_pc_type': 'bjacobi',
                'assembled_sub_pc_type': 'ilu',
```

gusto/solvers/solver_presets.py
```diff
@@ -97,28 +97,73 @@ def trace_nullsp(T):
...
+        #        }
+        #    }
+        # }
+
+        trace_params = {
+            'pc_type': 'ksp',
+            'ksp': {
+                # 'ksp_converged_rate': None,
+                'ksp_rtol': 1e-10,
+                'ksp_type': 'gmres',
```
JHopeCollins: Is there a particular reason for changing the parameters for the trace solve from the ones above?

Look like a lighter smoother on the levels, its great if it works as well!

jshipton: no reason - they're just copy-pasted from the ones you were using... do you think we should check them more widely before changing, or should I revert to the original? at the moment we recover KGOs, but I've not run the case studies... @tommbendall, what do you think?

JHopeCollins: Ah ok, that's fine then. I'd advocate for not changing the solver options for the trace system here so that this PR is only changing the way the elimination/back substitution is computed.

If there are better default options for the trace system then we could certainly look at that separately from the Fieldsplit vs SCPC comparison.

jshipton: OK I think I've put them back to the previous settings - would be great if you could check @JHopeCollins!

gusto/solvers/solver_presets.py
```diff
@@ -97,25 +97,43 @@ def trace_nullsp(T):
...
-            'pc_type': 'python',
-            'pc_python_type': 'firedrake.SCPC',
-            'pc_sc_eliminate_fields': '0,1',
-            # The reduced operator is not symmetric
-            'condensed_field': {
-                'ksp_type': 'fgmres',
+        trace_params = {
+            'pc_type': 'ksp',
+            'ksp': {
+                # 'ksp_converged_rate': None,
```
JHopeCollins:
```suggestion
```

JHopeCollins (dismissed):
One commented line left in, but other than that this looks good to me! Thanks!

JHopeCollins (dismissed):

gusto/solvers/solver_presets.py
```diff
@@ -97,25 +97,42 @@ def trace_nullsp(T):
...
+        }
+
+        slate_schur_params = {
+            'mat_type': 'matfree',
+            'ksp_type': 'preonly',
+            'pc_type': 'fieldsplit',
+            'pc_fieldsplit_type': 'schur',
+            'pc_fieldsplit_schur_fact_type': 'full',
+            'pc_fieldsplit_0_fields': '0,1',
+            'pc_fieldsplit_1_fields': '2',
```
tommbendall, line 123:
```suggestion
            # The reduced operator is not symmetric
            'pc_fieldsplit_1_fields': '2',
```
I suggest keeping this comment in case it's useful in the future

tommbendall (approved):

PR #731, opened 2026-04-30
Allow outputting to an existing directory

JHopeCollins:
Give gusto permission to write output into an existing directory

In particular this means you can call `stepper.run` multiple times in the same script.


JHopeCollins:
@jshipton this is a small fix to avoid having any hacky workarounds for running the DA scripts where you have several steppers and/or several calls to the same stepper.

It's about as unobtrusive as I could make it but you might want a more comprehensive approach to dealing with non-standard writing of results. Ideally it would be possible to run without any outputs but that would touch a lot more code.

jshipton (approved):
Great - thanks @JHopeCollins, this is indeed minimal.

PR #733, opened 2026-04-30
CI: consistent timeouts for pytest and gh actions

JHopeCollins:
There are two timeout limits set in the gusto CI, firstly a _per test_ limit given to pytest, and secondly a _per job step_ limit
given to github actions.
Currently the pytest limit is 1hr, so pytest will only timeout if a single tests takes more than an hour. However the per step limits given to gh actions are 1 hr or less, so the pytest limit is never hit.

This means that if a test takes too long or hangs the job gets killed by github, and we don't see any useful output. This PR reduces the pytest timeout limits so that hopefully hanging tests will be killed by pytest and we will see what test hung and where it hung.

Example timeout jobs without useful output:
https://github.com/firedrakeproject/gusto/actions/runs/25160405858/job/73753067758#step:11:726
https://github.com/firedrakeproject/gusto/actions/runs/24886988087/job/72869313732?pr=728

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ jobs:
       PYOP2_CFLAGS: -O0
       # Make sure that tests with >4 processes are not silently skipped
       PYTEST_MPI_MAX_NPROCS: 4
-      EXTRA_PYTEST_ARGS: --durations=100 --timeout=3600 --timeout-method=thread -o faulthandler_timeout=3660 --show-capture=no --verbose gusto-repo/unit-tests gusto-repo/integration-tests gusto-repo/examples
+      EXTRA_PYTEST_ARGS: --durations=100 --timeout-method=thread --show-capture=no --verbose gusto-repo/unit-tests gusto-repo/integration-tests gusto-repo/examples
     steps:
       - name: Fix HOME
         # For unknown reasons GitHub actions overwrite HOME to /github/home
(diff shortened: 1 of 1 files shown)


connorjward (approved):
Makes sense

JHopeCollins:
Well. This is vexing. The CI for this PR timed out in exactly the same way:
https://github.com/firedrakeproject/gusto/actions/runs/25162897434/job/73761596727?pr=733

Looking at the output it looks like all ranks actually finished fine - there are 4x2=8 ranks and 8 "passed" lines. So it must be hanging somewhere in the pytest finalisation itself weirdly.
I still think this PR is the right thing to do, but maybe it won't help with these particular timeouts.

JHopeCollins:
@connorjward I'm seeing lots of these warnings for the `nprocs=2` tests. Is this concerning? I thought pnetcdf was installed by this point in the CI?

>   /__w/gusto/gusto/venv-gusto/lib/python3.12/site-packages/gusto/core/io.py:1004: UserWarning: Serial netCDF in use even though you are running in parallel. This is especially inefficient at high core counts. Please refer to the documentation for information about installing a parallel version of netCDF.

connorjward:
> @connorjward I'm seeing lots of these warnings for the `nprocs=2` tests. Is this concerning? I thought pnetcdf was installed by this point in the CI?
> 
> > /__w/gusto/gusto/venv-gusto/lib/python3.12/site-packages/gusto/core/io.py:1004: UserWarning: Serial netCDF in use even though you are running in parallel. This is especially inefficient at high core counts. Please refer to the documentation for information about installing a parallel version of netCDF.

That's vexing. That shouldn't be happening.

connorjward:
> Well. This is vexing. The CI for this PR timed out in exactly the same way: https://github.com/firedrakeproject/gusto/actions/runs/25162897434/job/73761596727?pr=733
> 
> Looking at the output it looks like all ranks actually finished fine - there are 4x2=8 ranks and 8 "passed" lines. So it must be hanging somewhere in the pytest finalisation itself weirdly. I still think this PR is the right thing to do, but maybe it won't help with these particular timeouts.

When did these timeouts appear? And yes it seems like it's hanging when the interpreter is being finalised.

JHopeCollins:
> When did these timeouts appear? And yes it seems like it's hanging when the interpreter is being finalised.

They are not deterministic either, the current "fix" is just to rerun the job and it usually then passes.
This job is the first one I knew of but @jshipton may know if they have been happening for longer.
https://github.com/firedrakeproject/gusto/actions/runs/24886988087/job/72869313732?pr=728

jshipton:
> > When did these timeouts appear? And yes it seems like it's hanging when the interpreter is being finalised.
> 
> They are not deterministic either, the current "fix" is just to rerun the job and it usually then passes. This job is the first one I knew of but @jshipton may know if they have been happening for longer. https://github.com/firedrakeproject/gusto/actions/runs/24886988087/job/72869313732?pr=728

yes, they have - it's really annoying

connorjward:
Do they coincide with the 2026.4.0 release? I may be able to track them down at some point

jshipton:
> Do they coincide with the 2026.4.0 release? I may be able to track them down at some point

I don't think so... although I may be conflating this with a different issue that sporadically occurred that looked like it was related to netCDF... @tommbendall, do you recall the details better than me?

tommbendall:
I only remember seeing this in the last couple of weeks, but can't remember if I first saw it before or after the release

JHopeCollins:
@jshipton @tommbendall even if this doesn't actually prevent this particular timeout, I still think that its the right thing to do because it will still cause better failure modes for other hangs.

PR #734, opened 2026-05-01
New method for calculating vorticity diagnostic

tommbendall:
This implements a new method for calculating vorticity diagnostics for shallow-water equations.

We currently solve a linear variational problem to evaluate the vorticity in H1 from the velocity in HDiv, which is equivalent to the "weak curl". I am wondering if this can give very noisy results on the cubed-sphere...

... so this PR implements a different method, which first projects the velocity into HCurl, and then interpolates the "strong curl" of that velocity into a DG space.

### DIFF
--- a/gusto/diagnostics/shallow_water_diagnostics.py
+++ b/gusto/diagnostics/shallow_water_diagnostics.py
@@ -3,9 +3,11 @@
 
 from firedrake import (
     dx, TestFunction, TrialFunction, grad, inner, curl, Function, assemble,
-    LinearVariationalProblem, LinearVariationalSolver, conditional, interpolate
+    LinearVariationalProblem, LinearVariationalSolver, conditional, interpolate,
+    Projector
 )
 from gusto.diagnostics.diagnostics import DiagnosticField, Energy
+from gusto.core.function_spaces import is_cg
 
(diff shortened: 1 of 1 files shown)


gusto/diagnostics/shallow_water_diagnostics.py
```diff
@@ -159,19 +161,38 @@ def setup(self, domain, state_fields, vorticity_type=None):
             raise ValueError(f"vorticity type must be one of {vorticity_types}, not {vorticity_type}")
         space = domain.spaces("H1")
 
-        u = state_fields("u")
+        # Work out continuity requirements of space
+        if self.space is None or not is_cg(self.space):
+            # Using DG, calculate u in HCurl space as intermediary
+            self.hcurl_intermediary = True
+            assert self.method != 'solve', (
```
jshipton, line 171: This assertion is causing test failures... I'm not quite sure what you intend here...

tommbendall: Interesting... I wonder if that actually might explain a real problem with the vorticity diagnosics!

jshipton (approved):
thanks @tommbendall !

PR #735, opened 2026-05-06
Correct cloud conversion code

12shughes:
Correct `SWSaturationAdjustment()` to give correct conversion of cloud into water vapour. Compare `gamma_v*(q_v-q_sat)` to `q_c` rather than comparison without `gamma_v` and multiplication with `gamma_v` occurring later.


gusto/physics/shallow_water_microphysics.py
```diff
@@ -323,13 +317,16 @@ def __init__(self, equation, saturation_curve,
             assert not isinstance(gamma_v, FunctionType), "If time_varying_thermal_feedback is not True then gamma_v cannot be a Python function."
             self.gamma_v = gamma_v
 
+        sat_adj_expr = self.gamma_v * (self.water_v - self.saturation_curve) / self.tau
+        sat_adj_expr = max_value(sat_adj_expr, -self.cloud / self.tau)
```
tommbendall: Just for numerical safety, we could keep the `min_value` that ensures that water vapour won't go negative (for instance if `self.tau` is different to the timestep)
```suggestion
        # Clip the increment to avoid generating negative values
        sat_adj_expr = max_value(min_value(sat_adj_expr, self.water_v / self.tau), -self.cloud / self.tau)
```

gusto/physics/shallow_water_microphysics.py
```diff
@@ -323,13 +317,16 @@ def __init__(self, equation, saturation_curve,
...
+
         # Factors for multiplying source for different variables
         # the order matches the order in V_idx (vapour, cloud, depth, buoyancy)
-        factors = [self.gamma_v, -self.gamma_v]
+        factors = [1, -1]
```
tommbendall: This probably doesn't matter, but in other schemes we set these factors to be a `Constant`
```suggestion
        factors = [Constant(1.0), Constant(-1.0)]
```

tommbendall (changes requested):
Thanks for tackling this @12shughes ! I have left a couple of suggestions but happy for you to disagree

tommbendall (approved):

PR #744, opened 2026-06-12
Fix Ubuntu tests by moving mpiexec outside pytest

### LINKED ISSUE #742: Parallel tests failing on Ubuntu
Gusto's parallel tests are currently failing on Ubuntu installations because mpiexec is not starting up. Running mpiexec within pytest is known to be fragile and the solution is to run mpiexec outside pytest (as Firedrake does).

dacreman:
Parallel tests are failing on Ubuntu because mpiexec is being run inside pytest (see #742). This PR changes the Makefile so that mpiexec is run outside pytest which fixes the problem.

Running mpiexec outside pytest requires a separate invocation for each number of MPI ranks tested. This PR also modifies the verbosity of the test output so that it is easier to keep track of what has passed and what has failed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,22 +18,64 @@ lint:
 	@echo "    Linting gusto plotting scripts"
 	@python3 -m flake8 $(FLAKE8_FORMAT) plotting
 
+# Set the Gusto logging level to avoid showing too much output
+#export GUSTO_CONSOLE_LOG_LEVEL=WARNING
+
+# Disable threading as this can slow down tests
+export OMP_NUM_THREADS=1
+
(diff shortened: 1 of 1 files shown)


JHopeCollins (approved):
LGTM, but I think you will need an approval from @jshipton or @tommbendall before it can be merged.

CI failure looks completely unrelated.

tommbendall:
Thanks Dave, it looks good to me but I just want to run the test suite my home laptop before we merge it, so I'll get this to you once I've done that!

tommbendall (approved):

PR #750, opened 2026-07-09
An IMEX fix

tommbendall:
Ensures that the IMEX first guess on subsequent stages is set to be the value from the previous stage (stolen from @atb1995)


jshipton (commented):
@atb1995 could you confirm exactly what this does? It looks to me like the initial guess for each stage is set to be the output from the previous stage. Is that correct and is that the intention?

atb1995:
> @atb1995 could you confirm exactly what this does? It looks to me like the initial guess for each stage is set to be the output from the previous stage. Is that correct and is that the intention?

Yes correct - on this first stage this was throwing an error (oversight by me). 

To be fair I don't think we need this line at all, `self.x_out` should be xn, x1, x2 etc. already as it is set at the top of the `apply` and is the output of the solver, but I believe Tom wanted to get this on as he had already run with this code fix (written by me).

tommbendall:
> > @atb1995 could you confirm exactly what this does? It looks to me like the initial guess for each stage is set to be the output from the previous stage. Is that correct and is that the intention?
> 
> Yes correct - on this first stage this was throwing an error (oversight by me).
> 
> To be fair I don't think we need this line at all, `self.x_out` should be xn, x1, x2 etc. already as it is set at the top of the `apply` and is the output of the solver, but I believe Tom wanted to get this on as he had already run with this code fix (written by me).

I think we just need to get onto main whatever we will do for our various papers. If there is a better fix but we run results with that fix, then we should use the better fix

gusto/time_discretisation/imex_runge_kutta.py
```diff
@@ -266,7 +266,9 @@ def apply(self, x_out, x_in):
 
             self.solver = solver_list[stage-self.solver_start_stage]
             # Set initial solver guess
-            self.x_out.assign(self.xs[stage-1])
+            if stage != self.solver_start_stage:
+                self.x_out.assign(self.xs[stage-1])
+
```
atb1995:
```suggestion
```

atb1995: Removing this line should do the exact same thing

atb1995 (changes requested):

jshipton (approved):
ok, I get it now - thanks @tommbendall and @atb1995 !

atb1995 (approved):

PR #751, opened 2026-07-09
Mountain Tests

tommbendall:
Introduces the "consistent metrics", and changes the function space used to define the orography


jshipton (approved):
It would be good to document this somewhere / somehow...

PR #752, opened 2026-07-13
Reordering options to fix diverged solve

dacreman:
Higher resolution baroclinic wave configurations are crashing due to the solver failing to converge (DIVERGED_LINEAR_SOLVE error). The baroclinic wave runs successfully with a C34 cubed-sphere grid but fails with C48 resolution.
This PR fixes the crash by getting PETSc to work out a better ordering before calculating the ILU factorisation.


JHopeCollins (approved):
Yup, looks right to me. Glad we finally got to the bottom of this!

tommbendall (approved):
Thanks very much @dacreman and @JHopeCollins for getting to the bottom of this, and well done on persevering!

PR #753, opened 2026-07-15
Update mountain mesh to use DG1

tommbendall:
Defining the mesh coordinates in DG1 gives good mountain wave results for the lowest-order configuration, and seems to have no impact for the element_order=1 configuration, so this changes the example to make it easier to pick this up in the future.


jshipton (approved):
great - thanks Tom!

PR #754, opened 2026-07-15
MMR (mean mixing ratio) limiter update

ta440:
Updates the formulation and implementation of the mean mixing ratio (MMR) limiter. The previous limiter worked in the Terminator Toy test, but uses in other transport tests highlighted an isssue with conservation. The change is to now compute mean fields after transport, rather than transporting the mean fields throughout a single timestep.

This pull request includes the following modifications:
1. Computes the mean fields after each transport step, rather than transporting the mean fields. This avoids issues with dispersion errors in the lowest order space breaking conservation when blending.
2. The MMR augmentation does not require a modified residual any more, as the mean fields are not transported, but just computed via projection. This greatly shortens the code!
3. The limiting kernel uses WR instead of INC for lamda (the blending weights) to correct an error in the previous version.
4. The test now is for a transport case, rather than the Terminator Toy, as this is found to be a more thorough test of the limiter.


gusto/spatial_methods/augmentation.py
```diff
@@ -257,9 +254,10 @@ def update(self, x_in_mixed):
 
 class MeanMixingRatio(Augmentation):
     """
-    This augments a transport problem involving a mixing ratio to
-    include a mean mixing ratio field. This enables posivity to be
-    ensured during conservative transport.
+    This augments a transport problem involving a k=1 mixing ratio, by adding
+    a mean mixing ratio field. This enables posivity to be
```
tommbendall:
```suggestion
    a mean mixing ratio field. This enables positivity to be
```

tommbendall (changes requested):
I spotted one small typo but am otherwise very happy for this. And definitely pleased that the code is now simpler. Well done and thanks Tim!

ta440:
> I spotted one small typo but am otherwise very happy for this. And definitely pleased that the code is now simpler. Well done and thanks Tim!

Typo has been fixed.

tommbendall (approved):

PR #756, opened 2026-07-23
Remove unnecessary firedrake.__future__ import

connorjward:
I am removing this from Firedrake `main` in https://github.com/firedrakeproject/firedrake/pull/5269 (merged)

### DIFF
--- a/gusto/physics/chemistry.py
+++ b/gusto/physics/chemistry.py
@@ -1,7 +1,6 @@
 """Objects describe chemical conversion and reaction processes."""
 
 from firedrake import dx, split, Function, sqrt, exp, Constant, conditional, max_value, min_value, assemble
-from firedrake.__future__ import interpolate
 from firedrake.fml import subject
 from gusto.core.labels import prognostic, source_label
 from gusto.core.logging import logger
--- a/gusto/physics/chemistry.py
+++ b/gusto/physics/chemistry.py
@@ -1,6 +1,6 @@
 """Objects describe chemical conversion and reaction processes."""
 
-from firedrake import dx, split, Function, sqrt, exp, Constant, conditional, max_value, min_value, assemble
+from firedrake import dx, split, Function, sqrt, exp, Constant, conditional, max_value, min_value, assemble, interpolate
 from firedrake.fml import subject
 from gusto.core.labels import prognostic, source_label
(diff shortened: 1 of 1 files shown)


tommbendall (approved):
Thanks Connor!

tommbendall:
The failure is a KGO change, presumably coming from the Firedrake side. This test will carry on failing on future unless we update the file with a head-of-main version of Firedrake. I don't know if that's something we feel we need to do? Or just let this test fail until we merge future back into main?

connorjward:
> The failure is a KGO change, presumably coming from the Firedrake side. This test will carry on failing on future unless we update the file with a head-of-main version of Firedrake. I don't know if that's something we feel we need to do? Or just let this test fail until we merge future back into main?

I don't quite understand. Is this test failing because your `future` branch needs fixes from `main`?

jshipton:
> > The failure is a KGO change, presumably coming from the Firedrake side. This test will carry on failing on future unless we update the file with a head-of-main version of Firedrake. I don't know if that's something we feel we need to do? Or just let this test fail until we merge future back into main?
> 
> I don't quite understand. Is this test failing because your `future` branch needs fixes from `main`?

I think Tom means that the KGO files need regenerating (probably because of minor Firedrake changes rather than anything done in Gusto) and maybe we'd rather only do that once.

tommbendall:
> > > The failure is a KGO change, presumably coming from the Firedrake side. This test will carry on failing on future unless we update the file with a head-of-main version of Firedrake. I don't know if that's something we feel we need to do? Or just let this test fail until we merge future back into main?
> > 
> > 
> > I don't quite understand. Is this test failing because your `future` branch needs fixes from `main`?
> 
> I think Tom means that the KGO files need regenerating (probably because of minor Firedrake changes rather than anything done in Gusto) and maybe we'd rather only do that once.

Yes that's what I meant!

My preference is to have this test fail on future I think. Otherwise we end up having to maintain and generate two sets of KGOs (those on future, those on main, with corresponding Firedrake versions) -- and those on future are harder for us to understand.

Or I could just slacken the tolerance for this test (again!)

connorjward:
OK no problem. It would be great to have this merged then because it's causing the smoke tests to fail.

jshipton:
I am happy to accept failure...

PR #758, opened 2026-08-27
Solver parameters for conservative tracer transport

tommbendall:
This is an attempt to specify sensible parameters for the coupled conservative tracer transport equation:
```
d(rho)/dt + div(rho*u) = 0
d(m*rho)/dt + div(m*rho*u) = 0
```

Although this needs to use a `NonlinearVariationalSolver` (since the `m` equation has a non-linear time derivative), since we can solve for rho first, the solution should not actually require any non-linear solves.

I also do some reorganising of solver parameters for the base `TimeDiscretisation` and the `ExplicitTimeDiscretisation`


gusto/time_discretisation/time_discretisation.py
```diff
@@ -509,10 +520,12 @@ def setup(self, equation, apply_bcs=True, *active_labels):
         """
         super().setup(equation, apply_bcs, *active_labels)
 
-        # get default solver options if none passed in
-        self.solver_parameters.update(mass_parameters(
-            self.fs, equation.domain.spaces))
-        self.solver_parameters['snes_type'] = 'ksponly'
+        # Set default solver params for explicit schemes, if none were passed in
```
jshipton, line 526: Do these overwrite the parameters specified on line 390 and is this because those on line 390 are a sensible choice for non-explicit time steppers? Just double checking I understand...

tommbendall: @atb1995 realised that the current behaviour on main is that any `solver_parameters` passed into an `ExplicitTimeDiscretisation` will be overwritten by the default `mass_parameters`.

I've tried to fix this by:
- ensuring we only pick up the defaults for `TimeDiscretisation` if no solver parameters are passed as arguments
- ensuring we only pick up the `mass_parameters` defaults for `ExplicitTimeDiscretisation` if no solver parameters are passed. This has to be done in `ExplicitTimeDiscretisation.setup` as we need to know the final function space being used, and means it will overwrite the previous defaults set in `TimeDiscretisation`.

I haven't changed the parameters themselves, and have no idea if these were sensible, but if I change them then all KGOs will change...

jshipton: I see - thanks!

jshipton (approved):

PR #759, opened 2026-09-01
routine to generate balanced initial fields for SWE

jshipton:
Routine to initialise shallow water flows from a given vorticity field such that the initial flow is non divergent and has div(u_t)=0. The steps are:

1. Compute the streamfunction from the vorticity.
2. Compute the velocity from the streamfunction.
3. Solve a mixed Poisson equation for (v, depth) where v=u_t and the right hand side comes from the vector invariant form of the velocity equation and setting div(v)=0.
4. Adjust the depth to have the correct mean.

### DIFF
--- a/gusto/initialisation/sw_balance.py
+++ b/gusto/initialisation/sw_balance.py
@@ -2,16 +2,23 @@ from firedrake import TestFunction, TrialFunction, Function, \
     dot, grad, dx, VectorSpaceBasis, solve, TestFunctions, TrialFunctions, \
     inner, div, Constant, assemble
 
-def nondivergent_velocity(equation, zeta0, u0, D0):
+
+def nondivergent_flow(equation, zeta0, u0, D0):
     """
+    Returns u0 and D0, balanced velocity and depth fields, given a
+    vorticity field zeta0. Balance is defined as
 
     Args:
+        equation (:class:`PrognosticEquation`): the model's equation object.
+        zeta0 (:class:`ufl.Expr`): the input vorticity field.
+        u0 (:class:`Function`): the velocity to be returned.
+        D0 (:class:`Function`): the depth to be returned.
     """
 
     domain = equation.domain
     Vcg = domain.spaces("H1")
 
-    # compute initial streamfunction
+    # compute initial streamfunction from vorticity by solving Poisson equation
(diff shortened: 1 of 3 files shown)


gusto/initialisation/sw_balance.py
```diff
@@ -0,0 +1,57 @@
+from firedrake import TestFunction, TrialFunction, Function, \
+    dot, grad, dx, VectorSpaceBasis, solve, TestFunctions, TrialFunctions, \
+    inner, div, Constant, assemble
+
+
+def nondivergent_flow(equation, zeta0, u0, D0):
+    """
+    Returns u0 and D0, balanced velocity and depth fields, given a
+    vorticity field zeta0. Balance is defined as
```
tommbendall, line 9: There's a missing comment here!

gusto/initialisation/sw_balance.py
```diff
@@ -0,0 +1,57 @@
...
+    L = (
+        -(f + zeta0) * inner(p, domain.perp(u0)) * dx
+        + 0.5 * div(p) * dot(u0, u0) * dx
+    )
+    w = Function(W)
+    solve(a == L, w, nullspace=nullspace)
+    _, D = w.subfunctions
+    D0.assign(D)
+
+    # adjust depth to have initial mean of H as set in the parameters
```
tommbendall, line 53: Are there any situations where we might want to set this differently? e.g. the min/max, or specifying some mean that isn't `H`? If so we could specify that through the argument list

gusto/initialisation/sw_balance.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def nondivergent_flow(equation, zeta0, u0, D0):
+    """
+    Returns u0 and D0, balanced velocity and depth fields, given a
+    vorticity field zeta0. Balance is defined as
+
+    Args:
+        equation (:class:`PrognosticEquation`): the model's equation object.
+        zeta0 (:class:`ufl.Expr`): the input vorticity field.
```
tommbendall, line 13: Is it always the relative vorticity? Could that be mentioned in the comment?

tommbendall (changes requested):

PR #763, opened 2026-09-07
Update README.md

jshipton:
copying from the README also copied the $ characters so I've just removed them...


tommbendall (approved):
Makes sense, thanks!

